PR #258, opened 2021-02-03
Added routes

josephch405:
Most if not all changes should be in place replacements; don't expect any changes aside from functioning URLs now. Would be helpful if someone with cluster access could test the cluster screens.

### DIFF
--- a/dashboard/src/main/home/Home.tsx
+++ b/dashboard/src/main/home/Home.tsx
@@ -24,17 +24,19 @@ import ProjectSettings from './project-settings/ProjectSettings';
 import ConfirmOverlay from 'components/ConfirmOverlay';
 import Modal from './modals/Modal';
 import * as FullStory from '@fullstory/browser';
+import { Redirect, RouteComponentProps, withRouter } from 'react-router';
+import PorterUrls from 'shared/urls';
 
-type PropsType = {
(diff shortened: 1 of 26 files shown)


dashboard/src/main/home/Home.tsx
```diff
@@ -71,18 +72,18 @@ export default class Home extends Component<PropsType, StateType> {
...
-          this.setState({ currentView: 'provisioner', sidebarReady: true, });
-        });
-      } else if (this.state.ghRedirect) {
-        this.setState({ currentView: 'integrations', sidebarReady: true, ghRedirect: false });
-      } else {
-        this.setState({ currentView: 'provisioner'}, () => {
-          this.setState({ currentView: 'dashboard', sidebarReady: true });
-        })
-      }
+      // if (res.data.length > 0 && (!currentCluster && !includesCompletedInfraSet(res.data))) {
```
sunguroku, line 75: Assuming you just temporarily commented it out to test routes, but this directs users to the provisioning screen depending on the status of resources that have been provisioned by porter. This part needs the most work and is necessary, so I'll merge this in when you get to refactor this part

josephch405: This should be patched back in now!

sunguroku (approved):

PR #263, opened 2021-02-10
Fixed routing param updating

josephch405:
Cleaned up a bit of unnecessary state setting as well


sunguroku:
by default, user gets routed to `/dashboard` without any query param. Visiting `/dashboard` without any query param leads to a loading screen atm, should default to ?tab=overview

josephch405:
Should be fixed now, added a default redirect if tab param is malformed/missing

sunguroku (approved):

PR #403, opened 2021-03-18
Logging Storage + Search 

abelanger5:
## Pull request type

- [ ] Bugfix
- [X] Feature

## Pull request checklist

Users can only view logs from a running container, and these logs are not searchable or downloadable. 
 
## What is the new behavior?

Users should be able to view old container logs, particularly from containers that have failed. These logs should be searchable and there should be a filter for error logs. Users should be able to download a log dump. Users should be able to forward logs to a storage location of their choice. 

We will capture logs using [fluentd](https://docs.fluentd.org/), and by default forward logs to an in-cluster elasticsearch instance. We will execute elasticsearch queries to provide search functionality for the user. The user should be able to view logs for a specific container, and should be able to sort by logs with an error message. This will be implemented similarly to metrics, where a running elasticsearch and fluentd service are detected in the cluster, and if they exist, a logging tab will be rendered. If the user chooses to forward to a different log storage backend, we will not display the logging tab. 
 
This means that we need to configure fluentd to forward logs to elasticsearch:

```
  <match **>
      @type elasticsearch
      include_tag_key true
      host "#{ENV['ELASTICSEARCH_HOST']}"
      port "#{ENV['ELASTICSEARCH_PORT']}"
      logstash_format true

      <buffer>
        @type file
        path /opt/bitnami/fluentd/logs/buffers/logs.buffer
        flush_thread_count 2
        flush_interval 5s
      </buffer>
    </match>
```

For now, we should just capture all logs from all containers as raw text. That is:

```
    <source>
        @type tail
        path /var/log/containers/*.log
        # exclude Fluentd logs
        exclude_path /var/log/containers/*fluentd*.log
        pos_file /opt/bitnami/fluentd/logs/buffers/fluentd-docker.pos
        tag kubernetes.*
        read_from_head true
        <parse>
          @type regexp
          expression /^(?<message>.*)$/im
        </parse>
      </source>
```

We should add some `exclude_path`s to ignore logs from system components, such as coredns. 

We may also want to run a `CronJob` that removes logs past a certain timestamp from ElasticSearch, otherwise the storage requirements will monotonically increase until the limit is reached.


abelanger5:
This is fairly outdated and probably won't be implemented for another few release cycles. Closing for now.

PR #404, opened 2021-03-18
Postgres Backups

abelanger5:
## Pull request type

- [ ] Bugfix
- [X] Feature

## Pull request checklist

Users deploying a database through Porter currently have no way to configure backups on a schedule, or restore data from a backup. 

## What is the new behavior?

For certain charts, Porter will offer backups that will be restorable from a "Backups" tab in the case of data corruption or deletion. As different deployments have different implementations for restoring backups, we will start by implementing this for the Postgres chart.  

This will be done similarly to metrics, using `velero` as the backup agent. If velero is installed in the cluster, and a user has launched a postgres chart, a "Backups" tab will be added to the postgres deployment. From this tab, the user can set the cronjob schedule of the backup, and if the schedule already exists, the tab will show the list of backups with a "restore" option. Users will also be able to view backups by clicking into the velero chart. 

In order to restore these backups in-place, we must create the backup in a new (temporary) namespace, backing up the entire postgres release. Once postgres is running again in the new namespace, we will run a `pg_dump` that migrates from the new postgres instance in the temporary namespace to the old postgres instance in the old namespace. This job should also overwrite the postgres secret in the old namespace with the secret in the temporary namespace. After this job exits successfully, the temporary namespace should be deleted. 

Backups should be designed to handle two different disaster scenarios:

1. The postgres release is accidentally deleted. In this case, the user must click into the "Velero" chart, find the most recent backup for postgres, and click "Restore." This should launch a new postgres chart in the current namespace (secrets, pvcs, and pvs will be skipped if they already exist). 
2. The data has been corrupted or deleted, but the postgres instance is still running.

### DIFF
--- a/internal/kubernetes/backups/backups.go
+++ b/internal/kubernetes/backups/backups.go
@@ -1,0 +1,1 @@
+package backups


abelanger5:
This is fairly outdated and probably won't be implemented for another few release cycles. Closing for now.

PR #497, opened 2021-04-14
Beta.3.env groups fe

jusrhee:
## Pull request type

- [ ] Bugfix
- [X] Feature

## Pull request checklist

- [X] If it's a backend change, tests for the changes have been added and `go test ./...` runs successfully from the root folder.

## What is the new behavior?

EnvGroups support


abelanger5:
A few things:
- Filter out config maps that aren't porter environment groups on the backend -- probably best to add an annotation to the configmap 
- The sort field in the Env groups tab doesn't seem to work at all 
- I'm not sure if this is desired, but uploading from env group will override keys that I've already created, including from other env groups. There may be cases where a user will want to load multiple env groups for a single deployment. 
- The "Ingress annotations" field for a standard web server deployment (in the Advanced tab) has a "Load from Env Group" button, need to case on this somehow

abelanger5 (approved):

PR #534, opened 2021-04-19
[0.2.0] Hide default domain in form.yaml

abelanger5:
## Pull request type

- [ ] Bugfix
- [ ] Feature
- [X] Other (please describe): change behavior of default domain to prevent it from being editable in `form.yaml`. 

## Pull request checklist

Users may attempt to modify the default domain when this is not possible. 

## What is the new behavior?

Hide the domain from `form.yaml` when it's set to a default domain field. This field should still be editable so that users can set up their own domain.


sunguroku:
already done, the bug turned out to be unrelated to ingress annotations.

PR #622, opened 2021-05-03
[0.3.0] Add job stop button without deleting the job

abelanger5:
## Pull request type

- [ ] Bugfix
- [X] Feature

## Pull request checklist

Jobs are run until completion, or the default maximum age. This makes infinite loops difficult to deal with. 

## What is the new behavior?

Provide button on frontend that can stop a running job. 

- Not natively supported on Kubernetes. Because the job needs to persist logs, we need to kill the process running within the pod, rather than deleting the pod itself. 
- Working theory is to run a `killall5` command (or similar) within the pod and add an annotation describing that we manually stopped this job, then case this on the frontend to show the user the job was manually stopped.


abelanger5:
Implemented graceful shutdown of jobs through a job sidecar container that should [share a process namespace](https://kubernetes.io/docs/tasks/configure-pod-container/share-process-namespace/) with the main job (so `shareProcessNamespace` should be set to `true` in the job spec). 

The sidecar container waits for a termination signal and gracefully attempts to shut down the main process by sending `SIGTERM` to the main process, and optionally the child processes. The sidecar container contains a `signal.sh` file that can be remotely executed the begin the shutdown process, for example:

```sh
kubectl exec [job_pod] -c [sidecar_container_name] -- ./signal.sh
```

PR #696, opened 2021-05-18
[0.3.0] Show selected namespace pods in status and metric tabs

oshtman:
## Pull request type

- [ ] Bugfix
- [x] Feature

## Pull request checklist

- [x] Docs have been reviewed and added / updated if needed

When viewing active pods for a deployment and helm charts with the same name had been deployed in different namespaces, they would all show up in the "Status / Metric" tabs even if they did not reside in the namespace of the chart currently being viewed. 

Issue Number: N/A

## What is the new behavior?

This PR adds a namespace query parameter send along from the front end to only show charts deployed in the actively viewed namespace.

CC @abelanger5


abelanger5 (approved):
Reproduced and tested locally, works great @oshtman! Merged to `master`, will be up on production later today 🚀

Issue #251, opened 2021-01-28
Display cluster IP address on dashboard


sunguroku:
- [ ] AWS
- [ ] GCP
- [ ] DO

jnfrati:
Hey, I'm kinda new around here, and in order to get used to all the environment I was thinking that this could be shown like this
![image](https://user-images.githubusercontent.com/23369263/119542670-2f21f300-bd66-11eb-84a4-0c641b82958b.png)

The idea would be that the user is able to click anywhere on the link and copy it to the clipboard, as the URLs could be too long to be shown, and select it would make the dashboard go to the apps page.

Let me know if you like it and I'll open a PR!

abelanger5:
Thanks @jnfrati, looks great! Please go ahead and make a PR (just link this issue in the PR template) and I'll review it ASAP.

jnfrati:
Great! I have one more question, I don't know what your target browsers are and if this should be available for Safari too. If it should be compatible, I would like to add the [clipboardjs](https://clipboardjs.com/) package to keep it simple and support almost all environments.

abelanger5:
Yes @jnfrati, good catch. Perhaps you can create a shared component with copy-to-clipboard functionality using clipboardjs? There are two other places where we currently copy to clipboard (`SettingsSection.tsx` and `InviteList.tsx`), and it would be good to consolidate.

jnfrati:
Okay, I'll try to cover those too to keep the clipboard usage consistent

PR #740, opened 2021-05-26
Cluster ip address on dashboard

jnfrati:
## Pull request type

- [ ] Bugfix
- [x] Feature

## Pull request checklist

The cluster list on the dashboard wasn't showing the cluster address

Issue Number:  #251 

-->

## What is the new behavior?

Now each cluster displays a text with the cluster IP and allows the user to copy it to the clipboard. 

image

In order to keep backward compatibility, I used [clipboardjs](https://clipboardjs.com/) on a custom component (CopyToClipboard) and implemented that component on other places that the clipboard was being used.


abelanger5:
Thanks for the PR @jnfrati! Just request my review on the PR once it's ready to review, and I'll take a look.

jnfrati:
@abelanger5 I think it's ready. Just let me know if you see something off and I'll try to take care of that ASAP!

abelanger5 (changes requested):
- Tested locally, works great @jnfrati! Verified all three components (invite list, webhook, cluster list) copy to the clipboard on both Safari and Chrome. 
- Request: responsive UI after copying to the clipboard, at least for the cluster IP address (but feel free to update the invite list behavior and webhook copy as well!). The icon could transform or you could add accompanying "Copied to clipboard" text somewhere...it's up to you.

abelanger5 (changes requested):
@jnfrati, after discussing with the team, the original use-case was to return the LoadBalancer IP address for easy configuration of custom domains. Really sorry for the lack of clarity on the initial issue (we've started enforcing proper specs). 

The IP address that's shown is the cluster endpoint, but to get the load balancer IP address you will have to update the get clusters endpoint. The two changes I'm requesting:
- [ ] Attach the load balancer IP address, if it exists, to the `GET .../clusters` endpoint. Only display the load balancer endpoint if it exists. This function will retrieve the correct load balancer IP address for a cluster set up with nginx as the ingress controller: https://github.com/porter-dev/porter/blob/390ef3ad1de1a9a28a5b960705b7bd4e651d4136/internal/kubernetes/domain/domain.go#L20
- [ ] As mentioned above, add responsive UI when copying the IP address. 

Please don't hesitate to reach out on Discord with any questions.

jnfrati:
@abelanger5 Okay! I'll start working on that 👍 As I'm not very comfortable with GoLang I'll probably be asking some questions on the Discord Server. Thank you for the clarifications!

jnfrati:
@abelanger5 I was thinking that maybe the implementation could be a little better if I add some function to get the Ingress IP inside the Cluster model, but let me know what do you think. 
Also, it seems to struggle a little bit when getting the IP so it produces some delay on loading the cluster list, so if you have any idea how to improve this it would be great!

server/api/cluster_handler.go
```diff
@@ -108,6 +110,28 @@ func (app *App) HandleListProjectClusters(w http.ResponseWriter, r *http.Request
 	extClusters := make([]*models.ClusterExternal, 0)
 
 	for _, cluster := range clusters {
```
abelanger5, line 112: To make this more efficient, you can make the API requests concurrently rather than synchronously. Would be easiest to create a `WaitGroup` using `sync.WaitGroup{}`. Create a goroutine via `go func` that adds 1 to the wait group for each cluster. Construct the agent and call `domain.GetNGINXIngressServiceIP` inside the goroutine. Call `.Wait` on the wait group to wait for each goroutine to complete.

You can try to benchmark if this improves the response time for a decent number of clusters (`n>5`).

abelanger5 (commented):
Yes, I believe this could be more efficient -- see my comment. What sort of latency are you seeing for this request?

We can't store the ingress IP inside the cluster model, as the ingress IP can change without interaction from the Porter dashboard. We also want to support clusters that weren't created on Porter, in which case users could potentially change the ingress IP using `kubectl` or `helm`.

server/api/cluster_handler.go
jnfrati: First of all, I wanted to thank you for the time you're spending helping me to make this contribution.

So, I tried with only two clusters, one of them offline (one minikube that I was making some tests on before), and the request time is really unstable, as it may take between 170ms or 3 seconds, this makes the interface go kinda crazy (not mention that the dashboard is making two sets of requests for the same info, IDK if that was on purpose or if I can give you a hand on that). 

Request time for just the DigitalOcean cluster
image

And this is how it looks when enabling the two clusters.

https://user-images.githubusercontent.com/23369263/119930094-2cd1bb80-bf55-11eb-9435-01f7d3ce75ef.mp4


Probably there's a ton of room for improvement on the way I'm making the subroutines (this is my first time working with golang), but even if this could be improved, I think that it would be better for the UX to have a separated request for the URL, and just show something like *"Loading cluster URL"* or some spinner until the URL is provided. This would enable the user to keep doing stuff without being stuck on the case that one of the clusters is having any kind of problem.

Let me know what you think about this, and thanks again for your time and for helping me on this one!

abelanger5: No problem at all! Really appreciate the contribution, impressive that you got this done your first time touching Go! 

First off, I agree -- this is probably too slow, and the timeouts for unreachable clusters justify a separate endpoint. I like the idea of a spinner waiting until the separate endpoint returns. We also have plans of creating a separate "Cluster dashboard" that shows an overview of the cluster (IP address, machine types, node summary/health, some basic metrics, summary of applications/jobs on the cluster). So, the endpoint that you write could form the basis for getting detailed cluster information. You will have to create a separate struct to return detailed cluster information, which should be returned on the endpoint `GET /projects/{project_id}/clusters/{cluster_id}`. You should then call these endpoints for the various clusters asynchronously on the dashboard. 

Also, the multiple loads are definitely not intentional. You could file this as a separate issue to tackle if you'd like (or for someone else to tackle), this PR is getting pretty hefty as-is :)

abelanger5 (commented):
Added a slight problem to fix in the goroutine, but otherwise I think it looks good. However, since it probably makes sense to split this into a separate set of endpoints, you probably won't need the goroutine anymore -- even if you decide to split up the endpoint, at least it's a good learning opportunity!

server/api/cluster_handler.go
jnfrati: I just pushed the commits to make the request partially to get the ingress IP, I'll be working on the responsive UI when the user copies over the weekend.
What I was wondering is that the API route ` GET /projects/{project_id}/clusters/{cluster_id}` requires different permissions than the ` GET /projects/{project_id}/clusters`, is that okay? I should try making another endpoint for the detailed cluster?

abelanger5: I completely forgot that we had the `GET /projects/{project_id}/clusters/{cluster_id}` in the first place! But yes, what you have is good -- the middleware just checks that the `cluster_id` belongs to the project that is being referenced, it doesn't require different permissions. We don't have RBAC quite yet.

server/api/cluster_handler.go
```diff
@@ -82,7 +82,29 @@ func (app *App) HandleReadProjectCluster(w http.ResponseWriter, r *http.Request)
...
+
+	var agent *kubernetes.Agent
+
+	if app.ServerConf.IsTesting {
+		agent = app.TestAgents.K8sAgent
+	} else {
+		agent, _ = kubernetes.GetAgentOutOfClusterConfig(form.OutOfClusterConfig)
+	}
+
+	endpoint, found, _ := domain.GetNGINXIngressServiceIP(agent.Clientset)
```
abelanger5, line 103: Should handle an error here. If `found` is false, but an error is thrown, it likely means that the cluster is not reachable, or the user does not have sufficient permissions to view ingresses on the cluster. We should try to pass an informative error to the user. I would just handle the cases of a timeout and unauthorized. 

One pattern that may be useful is to write a helper function in the `/internal/kubernetes` package that parse an error from the Kubernetes API server and will send a specific error type to the user. For example, you could declare structs `ErrTimeout` and `ErrUnauthorized` that implement the `error` interface, which you could then case on using the `errors.As` package. See here for more info: https://blog.golang.org/go1.13-errors. You should note the difference between a sentinel error and an error type here! I am suggesting to create an error type, so that the detailed error (rather than a generic sentinel error) from the Kubernetes server can be passed through to the user. 

On the frontend, we should show a different message than if the ingress IP is not found if there's an error. It might make sense to add an `warnings` field to the DetailedCluster object, to signal that this is a "non-fatal" error. But I'm not sure if this is a good pattern -- would love some ideas here!

dashboard/src/main/home/dashboard/ClusterList.tsx
```diff
@@ -58,29 +93,54 @@ class Templates extends Component<PropsType, StateType> {
...
+          <Loading />
+        </Url>
       );
-    });
+    }
+
+    if (!ingressIp.length) {
+      return (
+        <Url>
+          <Bolded>Cluster IP:</Bolded>
```
abelanger5, line 108: Can we perhaps call this "Ingress IP" or "Endpoint" or something? Not sure what's the most clear, but "Cluster IP" is too confusing because it makes me think of the cluster API server endpoint.

abelanger5 (commented):
💯 Tested locally -- works really well!

Looking forward to merging this in when you add the responsive UI commit. I tested against some failing clusters though, and I think the error handling needs to be slightly more clear (added some comments).

server/api/cluster_handler.go
jnfrati: Hey @abelanger5, I'm having issues getting this done. I've been researching how to handle errors, error types, and that kind of stuff. But from what I've seen I'm only getting a sentinel error from the request when I have a timeout or a connection refused, I tried using the ReasonForError function from the [k8s errors api](https://pkg.go.dev/k8s.io/apimachinery/pkg/api/errors) but I had no success at all. 

The error that I'm getting doing err.Error() is the following one:
`Get "https://host.docker.internal:63708/api/v1/services?labelSelector=app.kubernetes.io%2Fmanaged-by%3DHelm": dial tcp 192.168.65.2:63708: connect: connection refused` and it has a generic error type.

The only thing that comes to my mind to handle this is checking the string to check the string connection refused but its extremely gross from my perspective. If you can give me a hand on this or some tips to get an idea on how should I handle this would be awesome

abelanger5: Hmm yeah, so it looks like you should primarily use `errors.IsForbidden` and `errors.IsTimeout`, but perhaps there is a way to capture `ECONNREFUSED` using the data above? If this isn't possible, you can also generically capture errors that don't match any known error types. I wouldn't check the string (although we do that in a few places), since the k8s API changes frequently. 

Just to be clear, I was suggesting that you create two new error type, very roughly just something like:

```go
// for ECONNREFUSED and errors.IsTimeout
type ErrConnection struct {
  k8sErr error // (or maybe url.Error type?) 
}

func (e *ErrConnection) Error() string {
  fmt.Errorf("Could not connect to cluster: %s", e.k8sError.Error())
}

// for errors.IsForbidden
type ErrUnauthorized struct {
  k8sErr error // (or maybe url.Error type?) 
}

func (e *ErrUnauthorized) Error() string {
  fmt.Errorf("Unauthorized: %s", e.k8sError.Error())
}
```

Even though the underlying `k8sErr` may be a sentinel error or a unique error type, they can both just be set as the field `k8sError` and passed through to the client. Not sure if that makes sense, but lmk if there are any other questions!

The most important thing is that the user is aware of an error, especially if the Porter server is unable to connect to the cluster. Also, if they're getting `ErrUnauthorized`, they have likely set something up incorrectly, so I wanted to distinguish between those two cases. As long as the error is getting passed to the user, it's ok if you can't case on every error type.

jnfrati: Thank you! That was super useful yesterday, I don't know if it's the appropriate way to implement a catch for the error, I tried to move all the logic for detecting what kind of error was into an errors.go file inside the kubernetes package. Let me know what you think and I'll try to wrap up the final frontend changes today 👍

abelanger5: Yep, this is exactly what I had in mind 👍 

Just let me know when the final frontend changes are completed and I'll test it locally + merge it in.

server/api/cluster_handler.go
```diff
@@ -107,10 +136,13 @@ func (app *App) HandleListProjectClusters(w http.ResponseWriter, r *http.Request
 
 	extClusters := make([]*models.ClusterExternal, 0)
 
+	var wg sync.WaitGroup
```
abelanger5, line 139:
```suggestion
```
remove?

server/api/cluster_handler.go
```diff
@@ -107,10 +136,13 @@ func (app *App) HandleListProjectClusters(w http.ResponseWriter, r *http.Request
...
+
 	for _, cluster := range clusters {
 		extClusters = append(extClusters, cluster.Externalize())
 	}
 
+	wg.Wait()
```
abelanger5, line 145:
```suggestion
```
remove?

abelanger5 (changes requested):
Just need to remove those two lines and then I'll merge!

jnfrati:
@abelanger5 thank you for all the help you gave me to get this done! I wanted to double-check if you checked the changes under kubernetes/internal/errors, as I added an interface and an ErrExternalize struct to be able to show up more clean objects on the frontend. 
Also, I'm having some issues executing the tests for the API on WSL, I don't know if you were able to make sure that they're working, if not, I'll try to make those work tomorrow morning.

Again, awesome experience for my first OSS contribution 🤝

server/api/cluster_handler.go
```diff
@@ -4,9 +4,12 @@ import (
 	"encoding/json"
 	"net/http"
 	"strconv"
+	"sync"
```
abelanger5, line 7:
```suggestion
```

abelanger5 (approved):

abelanger5:
Yep, checked it and it looks good! Merged it in to `master`. I made sure all of the relevant tests were passing locally -- the tests on `master` are currently broken but I have a branch with fixed tests. 

Thanks for all of the hard work on this @jnfrati!

PR #757, opened 2021-06-01
[0.5.0] Cluster dashboard - Cluster nodes list

jnfrati:
## Pull request type

- [ ] Bugfix
- [x] Feature

## Pull request checklist

Currently, the user has no way to check the cluster nodes status.

## What is the new behavior?

A new screen to show that will work as a cluster dashboard. By default, it will list all the nodes available within the cluster with some info as the status of the node, ram usage, and CPU usage.

As the cluster may have a large list of nodes, I think that the best approach would be a table that can be paginated if a cluster has a large number of nodes and would give a clean and simple first sight of each node status.

image

I have some other approaches if this doesn't suit the visual that is wanted for this page.

For the frontend the idea is to have a new route for the cluster dashboard with the table, as the project doesn't have a strong dependency on @material/core components, I don't know if it would be a good idea to use the material table to display the nodes list, but it would probably make the development of the feature faster.

And on the backend a new endpoint `GET /projects/{project_id}/clusters/{cluster_id}/nodes` to get the node list. I found the [NodeInterface](https://pkg.go.dev/k8s.io/client-go@v0.18.8/kubernetes/typed/core/v1#NodeInterface) that has a List method to get all the nodes, but some advice on other options would be awesome 👍

### DIFF
--- a/dashboard/src/components/Table.tsx
+++ b/dashboard/src/components/Table.tsx
@@ -1,0 +1,108 @@
+import React from "react";
+import styled from "styled-components";
+import { Column, Row, useTable } from "react-table";
+
+export type TableProps = {
+  columns: Column<any>[];
+  data: any[];
+  onRowClick?: (row: Row) => void;
+};
+
+const Table: React.FC<TableProps> = ({
+  columns: columnsData,
+  data,
+  onRowClick,
+}) => {
+  const {
+    getTableProps,
+    getTableBodyProps,
+    headers,
+    columns,
+    rows,
+    prepareRow,
+    headerGroups,
+  } = useTable({
+    columns: columnsData,
+    data,
+  });
+
+  return (
+    <TableWrapper>
+      <StyledTable {...getTableProps()}>
+        <StyledTHead>
+          {headerGroups.map((headerGroup) => (
+            <StyledTr {...headerGroup.getHeaderGroupProps()}>
+              {headerGroup.headers.map((column) => (
+                <StyledTh {...column.getHeaderProps()}>
+                  {column.render("Header")}
+                </StyledTh>
+              ))}
+            </StyledTr>
+          ))}
+        </StyledTHead>
+        <tbody {...getTableBodyProps()}>
+          {rows.map((row) => {
+            prepareRow(row);
+
+            return (
+              <StyledTr
+                {...row.getRowProps()}
+                onClick={() => onRowClick && onRowClick(row)}
+                selected={false}
+              >
+                {row.cells.map( cell => (
+                  <StyledTd {...cell.getCellProps()}>{cell.render("Cell")}</StyledTd>
+                ))}
+              </StyledTr>
+            );
+          })}
+        </tbody>
+      </StyledTable>
+    </TableWrapper>
+  );
+};
+
+export default Table;
+
+const TableWrapper = styled.div`
+  padding-bottom: 20px;
+`;
+
+export const StyledTr = styled.tr`
+  line-height: 2.2em;
+  cursor: ${(props: { disableHover?: boolean; selected?: boolean }) =>
+    props.disableHover ? "" : "pointer"};
+  background: ${(props: { disableHover?: boolean; selected?: boolean }) =>
+    props.selected ? "#ffffff11" : ""};
+  :hover {
+    background: ${(props: { disableHover?: boolean; selected?: boolean }) =>
+      props.disableHover ? "" : "#ffffff22"};
+  }
+`;
+
+export const StyledTd = styled.td`
+  font-size: 13px;
+  color: #ffffff;
+  padding-left: 32px;
+`;
+
+export const StyledTHead = styled.thead`
+  width: 100%;
+`;
+
+export const StyledTh = styled.th`
+  text-align: left;
+  font-size: 13px;
+  font-weight: 500;
+  color: #aaaabb;
+  padding-left: 32px;
+`;
+
+export const StyledTable = styled.table`
+  width: 100%;
+  margin-top: 5px;
+  padding-left: 32px;
+  padding-bottom: 20px;
+  min-width: 500px;
+  border-collapse: collapse;
+`;
--- a/dashboard/src/components/Table.tsx
+++ b/dashboard/src/components/Table.tsx
@@ -31,7 +31,7 @@ const Table: React.FC<TableProps> = ({
       <StyledTable {...getTableProps()}>
         <StyledTHead>
           {headerGroups.map((headerGroup) => (
-            <StyledTr {...headerGroup.getHeaderGroupProps()}>
+            <StyledTr {...headerGroup.getHeaderGroupProps()} disableHover={true}>
               {headerGroup.headers.map((column) => (
                 <StyledTh {...column.getHeaderProps()}>
                   {column.render("Header")}
@@ -83,7 +83,6 @@ export const StyledTr = styled.tr`
 export const StyledTd = styled.td`
   font-size: 13px;
   color: #ffffff;
-  padding-left: 32px;
 `;
 
 export const StyledTHead = styled.thead`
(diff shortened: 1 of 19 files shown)


abelanger5:
Adding that endpoint looks solid. A few questions/comments:
- Just to clarify, "Failed" nodes will show at the top of the list? 
- Is pagination/search purely a visual implementation, or are you trying to implement this on the backend as well? It might make sense to implement this as frontend-only. Some considerations:
  - If you implement search as a backend endpoint, there is no way to match against a search expression in the k8s API (for example, see https://github.com/kubernetes/kubernetes/issues/72196), which means you will have to read all objects into memory on the Go side for every search. 
  - Pagination would be supported on the backend with the Kubernetes client, which has `Limit` and `Continue` on the `metav1.ListOptions` parameters. 
- It would be good to sum all of the node resources and display them in the top as a total. 
- It would also be useful to display by the `LastTransitionTime` of the nodes somehow...if a node has crashed and then restarted, it might show "Running," but it would be useful to know when the transition time is.

jusrhee:
image

Think it makes sense to generalize the table component in RevisionSelector and adopt it for this use case. Row height should be adjustable because eventually it will be great to have inline graph previews for RAM and CPU:
image

jusrhee:
Also, for status a small color indicator should appear inline (see the objects nested under ChartList for reference)

image

jnfrati:
Thanks for the comments!

@jusrhee I'll try to generalize the table component so we can reuse it in other places as well, but probably it would be good to separate that on another PR to avoid changes on other modules/pages on this one, just let me know what you think about this.

And @abelanger5 in response to your comment: 

> Just to clarify, "Failed" nodes will show at the top of the list?

Yes, the failing nodes will be displayed on top of the table always

> Is pagination/search purely a visual implementation, or are you trying to implement this on the backend as well

Based on that issue I'll try to avoid a backend implementation of search and pagination (search on the front & pagination on the back could be too complex and time-consuming), instead, I'll make sure to make the payload of the request as light as possible (although this can be a problem when trying to implement graphs, but I think that can be covered on another PR)

>It would also be useful to display by the LastTransitionTime of the nodes somehow...if a node has crashed and then restarted, it might show "Running," but it would be useful to know when the transition time is.

I have to make a little research about the `LastTransitionTime` to check out which would the best way to implement it.

abelanger5:
> I'll try to generalize the table component so we can reuse it in other places as well, but probably it would be good to separate that on another PR to avoid changes on other modules/pages on this one, just let me know what you think about this.

Yes, just make the component generalizable for now but don't change the components anywhere else, this can be tackled later.

jnfrati:
As an update, for the table generalization, I implemented [react table](https://github.com/tannerlinsley/react-table), in order to be more extensible when we come to more complex cases for tables. 

And for the backend side, basically, I search for all the non terminated pods inside a node and then make a summary of the resources used by all the pods. The functions implemented on `internal/kubernetes/nodes/helpers.go` are using almost the same logic as kubectl to make all the calculations and data search.

I think the following TODOs would cover everything needed to wrap up the PR: 

- Error management
- Display the [conditions](https://pkg.go.dev/k8s.io/api/core/v1#NodeCondition) and status (if it's running or not) of the node 
- Try to improve performance (with a little number of nodes and pods the request takes approx 2sec to complete)
- Implement front end search for nodes
- Sort list by status
- Add button on cluster dashboard to remove the dashboard
- Style details
- Cleanup and improve naming

@abelanger5 and @jusrhee I would like if you can make a quick review of the changes and check if you consider adding something to the TODO list.

abelanger5:
Awesome @jnfrati! A few comments on the todo:

> - Try to improve performance (with a little number of nodes and pods the request takes approx 2sec to complete)
The immediate thing that comes to mind is using multiple threads (similar to the previous PR) to get all of the pods for the nodes. So these lines should be in a separate `go func`: 

```go
podList := getPodsForNode(clientset, node.Name)		
nodeUsage := DescribeNodeResource(podList, &amp;node)		
extNodeList[i] = nodeUsage.Externalize(&amp;node)
```

> Add button on cluster dashboard to remove the dashboard
Could you clarify what you mean by this?

The code generally looks pretty good, but there are some additional variables that we definitely don't need. I'll go through the code line-by-line when the PR is ready though!

jnfrati:
> > * Add button on cluster dashboard to remove the dashboard
> 
> Could you clarify what you mean by this?

This was a mistyping, I meant the button to remove the cluster, rn we implement that as a modal when you click the cluster on the sidebar. The idea would be to have a new button that opens that pop up around here:

image

Let me know what you think about it!

jnfrati:
> Replace the cluster dashboard icon with the one shown (should be same size as the one in Dashboard)

@jusrhee Do you mean that it should have the same styling as this icon that we show on the project dashboard? 
image

abelanger5:
> The idea would be to have a new button that opens that pop up around here

Can we instead make it a tab called "Cluster Settings"? I think we'll eventually remove the "Settings" tab from the sidebar for the project settings, and add it as a tab to the project dashboard as well.

jusrhee:
> > Replace the cluster dashboard icon with the one shown (should be same size as the one in Dashboard)
> 
> @jusrhee Do you mean that it should have the same styling as this icon that we show on the project dashboard?
> image

Yup, exactly.

jnfrati:
Hey @abelanger5 & @jusrhee, I think nothing's left for the PR in terms of functionality, I have to double-check everything to see if I'm missing something and update the PR with some comments about what's done, but if you want to start testing it and make maybe some reviews it would be super helpful!

dashboard/src/components/Table.tsx
```diff
@@ -0,0 +1,179 @@
...
+
+export default Table;
+
+const TableWrapper = styled.div`
+  padding-bottom: 20px;
+`;
+
+export const StyledTr = styled.tr`
+  line-height: 2.2em;
+  cursor: ${(props: { disableHover?: boolean; selected?: boolean }) =>
```
abelanger5, line 135: I think this is probably the line that's turning the table rows into `cursor: pointer` right? It should only be `cursor: pointer` if the element is clickable, otherwise just use the default cursor.

internal/kubernetes/nodes/helpers.go
```diff
@@ -0,0 +1,121 @@
...
+			} else {
+				value.Add(podLimitValue)
+				limits[podLimitName] = value
+			}
+		}
+	}
+	return
+}
+
+func PodRequestsAndLimits(pod *corev1.Pod) (reqs, limits corev1.ResourceList) {
```
abelanger5, line 32: This function should be commented since it is exported.

internal/kubernetes/nodes/helpers.go
```diff
@@ -0,0 +1,121 @@
...
+			continue
+		} else {
+			if quantity.Cmp(value) > 0 {
+				list[name] = quantity.DeepCopy()
+			}
+		}
+	}
+}
+
+func DescribeNodeResource(nodeNonTerminatedPodsList *corev1.PodList, node *corev1.Node) *NodeUsage {
```
abelanger5, line 85: This function should be commented since it is exported.

abelanger5 (changes requested):
There are a few minor things to change, but otherwise this looks really good! A few frontend/styling things:

- Sometimes the name will render to be really large, it might be better to render "Delete Cluster"

image

- If possible, it would be good to add warning text above the "Delete Cluster" button explaining that this action cannot be undone (see the project deletion tab in Settings for reference).

- I think it would be nice to try to re-use the search bar styling from the Github repos page. The search bar icon is very large and looks out-of-place. Reference:

image

- You need some `padding` on the table row, otherwise it looks very cramped when highlighted: 

image

dashboard/src/components/Table.tsx
jnfrati: For what I understand, the changes I made should accomplish this as well, but let me know if there's something I'm missing

abelanger5 (approved):

PR #769, opened 2021-06-09
[0.5.0] - Cluster dashboard namespaces list

jnfrati:
## Pull request type

- [ ] Bugfix
- [x] Feature

## Pull request checklist

The user doesn't have any option on the dashboard to either list or delete a namespace from a cluster

## What is the new behavior?

A new tab on the cluster dashboard to show all the namespaces available and an option to delete them, it can be on a grid or list format, each element on the list will have a dropdown button that will show the option for deletion and on top of the list, there will be a button that will trigger the NamespaceModal to create a new modal.

The deletion will have an extra step where the user has to write the name of the namespace to guarantee that it may not delete de namespace by error

Using the already implemented endpoints to implement the actions
`GET /projects/{project_id}/k8s/namespaces`
`POST /projects/{project_id}/k8s/namespaces/create`
`DELETE /projects/{project_id}/k8s/namespaces/delete`


jnfrati:
@jusrhee The two options provided to display all the namespaces would be 
image
image

I think the grid would be the prettiest option, but in case the idea is to implement some extra info on the cards in short, it would be better to have a list as we have on the applications screen

jusrhee:
> I think the grid would be the prettiest option, but in case the idea is to implement some extra info on the cards in short, it would be better to have a list as we have on the applications screen

I agree. Long term there might be a list/grid view option at the top, but let's go with list for now just because I'd definitely anticipate wanting to expand to show either more information or nested objects

FoxFortino:
Looks good 👍

jnfrati:
@jusrhee @abelanger5 The PR is ready for review! 

Some annotations to let you know about the implementations:

- The namespaces list is based entirely on WebSockets, as it provides a cleaner way to add, update or remove objects from the namespaces list.
- Added a filter on all places that requests the namespaces to remove terminating namespaces
- I ended up using a 2 column grid to avoid text wrap on large namespaces 
- The namespaces "default" or any other that starts with "kube-" are restricted for creation or deletion
- As a next step it would be nice to implement a stronger typing of namespaces and probably implement socket.io client or [react-use-websocket hook](https://www.npmjs.com/package/react-use-websocket) to provide a cleaner way to implement WebSockets on react

Let me know what you think!

dashboard/src/main/home/modals/NamespaceModal.tsx
```diff
@@ -21,7 +21,17 @@ export default class NamespaceModal extends Component<PropsType, StateType> {
     status: null as string | null,
   };
 
+  isValidName = (namespaceName: string) =>
+    !/(^default$)|(^kube-.*)/.test(namespaceName);
+
   createNamespace = () => {
+    if (!this.isValidName(this.state.namespaceName)) {
```
abelanger5, line 28: Would be good to check that the namespace doesn't already exist from the frontend as well.

dashboard/src/main/home/cluster-dashboard/dashboard/NamespaceList.tsx
```diff
@@ -0,0 +1,330 @@
...
+    let protocol = window.location.protocol == "https:" ? "wss" : "ws";
+    wsRef.current = new WebSocket(
+      `${protocol}://${window.location.host}/api/projects/${currentProject.id}/k8s/namespace/status?cluster_id=${currentCluster.id}`
+    );
+
+    wsRef.current.onopen = () => {
+      console.log("Connected to websocket");
+    };
+
+    wsRef.current.onmessage = (evt: MessageEvent) => {
```
abelanger5, line 35: Why are there two different handlers for `.onmessage`? This one doesn't seem to do anything, can we remove this logic and just keep the handlers in `useEffect` in `NamespaceList`?

abelanger5 (changes requested):
@jnfrati looks really great! Functionality-wise everything works really well. Added some minor code comments. Some usability issues:
- Sort the namespaces in a consistent way, since it's different on each load. I'd prefer to put the editable namespaces at the top, and the non-editable namespaces at the bottom. 
- Add a warning to the deletion modal explaining that deleting the namespace cannot be undone (like we do for the cluster modal).
- Make the namespace clickable, and it should take you to the applications page with that namespace selected as the filter.

jusrhee:
@jnfrati looks great. just three minor style items:

image

^Let's use the same font size/weight as the application/job/env-group list items for the namespace name (and optionally can add the namespace icon here as well)

image

^For the three-dot option on the right, let's also start using the same size/color as the inline job buttons and make this the standard (sorry for not communicating this earlier - we'll just have to standardize things like this together for now)

image
image

^Let's also standardize the delete popup to be flush to the right and below the three-dot button (and just use the same style as the other places we have a small popup like for logout)

jnfrati:
As a detailed summary of what was added since the last revision:

- Namespaces regex to allow only lowercase letters, numbers, and dashes on name
- ChartList component refactored to Function component to allow prevention of race conditions on quick changes on filter/namespace filter based on a redirection from namespaces list
- Namespaces list sorted alphabetically and always will show up the ones that are allowed to deletion
- Added warning on delete namespace modal
- Implemented style changes to respect style consistency with application/jobs listing

abelanger5 (approved):
Awesome @jnfrati! Will merge now.

PR #776, opened 2021-06-14
[0.5.0] [Frontend, API] Launchflow Improvements

igalakhov:
## Pull request type

- [ ] Bugfix
- [x] Feature

## Pull request checklist

- We can only request 100 git repositories from the user account at once, so users with a large number of repositories will have problems with the frontend loading them quickly. 
- Once the user selects the git repository and branch, they have the option to either select an application folder or use some default buildpack.
- Application settings has inputs like "post start command" or "pre stop command" and it's not very clear what they do.
- Procfile can only be detected from the root folder

## What is the new behavior?

- The user will be able to search for a specific repository by some parameter (similar to how Heroku does it). When the initial list is loaded, the user sees their 10 most recent used repos.
- The frontend tells the user if a buildpack is detected automatically as they navigate through the contents of their repo.
- form.yaml has a way to write clarifications for more complicated options, in the form of an "info" tooltip next to the option. Currently this is just for string inputs but can be generalized easily once forms gets more refactored.
- procfile can be detected (and then the process to run) from any folder

- There needs to be a new server endpoint for searching the user's repositories. Github API provides this so there's nothing technically challenging here.
- `RepoList` is now a functional component and will now use a ref to hold the ids of all the repositories (which are only loaded once) and then do and concatenate search requests from that.
- There is no good way to search repos through the github API, so to at least have some improvement the API method for listing user repos is now concurrent.
- The auto buildpack detection isn't actually setting the buildpack on the backend, it's just looking through the files in the git repo branch root and using the specs in the requirements [here](https://docs.getporter.dev/docs/auto-deploy-requirements#auto-build-with-cloud-native-buildpacks). I'll make an issue for improvements on this process once the PR is merged.


dashboard/src/components/Button.tsx
```diff
@@ -0,0 +1,57 @@
+import React from "react";
+import styled from "styled-components";
+
+interface Props {
+  disabled?: boolean;
+  children: React.ReactNode;
+  onClick: () => void;
+}
+
+const Button = ({ children, disabled, onClick }: Props) => {
```
jnfrati, line 10: Hey! Just for a bit of quick advice on typing React components. There's no need to manually type the children prop, as React.FC can do it for you with best practices and support for any future change on that prop 👍 

```suggestion
const Button: React.FC<Props> = ({ children, disabled, onClick }) => {
```

On the other hand if you need to type a component that will not use the children prop you can use React.FunctionComponent that will prevent from accidentally passing children to that component

igalakhov: Thanks! Didn't know that this was a thing, will refactor it.

igalakhov:
Here's a video of a prototype for the initial flow. Does anyone have any thoughts?

https://user-images.githubusercontent.com/25856165/122093414-0fb43e00-cdd9-11eb-8fec-6f85ac56c845.mov

sunguroku:
> Here's a video of a prototype for the initial flow. Does anyone have any thoughts?
> 
>  github_search_demo_0.mov

Looks great!

server/api/git_repo_handler.go
```diff
@@ -68,43 +68,80 @@ func (app *App) HandleListRepos(w http.ResponseWriter, r *http.Request) {
...
-			app.handleErrorInternal(err, w)
-			return
+	// make workers to get pages concurrently
+	const WCOUNT = 5
+	numPages := resp.LastPage + 1
+	var workerErr error
+	var mu sync.Mutex
+	var wg sync.WaitGroup
+
+	worker := func(cp int, wg *sync.WaitGroup) {
```
abelanger5, line 94: You shouldn't have to pass the `*sync.WaitGroup` to each `go func`. The functions should be able to access the waitgroup, and waitgroup is thread-safe.

server/api/git_repo_handler.go
```diff
@@ -68,43 +68,80 @@ func (app *App) HandleListRepos(w http.ResponseWriter, r *http.Request) {
...
+			mu.Unlock()
+
+			cp += WCOUNT
 		}
+	}
 
-		allRepos = append(allRepos, repos...)
+	wg.Add(WCOUNT)
+	// page 1 is already loaded so we start with 2
+	for i := 1; i <= WCOUNT; i++ {
```
abelanger5, line 124:
```suggestion
	for i := 1; i <= max(WCOUNT, numPages); i++ {
```

Pseudo-code, but we probably shouldn't call the `worker` more than we need to?

abelanger5 (changes requested):
Added a few minor comments based on what I can see. 

Can we benchmark by setting the number of repos per page to something trivial, like 2? I'd like to make sure this is a significant improvement when `numPages` is 10 or greater.

server/api/git_repo_handler.go
igalakhov: Didn't know that. Fixed

server/api/git_repo_handler.go
igalakhov: Fixed

igalakhov:
How should we signal to the user that a buildpack is detected automatically? I have something like below right now, but there are definitely ways to improve to make it more clear:

image

igalakhov:
Does anyone have thoughts about adding better logic for `show_if` in forms? For this PR, the idea is to be able to list variables like `show_if: a b c` and have them be interpreted as an and clause but if anyone has other suggestions that'd be useful too

sunguroku:
> Does anyone have thoughts about adding better logic for `show_if` in forms? For this PR, the idea is to be able to list variables like `show_if: a b c` and have them be interpreted as an and clause but if anyone has other suggestions that'd be useful too

not immediately salient, but supporting "or" as well as "and" would def be useful in the future too. A potential use case, for example, is a setting that's available only on a select few cloud providers.

abelanger5:
> Does anyone have thoughts about adding better logic for `show_if` in forms? For this PR, the idea is to be able to list variables like `show_if: a b c` and have them be interpreted as an and clause but if anyone has other suggestions that'd be useful too

This sounds good, curious about the use-case you're specifically targeting for this PR? 

And any reason why we don't want to explicitly support `and` and `or` operators? Obviously we should avoid building a query language (for now) but I could see cases where the `or` operator would be useful. 

If we decide to support multiple operators, a quick note on syntax moving forward: while most Helm chart developers will likely be writing `form.yaml`, it might be natural to adopt Helm templating syntax. But since the Helm templating syntax (which in this case would be `and (and a b) c` has gotten a fair amount of backlash, it'd seem better to adopt something a bit easier like `jq` operator syntax which would be the more natural `and` and `or` operators, i.e. `a and b and c` (and we're using `jq` elsewhere so this might make sense).

igalakhov:
The use case I had in mind is that, since we're introducing variables that allow the user to check if a certain type of cloud is being used, they might want to do more complicated statements for showing certain section. As a basic example, they might want to check if they're using some cloud AND that a variable that toggles some database on that cloud is true.

I agree that explicitly supporting `and` and `or` (and maybe like `not`) is a good idea but my main concern about it is backward compatibility (if we ever end up changing the way this is done, the old `form.yaml`s might break). Would we allow for nested statements (for example something like `a and (not b) and (c or d)`)? I think it'd be nice to come up with a short spec for this before I get to implementing any of it to avoid confusion later on.

sunguroku:
One bug I found: It seems that organizational repositories are not visible with the new search feature and in the list of initially loaded repos. For instance, this repository (`porter-dev/porter`) doesn't show up in the list of repos. Only the personal repositories (`sunguroku/repo_name`) that you own or are invited to are visible.

On current prod: 
image

On this branch:
image

igalakhov:
@sunguroku it seems like this isn't caused by anything in the backend/frontend logic but rather by the way that github organizational access works. Here's this branch vs prod for me:

abelanger5 (changes requested):
Working great @igalakhov! A few things:
- Capture an "Enter" keystroke on the search bar to trigger a search
- Benchmark the concurrency solution for listing repos to ensure it's an actual improvement. Just set the number of results per page to something like 2 and look at the network latency between the two solutions. There are cases where concurrency isn't necessarily an improvement and I'd like to make sure this scales to ~1000 repos, which would be 10 API requests total.

igalakhov:
Doing a test with my own github, about ~130 repos and 2 repos per page:
With concurrency: average around 6 seconds
No concurrency: average around 11 seconds

abelanger5 (approved):

jusrhee:
Restyled + made it just autosearch as the user changes the input (since we aren't searching on the backend anyways).

jusrhee:
> Restyled + made it just autosearch as the user changes the input (since we aren't searching on the backend anyways).
@igalakhov we'll just keep this as a search button for now. Curious to get your thoughts on autosearching though as they update the input

abelanger5 (approved):

PR #783, opened 2021-06-16
[0.5.0] -  Generalized websocket implementation

jnfrati:
## Pull request type

- [ ] Bugfix
- [x] Feature

## Pull request checklist

Code replication of WebSocket connection method

## What is the new behavior?

Setup a new hook that can be used to connect to WebSockets on future migration to functional components


dashboard/src/shared/hooks/useWebsockets.ts
```diff
@@ -0,0 +1,136 @@
...
+export const useWebsockets = () => {
+  const websocketMap = useRef<WebsocketMap>({});
+  const websocketConfigMap = useRef<WebsocketConfigMap>({})
+  
+  /**
+   * Setup for a new websocket, after calling new websocket you can open the connection with openWebsocket
+   * @param id Id to access later the websocket config/connection
+   * @param apiEndpoint Endpoint to connect the websocket e.g: /api/websocket
+   * @param options Websocket listeners
+   * @returns
```
igalakhov, line 31: This is very minor, but looks like this line isn't fully written out

dashboard/src/shared/hooks/useWebsockets.ts
```diff
@@ -0,0 +1,136 @@
...
+
+    if (prevWs) {
+      prevWs.close();
+    }
+    
+    const ws = new WebSocket(wsConfig.url);
+    ws.onopen = wsConfig.onopen;
+    ws.onmessage = wsConfig.onmessage;
+    ws.onerror = wsConfig.onerror;
+    ws.onclose = wsConfig.onclose;
```
igalakhov, line 90: This could possibly be condensed to just `Object.assign(ws, wsConfig)` if there's no issue with setting .url on ws

dashboard/src/main/home/cluster-dashboard/chart/ChartList.tsx
```diff
@@ -8,6 +8,7 @@ import { PorterUrl } from "shared/routing";
 
 import Chart from "./Chart";
 import Loading from "components/Loading";
+import { useWebsockets } from "shared/hooks/useWebsockets";
 
 type PropsType = {
```
igalakhov, line 13: Maybe just call this `Props`, `PropsType` feels a bit redundant especially with a functional component.

igalakhov (dismissed):
This looks like a really nice abstraction! Added some very minor comments/thoughts but looks good.

dashboard/src/shared/hooks/useWebsockets.ts
jnfrati: Good idea, didn't thought about it.

dashboard/src/main/home/cluster-dashboard/chart/ChartList.tsx
jnfrati: I leave the PropsType name as it was previously, but we could start using a `[NameOfComponent]Props` pattern to name the typing. What do you think about it?

dashboard/src/shared/hooks/useWebsockets.ts
jnfrati: Didn't knew what else to write in order to make it more declarative but still simple. I'll try to improve this, but if you have any Ideas it would be great!

sunguroku:
This standardization is great and the websockets are still working as expected. But there's one bug in ChartList - I confirmed that the websocket itself is functioning properly but it seems like the FC refactor might have missed a bit of original logic. The status in each chart should live reload when there is a websocket update on the controller that belongs to a chart. Each websocket channel that ChartList listens to sends updates for all controllers in the cluster of that type (i.e. the `deployment` websocket channel sends updates about all deployments in the cluster), and the frontend does the job of parsing out only the updates for a controller that belongs to the specific chart. This is the purpose of the `ChartLookupTable`, and it seems like with this refactor `ChartLookupTable` is never getting populated (L177~193), so all controller updates are getting filtered out.

End to end, this results in a bug where the status of a particular chart remains constant even when there's a change. To recreate, make a change or trigger a webhook on any existing chart. In production, that particular chart will change its status to "Loading" while change is being propagated. On this branch, status will remain the same as what was initially displayed.

dashboard/src/main/home/cluster-dashboard/chart/ChartList.tsx
igalakhov: My personal reason for calling it `Props` is that:
1. The fact that it's a type is immediately obvious from the places it is used, so adding `Type` to the end feels redundant
2. This type is never exported anywhere, so as long as there's only one component in the file having the component name in it also feels redundant
But I also don't think it's a big deal, if you choose to go with the pattern you mentioned or `PropsType`, I'll change the FC components I refactored to use that too

dashboard/src/shared/hooks/useWebsockets.ts
igalakhov: Maybe like `@returns WebsocketConfig object`? Not a big deal either way though

dashboard/src/main/home/cluster-dashboard/chart/ChartList.tsx
jnfrati: I'm absolutely okay with that! I'll try to attach to the simple `type Props` then 👍

sunguroku:
nice, the live status update on chartlist is fixed, confirmed that it works with webhook trigger! I'm not sure if this fix is already in `master` and maybe that's why it's not included here but going back and forth between job and applications results in unchanged chartlist. Is the fix just not in this branch cuz it's already in master?

jnfrati:
> nice, the live status update on chartlist is fixed, confirmed that it works with webhook trigger! I'm not sure if this fix is already in `master` and maybe that's why it's not included here but going back and forth between job and applications results in unchanged chartlist. Is the fix just not in this branch cuz it's already in master?

Yep, that is on master! I'll merge that up to sync with master

abelanger5 (approved):

PR #784, opened 2021-06-16
[0.5.0] - Expanded node view on cluster dashboard

jnfrati:
## Pull request type

- [ ] Bugfix
- [x] Feature

## Pull request checklist

Currently, the state of the node is showing on a modal and doesn't display useful data as LastTransition time

## What is the new behavior?

When the user clicks on a row of the table it will redirect to a screen similar to what we have on ExpandedChart, with the full table. This will allow also to add metrics or more settings for the node in a close future

The /cluster-dashboard route will have now a set of subroutes to be easier to implement and redirect the user to the expanded node view.


jusrhee:
Looks good - definitely not a must, but I wonder if we should just have a shared or generalized higher order component for these kind of "expanded card" views (application, job, node, envgroup, etc).

jnfrati:
> Looks good - definitely not a must, but I wonder if we should just have a shared or generalized higher order component for these kind of "expanded card" views (application, job, node, envgroup, etc).

Probably it would be a good idea, idk if like a HOC or maybe just as a component that wraps up all the content. I'll try to implement it on another PR as a refactor for those components that are currently using this.

jnfrati:
@jusrhee Just an update of the current expanded node view. I feel like it's a little bit scratchy, so if you have any comment on improvements for this it would be great 👍 

image

sunguroku:
Not immediately relevant and is prob more fit for the next PR about metrics, but one thing to consider is that we may also want to present metrics across multiple nodes as well. For instance, one of the primary use cases of the node level view will be to diagnose autoscaling and determine how much capacity is left until autoscaling gets triggered. So later we may want a similar "expanded" component displayed straight from the nodes tab before clicking into a single node.

abelanger5:
This looks really good so far! A few points:
- I believe at the moment, the "Last Transition" field is actually the last heartbeat. My nodes from `kubectl` are reporting last transition at `Wed, 10 Mar 2021 17:20:45 -0500`, while the dashboard is reporting `6/17/2021, 8:56:55 AM`. 
- Can we make the text selectable on the rows? It's a bit annoying that I can't copy the last transition time. 
- Can we render a help icon next to the resource amounts, and display a useful message about what `Mi` and `m` mean? For example, "2148 Mi is 2148 Mebibytes, which is equivalent 2252 megabytes". And "1170m is 1170 millicores of virtual CPUs, which is 1.17 vCPUs". It's a question that comes up a lot. 
- I don't think the "Last Heartbeat" field is that important to show. I think it would be best to remove it until the use-case to display it explicitly comes up. 
- I'm not seeing the allocatable/usage fields in the most recent push. Have you pushed all of the latest commits? Screenshot:
image
- Can we render the node label `node.kubernetes.io/instance-type` next to the name of the node, both on the node list and in the expanded node view? I just checked on both GKE and EKS, and both have this label attached.

abelanger5 (changes requested):
It's coming together @jnfrati! A few usability things:
- The CPU/RAM fields should be selectable as well.  
- Is there any way to make the units the same, and just display the usage as a ratio? I know we discussed that it's difficult to get the correct units, but I think this is necessary. We can stick to just `Mi` and `m` as the standard units for now. So instead of showing 4 fields for usage/allocatable, just show:
  - `CPU: 1270 / 1930 m`
  - `RAM: 2404 / 3400 Mi`
- Can we add the machine type to the table columns on the Node list?

Happy to discuss these items.

jnfrati:
> It's coming together @jnfrati! A few usability things:
> 
> * The CPU/RAM fields should be selectable as well.
> * Is there any way to make the units the same, and just display the usage as a ratio? I know we discussed that it's difficult to get the correct units, but I think this is necessary. We can stick to just `Mi` and `m` as the standard units for now. So instead of showing 4 fields for usage/allocatable, just show:
>   
>   * `CPU: 1270 / 1930 m`
>   * `RAM: 2404 / 3400 Mi`
> * Can we add the machine type to the table columns on the Node list?
> 
> Happy to discuss these items.

Changes requested just pushed! Also added a help link in case the user has doubts about the total memory value, as it corresponds to the allocatable and not the total on node, let me know what you think!

abelanger5 (changes requested):
Just one last thing -- can you convert all units to `m` and `Mi`? Just write a function: capture `Ki` and `Gi` -- for `Gi`, multiply the value by 1024, and for `Ki` divide the value by 1024. And if there's no `Ki`, `Gi`, or `Mi`, just divide by 1024 * 1024.

abelanger5 (approved):

PR #789, opened 2021-06-18
[0.5.0] [Docs] Frontend Contribution Guide

igalakhov:
## Pull request type

- [ ] Bugfix
- [ ] Feature
- [x] Other (please describe): Just some files to describe how one should contribute frontend changes

## Pull request checklist

- People who want to contribute to the frontend don't know what to do

## What is the new behavior?

- People who want to contribute to the frontend now have a better idea of what to do


abelanger5 (commented):
Thanks for this @igalakhov @jnfrati! This is going to be really helpful.

Could you make a list of the immediate items (next ~1-2 weeks) that you would like to accomplish in terms of this transition? Like which components you would like to refactor first, which ones you'd like to move, the first steps in terms of routing, etc.

igalakhov:
> Could you make a list of the immediate items (next ~1-2 weeks) that you would like to accomplish in terms of this transition? Like which components you would like to refactor first, which ones you'd like to move, the first steps in terms of routing, etc.

I think @jnfrati  has a better idea of what needs to be done for routing for the main pages, but I think one of the big things that this refactor needs to touch is the form components since those will get more complicated as time goes on, so the faster they are refactored the easier things will be from now. I think the main two things to accomplish is to:
* Make every form component functional
* Move everything possible (like metaState, for example) into a `FormContext` that can handle all the form validation logic so the form components are more atomic (which can later be moved into a UI library).

jnfrati:
I've been thinking a couple of ways that we can handle this refactoring and the ones that comes to my head are the following: 

- Migrate the `dashboard/src/main/home/cluster-dashboard/env-groups` so they stay up on the same folder but will now have an own routing system as it has the `dashboard/src/main/home/cluster-dashboard/dashboard` folder (credits to @abelanger5 for the idea). This should be the simpler one as the components doesn't handle too much logic

- Refactor the `dashboard/src/main/Main.tsx` and `dashboard/src/main/MainWrapper` components so we can have the basement to keep refactoring below those two components. In simple terms, those two components handle routing, logic and context, three things that should be separated by concerns. This one will provide the bases to migrate to a "module" like folder structure and routing.

- Refactor the `dashboard/src/main/home/cluster-dashboard/ClusterDashboard.tsx` component, this is one of the most complicated as four of the most critical screens depends on this component ( the real screen cluster dashboard, jobs, applications, and env groups). The problem of leaving this component behind is that we're going to have some real troubles to make significant changes over the folder structure due to the implications that has this component with the subrouting.

In my personal opinion we should start with a soft refactor of the two "Main" components, and then move towards the ClusterDashboard, as it's a component that shouldn't exist at all, moving the logic that this component handles will allow us to start the heavy refactor under the /jobs and /application routes that are probably the most critical ones, and the ones that need a deep cleaning.

abelanger5:
> In my personal opinion we should start with a soft refactor of the two "Main" components, and then move towards the ClusterDashboard, as it's a component that shouldn't exist at all, moving the logic that this component handles will allow us to start the heavy refactor under the /jobs and /application routes that are probably the most critical ones, and the ones that need a deep cleaning.

Will let others chime in if they're like, but this sounds pretty good to me. I'm curious about exactly how you plan to break this component up into routes, context, and logic though? I know we discussed creating a `Routes.tsx` file at the top-level, but context is already a separate component and I'm not sure what can be improved in terms of the logic.

jnfrati:
> > In my personal opinion we should start with a soft refactor of the two "Main" components, and then move towards the ClusterDashboard, as it's a component that shouldn't exist at all, moving the logic that this component handles will allow us to start the heavy refactor under the /jobs and /application routes that are probably the most critical ones, and the ones that need a deep cleaning.
> 
> Will let others chime in if they're like, but this sounds pretty good to me. I'm curious about exactly how you plan to break this component up into routes, context, and logic though? I know we discussed creating a `Routes.tsx` file at the top-level, but context is already a separate component and I'm not sure what can be improved in terms of the logic.

So basically we handle all the auth logic inside that component, what I was planning was to make the component as simple as possible, withdrawing all that logic and make components for authenticated routes that will use a custom hook to check if the user is authenticated or not. This will allow us, later on, to implement RBAC in a simpler way in terms of guarding routes.

igalakhov:
> > > In my personal opinion we should start with a soft refactor of the two "Main" components, and then move towards the ClusterDashboard, as it's a component that shouldn't exist at all, moving the logic that this component handles will allow us to start the heavy refactor under the /jobs and /application routes that are probably the most critical ones, and the ones that need a deep cleaning.
> > 
> > 
> > Will let others chime in if they're like, but this sounds pretty good to me. I'm curious about exactly how you plan to break this component up into routes, context, and logic though? I know we discussed creating a `Routes.tsx` file at the top-level, but context is already a separate component and I'm not sure what can be improved in terms of the logic.
> 
> So basically we handle all the auth logic inside that component, what I was planning was to make the component as simple as possible, withdrawing all that logic and make components for authenticated routes that will use a custom hook to check if the user is authenticated or not. This will allow us, later on, to implement RBAC in a simpler way in terms of guarding routes.

Related to this (some hook like `useUser` seems like a great idea), are there any good libraries that give basic RBAC functionality (or at least something to build off from) or will we have to write it from scratch?

jnfrati:
> Related to this (some hook like `useUser` seems like a great idea), are there any good libraries that give basic RBAC functionality (or at least something to build off from) or will we have to write it from scratch?

We could make some research and look if there are some tools available to simplify this (e.g: [CASL](https://casl.js.org/v5/en/)), but I think the best way to do this is just make a custom implementation, probably following some of the advice that companies like [Auth0 ](https://auth0.com/docs/authorization/rbac/) give.
Relying on an external hooks library to handle RBAC will probably cause more trouble on how we want to implement it on the back end. But we can double check this when we're about to implement RBAC.

abelanger5:
Following up on internal conversation, there are some more items:
1. How to add RBAC on the frontend to certain components, and
2. When #814 is closed, some documentation around the design of the context, the validator function, and the mutation function, so it's easy to add new components to the forms. 

Also @jnfrati it's still not clear to me how we're breaking apart context and routing. Could someone clarify the latest thinking on this?

Once these items are added, I think the contribution guide should be structured like the following:

- Getting Started (contains the information about FC components, naming conventions, etc)
- Data Flow (contains information about context, props, etc)
- Routing
- Advanced 
  - RBAC 
  - Form Components (contains info from #814)

Last thing -- @jnfrati and @igalakhov I think you should consider how a new contributor would go about using this guide. I don't think it's very clear at the moment, and I think that structure would help (along with a table of contents at the top of the guide). I'll go through it line-by-line after we've added some of these items.

jnfrati:
Roadmap that will be introduce for contributors

image

abelanger5 (approved):

PR #814, opened 2021-06-22
[0.5.0] [Frontend, Forms] Forms Refactor

igalakhov:
## Pull request type

- [ ] Bugfix
- [x] Feature

## Pull request checklist

## What is the new behavior?

Forms are going to be refactored to use functional components and hopefully make adding more complicated features easier in the future.

Can be found in a file for this refactor. So far, the refactor is done with the fields checked off below:
- [x] variable
- [x] heading
- [x] subtitle
- [x] service-ip-list
- [x] resource-list
- [x] checkbox
- [x] env-key-value-array
- [x] key-value-array
- [x] array-input
- [x] string-input
- [x] string-input-password
- [x] number-input
- [x] select
- [x] provider-select
- [x] velero-create-backup
- [ ] base-64
- [ ] base-64-password


jusrhee:
@igalakhov form debugger is crashing for me even after pasting in just the very first part of the web chart's form (doesn't happen on prod):

```
name: Web
hasSource: true
tabs:
- name: main
  label: Main
  sections:
  - name: section_one
    contents: 
    - type: heading
      label: Container Settings
    - type: variable
      variable: showStartCommand
      settings:
        default: true
```

jusrhee:
Update looks good - now only running into rendering issues on the Tailscale chart: https://github.com/porter-dev/porter-charts/blob/master/addons/tailscale-relay/form.yaml

jusrhee:
Current Issues:
- expandedchart form doesn't read in existing values from revision for:
  - checkbox
  - key-value array
- fix unit issues (when value isn't overridden by user the default doubles the unit - e.g. "100MiMi" -> deploy fails)
- switching revisions doesn't update displayed form values
- can't load in env group from launch
- (nan yaml -> json parsing issue when launching w/ health check - try to recreate)

jusrhee:
Just a few final items:
- [ ] Env group loader should respect namespace updates from the launch flow
- [ ] Form in ExpandedChart should display updated values when selecting different revisions
- [ ] Handle no form case (ex: Grafana)

PR #833, opened 2021-06-29
[0.6.0] [API, Frontend] Github organizational access

igalakhov:
## Pull request type

- [ ] Bugfix
- [x] Feature

## Pull request checklist

Once a user links their github account (we currently use an OAuth app), both their personal and organizational repositories (or at least the ones they allowed) are visible to them.

## What is the new behavior?

There should be a way to only allow porter to see/ have access to repositories from specific organizations and restrict it from showing you your personal repositories. 

There is now a GitHub application (access is configured through environment variables) with OAuth enabled. When the user authenticates through this app, a record is made in their user model to record their OAuth access token. When a user installs an app into either their account or an organization, a webhook is triggered and a pair (installation id, account id) is recorded into the database. Then, whenever we want to make a list of repos that the user has access to, we can just check which of the users organizations (and the users id) intersects with this table and use those installation ids to authenticate.


igalakhov:
Okay, after reading up on the difference on Github Apps vs OAuth Apps I think it makes more sense to keep using an OAuth App for these features for the following reasons:
* Although Github Apps allow you to specifically select which repositories it has access to, the app itself ends up getting installed to the repositories, so it shows up in the "Integrations" tab. For example, if I give a Github App access to my personal repositories, EVERY repository I have will list the App.
* This isn't really good behavior because right now Porter has no reason to be included in the Integrations tab - since we only want to control visibility.
So, I think it's reasonable to keep Porter as an OAuth app and keep track of a list of organization names (and personal repositories) that the user does not want Porter to see. We can also write some docs that explain how to configure this OAuth access to include other organizations in case they don't show up for the user (there can be reasons for this that the user might not readily know about - for example an organization not having 3rd party access).

abelanger5:
> Although Github Apps allow you to specifically select which repositories it has access to, the app itself ends up getting installed to the repositories, so it shows up in the "Integrations" tab. For example, if I give a Github App access to my personal repositories, EVERY repository I have will list the App.

To further clarify, Render and Heroku show up as Github Apps within a specific repository **only after selecting that repository as source for an application**, but they use OAuth endpoints to list repositories. This is different from GitGuardian, which installs as a Github App from the beginning, and then appears in each of your repositories. This makes sense for GitGuardian, but not as much for us.  

Just to outline the three cases to resolve: 

1. **User has no way to restrict access to personal repos and organizations that allow all third-party apps**. This means a lot of unnecessary repos get added when the user searches their repositories. This seems to be resolved by adding additional settings to the **Integrations > Github** page. 

2. **After going through the OAuth consent screen, if a user fails to click on the "Request" button for a certain organization, it is difficult to request access**. They will have to go to the OAuth app from their personal settings page, and then [follow these instructions](https://docs.github.com/en/github/setting-up-and-managing-your-github-user-account/managing-your-membership-in-organizations/requesting-organization-approval-for-oauth-apps). This seems to be resolved via documentation. 

3. **If an organization restricts third-party applications, the organization does not show up on the OAuth consent screen, and the user does not know how to turn on access for this organization**. This is typically the case if an organization is relatively new, since the default setting for an organization is to restrict all third-party access. This seems to be resolved via documentation.

jusrhee:
Adversarial scenario, but if the user's whose repos we're filtering to connect to the organization deletes their account (or leaves the company), it's a bit weird that the connection is lost or tied to the user's account when they're really just trying to connect to the org. Can imagine this being an actual issue for sufficiently large companies (though not an immediate concern).

What are the main benefits of doing it the same way as Heroku/Render with a combination of OAuth and GitHub apps?

sunguroku:
If the reasoning against a GitHub App is mainly about Porter showing up on the integrations tab, I don't think that's a problem. It looks like Vercel and Netlify also follow the GitGuardian approach (shows up on integration tab of all repos) - if we want to avoid showing up in all repositories, we can do the Heroku/Render style workaround. I don't think porter showing up on integrations tab is a dealbreaker though.

If using a GitHub app introduces less points of failure on the user's end, I think github app is a better move - the less it needs to be supplemented by documentation the better. How do those three cases pan out if we were to use a GitHub app? Naively I'm inclined to use GitHub Apps like what other products (Netlify, Vercel, Heroku, Render) have done, and I would assume there's an advantage there.

igalakhov:
> If using a GitHub app introduces less points of failure on the user's end, I think github app is a better move - the less it needs to be supplemented by documentation the better. How do those three cases pan out if we were to use a GitHub app? Naively I'm inclined to use GitHub Apps like what other products (Netlify, Vercel, Heroku, Render) have done, and I would assume there's an advantage there.

The Github app access takes care of the 1st and 2nd points (3rd point still needs documentation since it's a settings issue) since the user is given the ability to select which repos the app is installed to (and thus, which repos Porter can see). My argument against using it though is that the installed GitHub app basically has no other purpose besides showing us visibility (since we take care of every other functionality an app could provide through GitHub actions anyways). And, it seems like OAuth apps is what GitHub wants us to use in cases like that to prevent further user confusion.

abelanger5:
> The Github app access takes care of the 1st and 2nd points (3rd point still needs documentation since it's a settings issue) since the user is given the ability to select which repos the app is installed to (and thus, which repos Porter can see). 

Please correct me if I'm wrong, but the equivalent case for the 2nd point for a Github app is after going to the Github app configuration screen, the user fails to request organization access, and would need some documentation on how to return to that screen anyway, correct? 

So it seems that the 1st point is the only one "automatically" resolved by a Github app, but I actually do think it's a major issue that the Github app is automatically added to every repository within that organization. For example, an organization might have multiple sensitive repositories with configuration stored as secrets, and the capabilities of our Github app would be to trigger new workflows, write files to the repo, and thereby read secrets. This introduces a massive single point of failure if the credentials are accidentally exposed. 

This is true of the OAuth app credentials as well, but this is why I think the Render approach is ideal in the longer term. The OAuth credentials are should just be used to list repositories, while the Github app gets added to repositories after the user selects that repo as source. It's thus added on an opt-in basis. This would also resolve this concern:

> Adversarial scenario, but if the user's whose repos we're filtering to connect to the organization deletes their account (or leaves the company), it's a bit weird that the connection is lost or tied to the user's account when they're really just trying to connect to the org.

Since the Github app that's tied to the repositories acts as its own user. So the user deleting the account would break the repository list, but not any integration that the Github app requires.

One last item that's unclear to me: GitGuardian has a screen on their own settings page listing repositories, but you can also configure the repositories by going to the Github app configuration screen. Do these simply map to the exact same operation? Why does GitGuardian have a separate integration screen in the first place (i.e. why not just redirect to the Github app configuration screen)?

igalakhov:
> Please correct me if I'm wrong, but the equivalent case for the 2nd point for a Github app is after going to the Github app configuration screen, the user fails to request organization access, and would need some documentation on how to return to that screen anyway, correct?

> One last item that's unclear to me: GitGuardian has a screen on their own settings page listing repositories, but you can also configure the repositories by going to the Github app configuration screen. Do these simply map to the exact same operation? Why does GitGuardian have a separate integration screen in the first place (i.e. why not just redirect to the Github app configuration screen)?

To answer both of these: at least for me, gitguardian redirects me to a screen that lets me install the app in more repositories when I click the configure button on their website. So, what I meant is that using that the 2nd point is taken care of because we can just have a button that directs the user to the same screen.

jusrhee:
> Since the Github app that's tied to the repositories acts as its own user. So the user deleting the account would break the repository list, but not any integration that the Github app requires.

I'm guessing that when multiple users are added to the same Render project, the repos are independently listed per user but they can each add the GitHub app to a specific repo assuming they have access. The less we can tie the connection/integration to a specific user account the better - just a matter of what teams/companies will expect at any reasonable scale. 

Even if the behavior is equivalent, it's more fragile as just an OAuth app - people leave organizations or move around internally, and it's uncanny to have to reset or risk losing the integration for the service you're using to manage your production environment if/when that happens... I think this is why it makes sense to do it the same way as Render and Heroku.

sunguroku:
Agree that eventual behavior should be Render/Heroku approach of combining both oauth and github apps (to avoid install across all repositories) - I'm ambivalent with any implementation as long as the current solution paves its way towards that approach

igalakhov:
I agree that the ideal implementation should not be tied to any user - but this specific issue (filtering certain organizations/personal repos) is user-specific so an OAuth app fits better since that's user-specific. If we want to prevent the user from deleting their github/losing access from breaking the deployment, the correct solution is probably to install an app to the repo at some point where they're done setting up their application (like how we do github actions), but I think that's something that's going to have to be done later and is unrelated to this issue.

jnfrati:
So, I'm sorry if I'm repeating something that's probably been said. If the problem is to filter repositories on organization level, and we currently implement an Oauth App, how is going to be the new approach using the same app to give granular access to repos on a technical perspective?

igalakhov:
> So, I'm sorry if I'm repeating something that's probably been said. If the problem is to filter repositories on organization level, and we currently implement an Oauth App, how is going to be the new approach using the same app to give granular access to repos on a technical perspective?

We don't expect users to be in too many organizations, so it makes sense to just store a field in the database that's a list of organizations the user doesn't want to see when they're listing repositories and give the user some sort of frontend UI for modifying this list.

igalakhov:
Okay, after the meeting and thinking about it further, it turns out that GitHub apps give a bit more functionality I like in this case. Specifically, GitHub apps are not tied to any specific user, and since that's the behavior that we want for these projects, it makes sense to tie a GitHub app installation (which comes with its own token) to a project. This basically gives all the control a user wants with very minimal overhead. So, once done, we will have the following changes:
* The integrations tab will now switch "git repository" to "Github" (since apps are GitHub specific)
* The user can then click through this option and install the Porter app into all repositories that they later want to be visible
* Authentication for the GitHub API will now happen through the app installation token instead of the user OAuth token

jusrhee:
> The user can then click through this option and install the Porter app into all repositories that they later want to be visible.

Makes sense. Should also include a link to the same install app screen from the launch flow:
image

abelanger5:
Sounds good, just one last thing I'm unclear on: is the list of repositories still attached to the currently logged-in user, or is it attached to the project? 

This will probably get ugly either way, since the `installation_id` is different between the installation of an app in an organization and in a personal user's account. I just checked on Render, and "disconnecting" the Github integration removes the list of all Github repositories from Render, but keeps the installations around on Github.

igalakhov:
> Sounds good, just one last thing I'm unclear on: is the list of repositories still attached to the currently logged-in user, or is it attached to the project?
> 
> This will probably get ugly either way, since the `installation_id` is different between the installation of an app in an organization and in a personal user's account. I just checked on Render, and "disconnecting" the Github integration removes the list of all Github repositories from Render, but keeps the installations around on Github.

Yeah, so it looks like each installation (to a personal account, or any org) gets a separate installation id (but only once), which is a bit concerning to think about in terms of how this will work between different projects. But, it seems like no matter oauth/app, the installation of an app is inherently tied to the user that installs it, so the setting might have to go somewhere in the account dropdown instead of what I originally said.

igalakhov:
Functionality for modal (and linking up app/installing it in appropriate places) is done, although it doesn't affect the repo listing/release logic yet. Still have to write database tests though

cmd/migrate/main.go
```diff
@@ -57,6 +57,9 @@ func main() {
 		&ints.ClusterTokenCache{},
 		&ints.RegTokenCache{},
 		&ints.HelmRepoTokenCache{},
+		&ints.GithubAppInstallation{},
+		&ints.GithubAppOAuthIntegration{},
+		&ints.GithubAppOAuthIntegration{},
```
abelanger5, line 62: Remove duplicate line

cmd/app/main.go
```diff
@@ -72,6 +72,7 @@ func main() {
 		&ints.ClusterTokenCache{},
 		&ints.RegTokenCache{},
 		&ints.HelmRepoTokenCache{},
+		&ints.GithubAppInstallation{},
```
abelanger5, line 75: Don't you need to add `&ints.GithubAppOAuthIntegration{}` here as well?

docs/guides/linking-github-account.md
```diff
@@ -0,0 +1,15 @@
+# Configuring Github Access
+
+> 🚧
+>
+> **Note:** Porter currently uses an oauth app to authenticate and gain access to repositories. This mechanism will be phased out
+> over the next few weeks to transition to the authentication method below. After this, all old applications will still work as intended 
+> but new applications will need to be authenticated through the GitHub Application.
+
+
+Porter uses a GitHub application to authorize and gain access to your GitHub repositories.
```
abelanger5, line 10: Could you add screenshots and organize this? See this guide for a reference: https://docs.getporter.dev/docs/github

(you can wait to add screenshots until the restyling is done)

internal/repository/gorm/auth.go
```diff
@@ -1087,3 +1087,84 @@ func (repo *AWSIntegrationRepository) DecryptAWSIntegrationData(
 
 	return nil
 }
+
+// GithubAppInstallationRepository implements repository.GithubAppInstallationRepository
+type GithubAppInstallationRepository struct {
+	db *gorm.DB
+}
+
+func NewGithubAppInstallationRepository(db *gorm.DB) repository.GithubAppInstallationRepository {
```
abelanger5, line 1096: Add comments to all exported functions in this file.

server/api/integration_handler.go
```diff
@@ -377,3 +385,173 @@ func (app *App) HandleListProjectOAuthIntegrations(w http.ResponseWriter, r *htt
...
+	url := app.GithubAppConf.AuthCodeURL(state, oauth2.AccessTypeOffline)
+
+	http.Redirect(w, r, url, 302)
+}
+
+func (app *App) HandleGithubAppInstall(w http.ResponseWriter, r *http.Request) {
+	http.Redirect(w, r, fmt.Sprintf("https://github.com/apps/%s/installations/new", app.GithubAppConf.AppName), 302)
+}
+
+type HandleListGithubAppAccessResp struct {
```
abelanger5, line 458: Comment exported function

server/api/integration_handler.go
```diff
@@ -377,3 +385,173 @@ func (app *App) HandleListProjectOAuthIntegrations(w http.ResponseWriter, r *htt
...
+		return
+	}
+
+	// specify access type offline to get a refresh token
+	url := app.GithubAppConf.AuthCodeURL(state, oauth2.AccessTypeOffline)
+
+	http.Redirect(w, r, url, 302)
+}
+
+func (app *App) HandleGithubAppInstall(w http.ResponseWriter, r *http.Request) {
```
abelanger5, line 454: Comment exported function

server/api/integration_handler.go
```diff
@@ -377,3 +385,173 @@ func (app *App) HandleListProjectOAuthIntegrations(w http.ResponseWriter, r *htt
...
+	http.Redirect(w, r, fmt.Sprintf("https://github.com/apps/%s/installations/new", app.GithubAppConf.AppName), 302)
+}
+
+type HandleListGithubAppAccessResp struct {
+	HasAccess bool     `json:"has_access"`
+	LoginName string   `json:"username,omitempty"`
+	Accounts  []string `json:"accounts,omitempty"`
+}
+
+func (app *App) HandleListGithubAppAccess(w http.ResponseWriter, r *http.Request) {
```
abelanger5, line 464: Comment exported function

server/api/oauth_github_handler.go
```diff
@@ -294,3 +296,94 @@ func (app *App) updateProjectFromToken(projectID uint, userID uint, tok *oauth2.
 
 	return err
 }
+
+func (app *App) HandleGithubAppOAuthCallback(w http.ResponseWriter, r *http.Request) {
+	session, err := app.Store.Get(r, app.ServerConf.CookieName)
+
+	fmt.Println("hello...")
```
abelanger5, line 303: Remove debug line

abelanger5 (changes requested):
Great work @igalakhov! Added minor comments, but please make sure to comment all exported functions and variables on the backend. 

Note I haven't fully tested on the dev cluster yet, so will approve once these changes are complete and it's been fully tested.

igalakhov:
Requested changes and docs are now updated - this PR should be more or less done.

server/api/integration_handler.go
```diff
@@ -377,3 +385,177 @@ func (app *App) HandleListProjectOAuthIntegrations(w http.ResponseWriter, r *htt
 		return
 	}
 }
+
+func (app *App) HandleGithubAppEvent(w http.ResponseWriter, r *http.Request) {
+	payload, err := ioutil.ReadAll(r.Body)
```
abelanger5, line 390: I just realized, shouldn't we be checking a payload token somehow? This endpoint is fully exposed and writes the installation id to the DB. It looks like there are ways to generate secret tokens that Github and the app can use.

abelanger5 (changes requested):

abelanger5 (approved):

PR #856, opened 2021-07-04
Environment group merging and previews

### LINKED ISSUE #821: Loading in an environment group completely overwrites existing env vars
# Location

- [x] Browser
- [ ] CLI
- [ ] API

# Motivation
ATM, loading in an environment group into a deployment's Environment tab overwrites the existing env variables completely (e.g. if you had 15 env variables and load in a new env group with just 1 variable, the deployment will only have 1 env variable)

# Requirements
Users should be able to load in an env group without losing existing env vars

# Additional Details

Xetera:
## Pull request type

- [ ] Bugfix
- [x] Feature

## Pull request checklist

When trying to load environment variables, all existing variables are overwritten by the variables in the selected environment group.

current implementation

Closes #821

## What is the new behavior?

- When selecting environment variables, you can view the contents of the variables
- You can see any conflicting variables that are both defined in the current context and incoming group 

image

image

A few missing types were added.

`SaveButton` implementation was tweaked to prevent the longer status texts from overflowing outside the modal padding by adding a `left` position and `justify-content: space-between`.

The modal main content was split in half to allow adding dynamic content without having to vertically resize the modal.

### DIFF
--- a/dashboard/src/main/home/modals/LoadEnvGroupModal.tsx
+++ b/dashboard/src/main/home/modals/LoadEnvGroupModal.tsx
@@ -213,6 +213,7 @@ const GroupEnvPreview = styled.pre`
   margin: 0 0 10px 0;
   white-space: pre-line;
   word-break: break-word;
+  user-select: text;
 `;
 
 const ClashIcon = styled.i`
--- a/dashboard/src/main/home/modals/LoadEnvGroupModal.tsx
+++ b/dashboard/src/main/home/modals/LoadEnvGroupModal.tsx
@@ -9,7 +9,10 @@ import { Context } from "shared/Context";
 import Loading from "components/Loading";
 import SaveButton from "components/SaveButton";
 import { KeyValue } from "components/values-form/KeyValueArray";
-import { EnvGroupData } from "../cluster-dashboard/env-groups/EnvGroup";
+import {
+  EnvGroupData,
+  formattedEnvironmentValue,
+} from "../cluster-dashboard/env-groups/EnvGroup";
 
 type PropsType = {
   namespace: string;
@@ -130,7 +133,9 @@ export default class LoadEnvGroupModal extends Component<PropsType, StateType> {
                 {this.props.existingValues[key] || emptyValue}
               </ClashingKeyValue>
               <ClashingKeyLabel>Replaced by</ClashingKeyLabel>
-              <ClashingKeyValue>{value || emptyValue}</ClashingKeyValue>
+              <ClashingKeyValue>
+                {formattedEnvironmentValue(value) || emptyValue}
+              </ClashingKeyValue>
             </ClashingKeyDefinitions>
             {i !== clashingKeys.length - 1 && <ClashingKeyRowDivider />}
           </li>
@@ -164,7 +169,10 @@ export default class LoadEnvGroupModal extends Component<PropsType, StateType> {
             <SidebarSection>
               <GroupEnvPreview>
                 {Object.entries(this.state.selectedEnvGroup.data)
-                  .map(([key, value]) => `${key}=${value}`)
+                  .map(
+                    ([key, value]) =>
+                      `${key}=${formattedEnvironmentValue(value)}`
+                  )
                   .join("\n")}
               </GroupEnvPreview>
               {clashingKeys.length > 0 && (
--- a/dashboard/src/main/home/modals/LoadEnvGroupModal.tsx
+++ b/dashboard/src/main/home/modals/LoadEnvGroupModal.tsx
@@ -1,5 +1,5 @@
 import React, { Component } from "react";
-import styled from "styled-components";
+import styled, { css } from "styled-components";
 import close from "assets/close.png";
 import sliders from "assets/sliders.svg";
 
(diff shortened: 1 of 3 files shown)


abelanger5 (commented):
Hey @Xetera, thanks so much for this PR! Just tested locally and the core functionality is working great. I requested @jnfrati to take a closer look at the code. 

A few high-level considerations, interested to hear your thoughts on these:
- One edge case here is the "secret" environment variables. These get loaded into the frontend as "dummy" environment variables with the data `PORTERSECRET_<name-of-env-group>`, so the "Replaced by..." text is slightly confusing. Perhaps we should add a case to make it clear that this is a "secret" env variable?
- Perhaps increase the size of the modal slightly? The split modal is super useful with the env group previews for what is getting replaced, but it's starting to feel a bit cramped.

jnfrati (commented):
It looks good to me! Although I would expect @jusrhee to have comments on how it looks like, but we can achieve those on another PR if needed.

Xetera:
I think the env group should be taking up 100% of the width before one is selected so we don't have to spend 50% of the space on a placeholder text.

Also yes now that the modal has more responsibility it should definitely be taking up a little more space. Should I just increase it in to take up more space in pixels or is there a bigger refactoring that should happen here?

abelanger5:
> I think the env group should be taking up 100% of the width before one is selected so we don't have to spend 50% of the space on a placeholder text.

Yes, I agree. 

> Also yes now that the modal has more responsibility it should definitely be taking up a little more space. Should I just increase it in to take up more space in pixels or is there a bigger refactoring that should happen here?

Yep, just increasing the pixel value is fine! I'm guessing the modals will be some of the last items that are refactored, so don't worry about it.

Xetera:
Hopefully this addresses the issues. The preview still looks a bit out of place and kind of ugly, especially when you have real environment variables that end up wrapping but I didn't want to add a syntax highlighting lib without consulting the maintainers.

The `<hr />` between the preview and warnings still feels a bit out of place but I couldn't figure out how to make that work.

image

image

dashboard/src/components/SaveButton.tsx
```diff
@@ -74,17 +74,22 @@ const LoadingGif = styled.img`
   margin-bottom: 0px;
 `;
 
+const StatusContainer = styled.div`
+  flex: 1;
```
abelanger5, line 78: Setting the `StatusContainer` to `flex` causes the status updates on `ExpandedChart` to render completely on the left of the chart:

image

Is `flex` necessary here/perhaps we parameterize where we want the status text to show?

dashboard/src/components/SaveButton.tsx
```diff
@@ -74,17 +74,22 @@ const LoadingGif = styled.img`
...
   font-size: 13px;
   color: #ffffff55;
   margin-right: 25px;
-  padding: 0 10px;
+  flex: 1;
+  padding-inline-end: 10px;
 
   max-width: 500px;
-  white-space: nowrap;
+  white-space: break-spaces;
```
abelanger5, line 92: This has some unintended consequences I think, this is the reason `white-space` is set to `nowrap `: 

image

(our errors aren't normally this long, but there is a way to trigger it by setting the raw Helm values incorrectly on purpose. In this case, I manually set `cpu: 100mASDF` in the helm values tab.)

I'm sure there are other solutions, but the ones that come to mind are: set the maximum height to 2 lines, trim the text that gets rendered for the status, or change `white-space` back to `nowrap` and shorten the error (since the modal is larger now, the longest error is pretty close to 1 line).

abelanger5 (changes requested):
Awesome, it's working great, really cleans up the environment var flow @Xetera 💯

Sorry that I didn't catch these last time, but tweaking the `SaveButton` has some consequences in other parts of the platform. See the review comments.

dashboard/src/main/home/modals/LoadEnvGroupModal.tsx
```diff
@@ -96,7 +102,55 @@ export default class LoadEnvGroupModal extends Component<PropsType, StateType> {
...
+                <ClashIcon className="material-icons">sync_problem</ClashIcon>
+              </ClashIconWrapper>
+              <ClashingKeyExplanation>
+                <b>{key}</b> is defined in both environments
+              </ClashingKeyExplanation>
+            </ClashingKeyTop>
+            <ClashingKeyDefinitions>
+              <ClashingKeyLabel>Old</ClashingKeyLabel>
+              <ClashingKeyValue>
+                {this.props.existingValues[key] || emptyValue}
```
abelanger5, line 137: Shouldn't this use `formattedEnvironmentValue` as well? When testing, the "Old" keys are still showing as `PORTER_SECRET`: 

image

Xetera: Totally, I have no idea how that slipped by :grimacing:

Xetera:
I moved the save button changes to a separate PR (#878) because it was starting to clutter this one. Everything else related to environment merging is ready on my side.

abelanger5 (approved):

abelanger5:
Perfect, tested one last time and merged! Should be on the hosted version within the hour. I'll take a look at the `SaveButton` PR. 

Thanks @Xetera!

PR #861, opened 2021-07-08
Fixed namespace list reference not updating useMemo sorted namespaces

### LINKED ISSUE #855: Remove namespace or update status after deletion
# Description

After deleting a namespace from the cluster dashboard the deleted namespace hangs around or doesn't update status until component refresh.

# Location

- [X] Browser
- [ ] CLI
- [ ] API

# Steps to reproduce

1. Go to namespace management tab of the cluster dashboard
2. Delete a namespace
3. Namespace will persist on UI

# Additional Details

jnfrati:
## Pull request type

- [x] Bugfix
- [ ] Feature

## Pull request checklist

When triggering a delete, the namespace wasn't updating the status, so it seemed like it wasn't updating.

Closes #855 

## What is the new behavior?

Fixed behavior, now when you trigger a delete for a namespace, the status will update to "Terminating" so the user can see that the deletion is in process

The problem was on the WebSocket update event, when we tried to update the list of namespaces, we need to pass the setNamespaces function a new object so it can trigger updates on functions like useMemo.

In this case, I just replace the `return oldNamespaces` with a return `[...oldNamespaces]` as it will return a new reference to the "new" array.

UPDATE:
Also, just to prevent double requests to deletion, I added a condition to not be able to remove a namespace unless it is on Active phase.

### DIFF
--- a/dashboard/src/main/home/cluster-dashboard/dashboard/NamespaceList.tsx
+++ b/dashboard/src/main/home/cluster-dashboard/dashboard/NamespaceList.tsx
@@ -109,7 +109,7 @@ export const NamespaceList: React.FunctionComponent = () => {
             (namespace) => namespace.metadata.name === data.Object.metadata.name
           );
           oldNamespaces.splice(oldNamespaceIndex, 1, data.Object);
-          return oldNamespaces;
+          return [...oldNamespaces];
         });
       }
     };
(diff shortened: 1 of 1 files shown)


jusrhee (approved):
Looks good!

PR #865, opened 2021-07-09
[0.6.0] - Implement authorization module

jnfrati:
## Pull request type

- [ ] Bugfix
- [x] Feature

## Pull request checklist

Although the authorization module already exists, is not implemented anywhere around the application, so even if the user has no access to a resource it will still be able to get into it

## What is the new behavior?

Implemented the authorization module, the user will be allowed or restricted to access some pages/resources/components according to the policy document that their respective role possesses. 

In order to make this PR as quickly as possible I had to reduce some of the implementation features that were planned:
- In this PR we are NOT checking if the user has access, for example to a specific namespace or application, it has full access or no access at all
- It's NOT checking if the user has multiple policy documents, it will be able to handle only one at the time.
- The integration with the API will be resolved in another PR to avoid changes on the integration too.

Also some implementation notes:
- Added a select button in order to switch frontend-mocked roles (this will be removed before merging to master
- Added a useAuth hook to be able to access the isAuthorized function with the policy already populated
- Refactored the withAuth HOC to have the same functionality as the useAuth. It now passes by props the isAuthorized function.
- Tried to make the minor refactoring as possible, but you should probably check out the refactor done on `dashboard/src/main/home/cluster-dashboard/ClusterDashboard.tsx` as it changed the routes implementation

### DIFF
--- a/dashboard/src/components/UnauthorizedPage.tsx
+++ b/dashboard/src/components/UnauthorizedPage.tsx
@@ -1,72 +1,16 @@
-import React, { Component } from "react";
+import React from "react";
 import styled from "styled-components";
 
-type PropsType = {};
-
-type StateType = {};
-
-class UnauthorizedPage extends Component<PropsType, StateType> {
-  state = {};
-
-  render() {
-    return (
-      <StyledUnauthorizedPage>
-        <Mega>
-          401
-          <Inside>You're not authorized to access this page</Inside>
-        </Mega>
-      </StyledUnauthorizedPage>
-    );
-  }
-}
+const UnauthorizedPage: React.FunctionComponent = () => (
+  <StyledUnauthorizedPage>
+    <Mega>
+      401
+      <Inside>You're not authorized to access this page</Inside>
+    </Mega>
+  </StyledUnauthorizedPage>
+);
 
 export default UnauthorizedPage;
-
-const Splitter = styled.div`
-  margin: 0 20px;
-  font-size: 27px;
-  font-weight: 200;
-  color: #ffffff15;
-`;
-
-const Flex = styled.div`
-  display: flex;
-  align-items: center;
-  justify-content: center;
-`;
-
-const Helper = styled.div`
-  font-size: 15px;
-  max-width: 550px;
-  margin-right: -50px;
-`;
-
-const BackButton = styled.div`
-  display: flex;
-  align-items: center;
-  justify-content: space-between;
-  cursor: pointer;
-  font-size: 13px;
-  height: 35px;
-  padding: 5px 16px;
-  padding-right: 15px;
-  border: 1px solid #ffffff55;
-  border-radius: 100px;
-  width: ${(props: { width: string }) => props.width};
-  color: white;
-  background: #ffffff11;
-
-  :hover {
-    background: #ffffff22;
-  }
-
-  > i {
-    color: white;
-    font-size: 16px;
-    margin-right: 6px;
-    margin-left: -2px;
-  }
-`;
 
 const StyledUnauthorizedPage = styled.div`
   font-family: "Work Sans", sans-serif;
--- a/dashboard/src/components/UnauthorizedPage.tsx
+++ b/dashboard/src/components/UnauthorizedPage.tsx
@@ -1,0 +1,115 @@
+import React, { Component } from "react";
+import styled from "styled-components";
+
+type PropsType = {};
+
+type StateType = {};
+
+class UnauthorizedPage extends Component<PropsType, StateType> {
+  state = {};
+
+  render() {
+    return (
+      <StyledUnauthorizedPage>
+        <Mega>
+          401
+          <Inside>You're not authorized to access this page</Inside>
+        </Mega>
+      </StyledUnauthorizedPage>
+    );
+  }
+}
+
+export default UnauthorizedPage;
+
+const Splitter = styled.div`
+  margin: 0 20px;
+  font-size: 27px;
+  font-weight: 200;
+  color: #ffffff15;
+`;
+
+const Flex = styled.div`
+  display: flex;
+  align-items: center;
+  justify-content: center;
+`;
+
+const Helper = styled.div`
+  font-size: 15px;
+  max-width: 550px;
+  margin-right: -50px;
+`;
+
+const BackButton = styled.div`
+  display: flex;
+  align-items: center;
+  justify-content: space-between;
+  cursor: pointer;
+  font-size: 13px;
+  height: 35px;
+  padding: 5px 16px;
+  padding-right: 15px;
+  border: 1px solid #ffffff55;
+  border-radius: 100px;
+  width: ${(props: { width: string }) => props.width};
+  color: white;
+  background: #ffffff11;
+
+  :hover {
+    background: #ffffff22;
+  }
+
+  > i {
+    color: white;
+    font-size: 16px;
+    margin-right: 6px;
+    margin-left: -2px;
+  }
+`;
+
+const StyledUnauthorizedPage = styled.div`
+  font-family: "Work Sans", sans-serif;
+  color: #6f6f6f;
+  font-size: 16px;
+  user-select: none;
+  padding-bottom: 20px;
+  width: 100%;
+  height: 100%;
+  display: flex;
+  flex-direction: column;
+  align-items: center;
+  justify-content: center;
+`;
+
+const Mega = styled.div`
+  font-size: 200px;
+  color: #ffffff06;
+  position: relative;
+  font-weight: bold;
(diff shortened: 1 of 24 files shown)


dashboard/src/components/UnauthorizedPage.tsx
```diff
@@ -0,0 +1,115 @@
+import React, { Component } from "react";
+import styled from "styled-components";
+
+type PropsType = {};
+
+type StateType = {};
+
+class UnauthorizedPage extends Component<PropsType, StateType> {
```
igalakhov, line 8: Probably reasonable to have this as a functional component? That saves having to write a PropsType and StateType too, especially if they're empty.

dashboard/src/main/home/cluster-dashboard/env-groups/ExpandedEnvGroup.tsx
```diff
@@ -170,32 +191,44 @@ export default class ExpandedEnvGroup extends Component<PropsType, StateType> {
                 setValues={(x: any) => this.setState({ envVariables: x })}
                 fileUpload={true}
                 secretOption={true}
+                disabled={
+                  !this.props.isAuthorized("env_group", "", [
```
igalakhov, line 195: Semantically, I think it would be cleaner if we also provided an `isNotAuthorized` function (since it would make things that are disabled due to some auth reason easier to read).

dashboard/src/main/home/cluster-dashboard/dashboard/NamespaceList.tsx
```diff
@@ -68,6 +69,8 @@ export const NamespaceList: React.FunctionComponent = () => {
     setCurrentModal("DeleteNamespaceModal", namespace);
   };
 
+  const [isAuthorized] = useAuth();
```
igalakhov, line 72: Probably would be cleaner if `useAuth()` returned a dictionary instead of an array (especially if we wanted to have the hook provide more functionality later).

dashboard/src/main/home/cluster-dashboard/env-groups/ExpandedEnvGroup.tsx
```diff
@@ -170,32 +191,44 @@ export default class ExpandedEnvGroup extends Component<PropsType, StateType> {
...
+                }
               />
             </InnerWrapper>
-            <SaveButton
-              text="Update"
-              onClick={() => this.handleUpdateValues()}
-              status={this.state.saveValuesStatus}
-              makeFlush={true}
-            />
+            {this.props.isAuthorized("env_group", "", ["get", "update"]) && (
```
igalakhov, line 204: Also, this `{authorization condition && JSX}` pattern is something that pops up a lot. Maybe later on it would make sense to have a component like `<AuthorizationRequired for={"env_group"}, perms={["get, "update"]}>` (and similarly for authorization not being required) that renders the children conditionally.

dashboard/src/main/home/cluster-dashboard/env-groups/ExpandedEnvGroup.tsx
```diff
@@ -246,6 +279,10 @@ export default class ExpandedEnvGroup extends Component<PropsType, StateType> {
     }
   };
 
+  getTabOptions = () => {
```
igalakhov, line 282: What's the point of this function?

dashboard/src/main/home/navbar/Navbar.tsx
```diff
@@ -49,6 +54,22 @@ export default class Navbar extends Component<PropsType, StateType> {
   render() {
     return (
       <StyledNavbar>
+        <AuthContext.Consumer>
```
igalakhov, line 57: Just wondering, but is the reason for this implementation instead of attaching the context to the component because this is a class component and therefore can't have multiple contexts?

dashboard/src/shared/auth/AuthContext.tsx
```diff
@@ -17,24 +20,64 @@ const AuthProvider: React.FC = ({ children }) => {
...
-          ADMIN_POLICY_MOCK,
-          POLICY_HIERARCHY_TREE,
-          ADMIN_POLICY_MOCK.scope,
-          ADMIN_POLICY_MOCK.verbs
-        )
-      );
+    setPolicy("admin");
+  }, []);
+
+  // This is just for test case, should be deleted before merged with master
```
igalakhov, line 43: Should this be deleted?

dashboard/src/shared/auth/RouteGuard.tsx
```diff
@@ -26,7 +28,7 @@ const GuardedRoute: React.FC<RouteProps & GuardedRouteProps> = ({
     <Route
       {...rest}
       render={(props) =>
-        auth === true ? <Component {...props} /> : <Redirect to="/" />
+        auth ? children || <Component {...props} /> : <UnauthorizedPage />
```
igalakhov, line 31: this ternary statement is a bit visually unclear, maybe add parenthesis

dashboard/src/shared/auth/useAuth.ts
```diff
@@ -0,0 +1,21 @@
...
+  const isAuth = useCallback(
+    (
+      scope: ScopeType,
+      resource: string | string[],
+      verb: Verbs | Array<Verbs>
+    ) => isAuthorized(authContext.currentPolicy, scope, resource, verb),
+    [authContext.currentPolicy]
+  );
+
+  return [isAuth];
```
igalakhov, line 18: Same as the previous comment, but maybe it would make sense to return an object instead of an array to make future features easier

igalakhov (commented):
Looks good @jnfrati! Added some minor comments and some thoughts about `useAuth` for reference. Also, not super sure on what components should have what permissions (as I'm not too familiar with those parts of the frontend), so someone should review that as well.

dashboard/src/main/home/navbar/Navbar.tsx
jnfrati: This is just temporary, the whole section will be removed when merging to master on the implement-rbac PR. This was just made to be able to have a simpler way to change the current role on the context without modifying the hooks & HOCs nor their typing.
Although a class component cannot have multiple context, it is possible via the withAuth HOC implemented on the auth module

dashboard/src/shared/auth/AuthContext.tsx
jnfrati: This will be deleted on the implement-rbac PR

dashboard/src/main/home/cluster-dashboard/env-groups/ExpandedEnvGroup.tsx
jnfrati: Mmm I'm not quite sure about this one, although it makes sense, it's also a really common practice to avoid duplicated functions like isAuthorized and isNotAuthorized and just use a NOT operator.

igalakhov: That's fair then, was just wondering

dashboard/src/main/home/cluster-dashboard/dashboard/NamespaceList.tsx
jnfrati: As the useAuth rn only returns one function it didn't make sense to make it a dictionary, as an array is easier to rename the returned function. Probably it would make sense in the case we want to add some extra functionality but rn I don't see too much improvement on changing this.

dashboard/src/main/home/cluster-dashboard/env-groups/ExpandedEnvGroup.tsx
jnfrati: I have mixed feelings about the AuthorizationRequired component, on one hand, is actually a really good way to clean the JSX logic, but on the other one, I feel that it can be annoying when trying to mix up logic.

dashboard/src/main/home/cluster-dashboard/dashboard/NamespaceList.tsx
igalakhov: Yep fair

dashboard/src/shared/auth/RouteGuard.tsx
jnfrati: Yep, fixing that rn!

dashboard/src/main/home/cluster-dashboard/env-groups/ExpandedEnvGroup.tsx
jnfrati: Oh, this is not supposed to be there, sorry!

PR #891, opened 2021-07-14
Allow renaming an environment group

### LINKED ISSUE #841: Renaming an Environment Group
# Location

- [x] Browser
- [ ] CLI
- [x] API

# Motivation
# Requirements
Users should be able to rename an existing environment group

# Additional Details

anukul:
## Pull request type

- [ ] Bugfix
- [x] Feature

## Pull request checklist

- [ ] If it's a backend change, tests for the changes have been added and `go test ./...` runs successfully from the root folder

Draft PR, can add tests / docs if this makes sense.

Issue Number: Closes #841 

## What is the new behavior?

https://user-images.githubusercontent.com/44864521/125695461-2feb9fc0-8656-48d9-91bd-e1add1fab868.mov

- Implemented a rough design for draft PR, taken from the Delete option.  
Success (close modal) and Failure (do nothing) behaviour is also similar.

### DIFF
--- a/dashboard/src/main/home/cluster-dashboard/env-groups/ExpandedEnvGroup.tsx
+++ b/dashboard/src/main/home/cluster-dashboard/env-groups/ExpandedEnvGroup.tsx
@@ -6,6 +6,7 @@ import _ from "lodash";
 
 import { ChartType, StorageType, ClusterType } from "shared/types";
 import { Context } from "shared/Context";
+import { isAlphanumeric } from "shared/common";
 import api from "shared/api";
 
 import SaveButton from "components/SaveButton";
@@ -15,6 +16,7 @@ import TabRegion from "components/TabRegion";
 import EnvGroupArray, { KeyValueType } from "./EnvGroupArray";
 import Heading from "components/values-form/Heading";
 import Helper from "components/values-form/Helper";
+import InputRow from "components/values-form/InputRow";
 import { withAuth, WithAuthProps } from "shared/auth/AuthorizationHoc";
 
 type PropsType = WithAuthProps & {
@@ -30,6 +32,7 @@ type StateType = {
   showDeleteOverlay: boolean;
   deleting: boolean;
   saveValuesStatus: string | null;
+  envGroupName: string;
   envVariables: KeyValueType[];
   tabOptions: { value: string; label: string }[];
 };
@@ -46,6 +49,7 @@ class ExpandedEnvGroup extends Component<PropsType, StateType> {
     showDeleteOverlay: false,
     deleting: false,
     saveValuesStatus: null as string | null,
+    envGroupName: null as string,
     envVariables: [] as KeyValueType[],
     tabOptions: [
       { value: "environment", label: "Environment Variables" },
@@ -54,6 +58,7 @@ class ExpandedEnvGroup extends Component<PropsType, StateType> {
   };
 
   componentDidMount() {
+    const envGroupName = this.props.envGroup.metadata.name;
     // parse env group props into values type
     let envVariables = [] as KeyValueType[];
     let envGroupData = this.props.envGroup.data;
@@ -68,7 +73,7 @@ class ExpandedEnvGroup extends Component<PropsType, StateType> {
       });
     }
 
-    this.setState({ envVariables });
+    this.setState({ envGroupName, envVariables });
 
     // Filter the settings tab options as for now it only shows the delete button.
     // In a future this should be removed and return to a constant if we want to show data
@@ -85,6 +90,30 @@ class ExpandedEnvGroup extends Component<PropsType, StateType> {
       };
     });
   }
+
+  handleRename = () => {
+    const { envGroup } = this.props;
+    const name = envGroup.metadata.name;
+    const namespace = envGroup.metadata.namespace;
+    const newName = this.state.envGroupName;
+
+    api
+      .renameConfigMap(
+        "<token>",
+        {
+          name,
+          namespace,
+          new_name: newName,
+        },
+        {
+          id: this.context.currentProject.id,
+          cluster_id: this.props.currentCluster.id,
+        }
+      )
+      .then((res) => {
+        this.props.closeExpanded();
+      });
+  };
 
   handleUpdateValues = () => {
     let { envGroup } = this.props;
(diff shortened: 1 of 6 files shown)


anukul:
cc @abelanger5

abelanger5:
Looks really promising @anukul! My apologies for not reviewing the draft PR earlier. A few notes:
1. Could you outline how you intend to avoid data loss if the server exits while serving the request? It seems that the way this is currently implemented, the configmap + secret are deleted before they are created again. 
2. Have you tested that this properly mounts secret env variables into a container? It might be helpful to document how the  secret data is read/mounted into the container (perhaps we should create a doc in `/docs/developing` with a description of this architecture). Let's say I have a "secret environment variable" named `TEST` in an environment group named `testing`. The ConfigMap will have the following data:

```yaml
metadata:
  name: testing
...
data:
  TEST: PORTERSECRET_testing
```

While the secret will look like:

```yaml
metadata:
  name: testing
...
data:
  TEST: <secret-data>
```

So the way that this gets mounted into the container is that the Helm chart [parses](https://github.com/porter-dev/porter-charts/blob/79f87670d8d836a90424129e10e350102b80f9ac/applications/web/templates/deployment.yaml#L92) the `PORTERSECRET` configmap values and references them as a secret with the following spec:

```yaml
... # pod spec
env:
  name: TEST
  valueFrom:
    secretKeyRef:
      name: testing
      key: TEST
```

All that said, the implementation looks like it will work with secret env variables, since the line `configMap.EnvVariables[key] = fmt.Sprintf("PORTERSECRET_%s", configMap.Name)` will overwrite the keys that are set to the "incorrect" secret value. But I'm not sure if there's a cleaner way to override these values.

3. It looks like the CRLF endings got replaced with LF endings in the file `/internal/kubernetes/agent.go`, creating a really large diff. Could you revert this? It makes the code a bit harder to review. While we should be using LF endings, this should be in a different PR. 

Once you'd like me to test it and review the code, just mark the PR as "ready to review". Thanks!

abelanger5:
Hi @anukul, just checking in here! Let me know if there's anything else you need to complete this PR 👍

anukul:
@abelanger5 wrt the notes,

1. I have now moved the delete operation after the create operation. It's not "atomic", but no data loss as such. Is this acceptable behavior?
2. My local setup is a bit hacky so I haven't configured it to the point where I can actually run apps yet.
Either way, I don't think this PR involves overriding any values since it deletes the old secret and creates a new one, with the new name. I did test this part, so I believe it should work correctly since I haven't modified anything else.
3. Sure, I see it has now been updated, so I've rebased on master.

server/api/k8s_handler.go
```diff
@@ -559,6 +565,92 @@ func (app *App) HandleUpdateConfigMap(w http.ResponseWriter, r *http.Request) {
...
+	}
+
+	newConfigMap := &forms.ConfigMapForm{
+		Name:               renameConfigMapForm.NewName,
+		Namespace:          configMap.Namespace,
+		EnvVariables:       configMap.Data,
+		SecretEnvVariables: decodedSecretData,
+	}
+
+	if err := createConfigMap(agent, newConfigMap); err != nil {
```
abelanger5, line 635: Before this function call, we should check if the new name of the config map is different from the old name, otherwise the backend will throw a `500` error with the following error log:

```
{"level":"warn","error":"secrets \"test-env-rename-2\" already exists","errExt":"{\"code\":500,\"errors\":[\"internal server error\"]}","time":"2021-07-29T18:52:22Z"}
```

(ideally this is checked from the frontend as well)

dashboard/src/main/home/cluster-dashboard/env-groups/ExpandedEnvGroup.tsx
```diff
@@ -86,6 +91,30 @@ class ExpandedEnvGroup extends Component<PropsType, StateType> {
...
+          namespace,
+          new_name: newName,
+        },
+        {
+          id: this.context.currentProject.id,
+          cluster_id: this.props.currentCluster.id,
+        }
+      )
+      .then((res) => {
+        this.props.closeExpanded();
```
abelanger5, line 114: Instead of closing the expanded env group here, it seems more intuitive to just rename the expanded env group and re-pull the configmap data. What do you think?

abelanger5 (changes requested):
Hey @anukul, just two minor comments, but otherwise it's working really well! 

> I have now moved the delete operation after the create operation. It's not "atomic", but no data loss as such. Is this acceptable behavior?

Yes, this seems acceptable, and I'm not sure how we would make this an atomic...

server/api/k8s_handler.go
anukul: Oh right.. thanks, fixed!

dashboard/src/main/home/cluster-dashboard/env-groups/ExpandedEnvGroup.tsx
anukul: I agree, that makes more sense.
I've pushed a commit for this, it involved some refactoring to store env group in state.

dashboard/src/main/home/cluster-dashboard/env-groups/ExpandedEnvGroup.tsx
```diff
@@ -49,31 +55,48 @@ class ExpandedEnvGroup extends Component<PropsType, StateType> {
...
       { value: "settings", label: "Settings" },
     ],
+    newEnvGroupName: null as string,
   };
 
-  componentDidMount() {
-    const envGroupName = this.props.envGroup.metadata.name;
+  populateEnvGroup = (envGroup: any) => {
+    const {
+      metadata: { name, creationTimestamp: timestamp },
```
anukul, line 72: Side note - the `creationTimestamp` field is used to display "Last updated ...", which does not change when the user updates config map variables.

abelanger5: Ah good catch -- I've filed an issue (#946) which we can tackle in a later PR.

abelanger5 (approved):
Thanks so much @anukul! Approved and merged into `master`, should be on the hosted version later today 👍

dashboard/src/main/home/cluster-dashboard/env-groups/ExpandedEnvGroup.tsx
abelanger5: The state refactor is cleaner as well, thanks!

PR #894, opened 2021-07-15
[0.6.0] [Frontend, API] Trying to improve Expanded Chart performance

igalakhov:
## Pull request type

- [ ] Bugfix
- [ ] Feature

## Pull request checklist

Expanded Chart makes a lot of requests to the backend, a lot of which are duplicated:

image

Many of these take a long time to happen (for example, the history endpoint in the image above takes quite a while to load and is requested 3 times) and lots of requests propagate state updates upwards which also causes problems.

## What is the new behavior?

The idea is to try to reduce the number of requests these components make by creating a context that handles some part of these requests.

### DIFF
--- a/dashboard/src/main/home/cluster-dashboard/expanded-chart/ExpandedChart.tsx
+++ b/dashboard/src/main/home/cluster-dashboard/expanded-chart/ExpandedChart.tsx
@@ -75,10 +75,9 @@ const ExpandedChart: React.FC<Props> = (props) => {
   const [url, setUrl] = useState<string>(null);
   const [showDeleteOverlay, setShowDeleteOverlay] = useState<boolean>(false);
   const [deleting, setDeleting] = useState<boolean>(false);
-  const [imageIsPlaceholder, setImageIsPlaceholer] = useState<boolean | null>(
-    null
-  );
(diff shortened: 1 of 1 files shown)


jusrhee:
Definitely substantially faster already. Here are the issues I've been able to identify so far (some of which you've already noted):

- All fields are read-only and config can't be changed. Something seems to be messing with basic RBAC (maybe worth asking @jnfrati to take a look?)
- Issues selecting old revisions: tabs become yellow after one click but "active" tabs (settings, metrics, status) only disappear/reappear after double-click. Revision selector itself (the expanded table) also doesn't seem to indicate which revision is selected.
- Reverting to a past revision still seems to work, but the newly deployed revision should be auto-selected after rollback

jusrhee:
Looks good. One thing I did notice though is that the internal URI / URL of the web service does seem to take a fair bit longer to load in now (always loads in with the tab contents even though this doesn't seem to happen on prod). Hard to benchmark though.

jusrhee:
Also, for applications being deployed the status tab flashes for a second first before this placeholder comes up which doesn't happen on the older version of the expandedchart:
image

jusrhee (approved):
Tested locally - looks good!

PR #895, opened 2021-07-15
[0.6.0] Github Org Access Part 2

igalakhov:
## Pull request type

- [ ] Bugfix
- [x] Feature

## Pull request checklist

Repo search and access old OAuth integration.

## What is the new behavior?

Repo search and access now uses the new GitHub app which should be able to act standalone for all operations we want it to do.


jusrhee (approved):
Tested on dev - looks good

abelanger5 (changes requested):
Looks pretty good @igalakhov! Added some minor comments. Will give it one more run-through when the access token/refresh token issue that we discussed are resolved.

server/api/oauth_github_handler.go
```diff
@@ -323,15 +323,6 @@ func (app *App) HandleGithubAppOAuthCallback(w http.ResponseWriter, r *http.Requ
 		return
 	}
 
-	if r.URL.Query().Get("state") != session.Values["state"] {
```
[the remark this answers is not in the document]
igalakhov: To handle installs that happen from other websites (for instance, us linking the app to someone or someone wanting to install it from GitHub directly).

server/api/release_handler.go
```diff
@@ -939,29 +940,32 @@ func (app *App) HandleUpgradeRelease(w http.ResponseWriter, r *http.Request) {
...
-					ProjectID:      uint(projID),
-					ReleaseName:    name,
-					GitBranch:      gitAction.GitBranch,
-					DockerFilePath: gitAction.DockerfilePath,
-					FolderPath:     gitAction.FolderPath,
-					ImageRepoURL:   gitAction.ImageRepoURI,
-					BuildEnv:       cEnv.Container.Env.Normal,
+					ServerURL:            app.ServerConf.ServerURL,
+					GitIntegration:       gr,
+					GithubInstallationID: gitAction.GitRepoID,
```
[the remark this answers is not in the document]
igalakhov: Yeah, I wasn't really sure what to do but GitRepoID now refers to Installation IDs instead of Repo IDs

server/router/router.go
```diff
@@ -1129,14 +1129,13 @@ func New(a *api.App) *chi.Mux {
 				),
 			)
 
-			r.Method(
+			r.Method( // this endpoint isn't going to be used anymore, should it get deleted?
```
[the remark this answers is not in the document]
igalakhov: Done. Frontend repo list is still technically there but isn't accessible to frontend user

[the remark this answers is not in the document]
igalakhov: Actually, removed from frontend too

server/api/integration_handler.go
```diff
@@ -581,9 +583,30 @@ func (app *App) getGithubUserTokenFromRequest(r *http.Request) (*oauth2.Token, e
 		return nil, err
 	}
 
+	_, _, err = oauth.GetAccessToken(oauthInt.AccessToken, oauthInt.RefreshToken, &oauth2.Config{
```
abelanger5, line 586: You may want to change the embedded `oauth2.Config` to a pointer and then pass `app.GithubAppConf.Config` as the `oauth2.Config` param, so you're not instantiating this each time.

server/middleware/auth.go
```diff
@@ -394,53 +399,110 @@ func (auth *Auth) DoesUserHaveRegistryAccess(
...
+// note that this makes a github API request, but the endpoint is fast so this doesn't add
+// much overhead
+func (auth *Auth) DoesUserHaveGitInstallationAccess(
 	next http.Handler,
-	projLoc IDLocation,
 	gitRepoLoc IDLocation,
 ) http.Handler {
 	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
-		grID, err := findGitRepoIDInRequest(r, gitRepoLoc)
+		// TODO: needs to use new github integration implementation
```
abelanger5, line 411: Remove TODO statement?

server/middleware/auth.go
```diff
@@ -394,53 +399,110 @@ func (auth *Auth) DoesUserHaveRegistryAccess(
...
+		oauthInt, err := auth.repo.GithubAppOAuthIntegration.ReadGithubAppOauthIntegration(user.GithubAppIntegrationID)
 
 		if err != nil {
-			http.Error(w, http.StatusText(http.StatusInternalServerError), http.StatusInternalServerError)
+			http.Error(w, http.StatusText(http.StatusForbidden), http.StatusForbidden)
 			return
 		}
 
-		doesExist := false
+		client := github.NewClient(auth.GithubProjectConf.Client(oauth2.NoContext, &oauth2.Token{
```
abelanger5, line 457: Should this be using `auth.GithubProjectConf` since this method is verifying that the user has access to the Github installation id?

abelanger5 (changes requested):

server/api/release_handler.go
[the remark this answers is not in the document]
abelanger5: Perhaps you can rename `GitRepoID` -> `GithubInstallationID`?

internal/integrations/ci/actions/actions.go
```diff
@@ -197,22 +200,40 @@ func (g *GithubActions) GetGithubActionYAML() ([]byte, error) {
...
 
-	tok := &oauth2.Token{
-		AccessToken:  string(oauthInt.AccessToken),
-		RefreshToken: string(oauthInt.RefreshToken),
-		TokenType:    "Bearer",
+		if err != nil {
+			return nil, err
+		}
+
+		tok := &oauth2.Token{
```
abelanger5, line 214: I might be missing something, but shouldn't this also be using `GetAccessToken`?

internal/oauth/config.go
```diff
@@ -90,16 +94,31 @@ func CreateRandomState() string {
 	return state
 }
 
+// MakeUpdateOAuthIntegrationTokenFunction creates a function to be passed to GetAccessToken that updates the OauthIntegration
+// if it needs to be updated
+func MakeUpdateOAuthIntegrationTokenFunction(o *integrations.OAuthIntegration, repo repository.Repository) func(accessToken []byte, refreshToken []byte, expiry time.Time) error {
```
abelanger5, line 99: Would be good to split this over multiple lines.

internal/oauth/config.go
```diff
@@ -90,16 +94,31 @@ func CreateRandomState() string {
...
 	conf *oauth2.Config,
-	repo repository.Repository,
+	updateToken func(accessToken []byte, refreshToken []byte, expiry time.Time) error,
 ) (string, *time.Time, error) {
 	tokSource := conf.TokenSource(context.TODO(), &oauth2.Token{
-		AccessToken:  string(o.AccessToken),
-		RefreshToken: string(o.RefreshToken),
+		AccessToken:  string(AccessToken),
+		RefreshToken: string(RefreshToken),
 		TokenType:    "Bearer",
```
abelanger5, line 122: I think you need to add the expiry here. Docs for `TokenSource` mention that if the expiry is the zero-value, refresh token isn't used at all.

abelanger5 (changes requested):

server/api/integration_handler.go
igalakhov: Ah, didn't know that was a thing

server/middleware/auth.go
igalakhov: Yeah, it should (no error happened bc both use the same endpoint to auth)

abelanger5 (approved):

PR #899, opened 2021-07-19
[0.6.0] Refactor segment implementation

jnfrati:
## Pull request type

- [ ] Bugfix
- [ ] Feature
- [x] Other (please describe): Refactor

## Pull request checklist

It was difficult to track the different ways that we were implementing segment as it was all over the API instead of having a centralized way of doing it

## What is the new behavior?

Created a new analytics package that exports a function to initialize a segment client with helper functions to make tracking more simpler and centralized


internal/analytics/segment.go
```diff
@@ -0,0 +1,64 @@
...
+	Identify(userId string, userEmail string, isGithub bool) error
+	Track(userId string, event string, properties segment.Properties) error
+}
+
+type AnalyticsSegment struct {
+	segment.Client
+	isEnabled bool
+}
+
+func Initialize(segmentClientKey string) AnalyticsSegmentClient {
```
abelanger5, line 19: Should have comments and called `NewAnalyticsSegment`

internal/analytics/segment.go
```diff
@@ -0,0 +1,64 @@
...
+
+	return &AnalyticsSegment{
+		isEnabled: false,
+	}
+}
+
+func (c *AnalyticsSegment) Identify(userId string, userEmail string, isGithub bool) error {
+
+	if !c.isEnabled {
+		fmt.Println("Analytics not enabled")
```
abelanger5, line 38: You should pass in a Logger interface and use that instead of `fmt...`

internal/analytics/segment.go
```diff
@@ -0,0 +1,64 @@
+package analytics
```
abelanger5, line 1: You should declare the precise analytics that are being tracked in this package, perhaps in a file called `analytics.go`? For example, instead of `segment.NewProperties().Set("email", user.Email)`, have an analytics type called `User` which has a method that returns `id` and a method that returns a `segment.Property` (or whatever the struct is called).

abelanger5 (changes requested):

internal/analytics/identifiers.go
```diff
@@ -0,0 +1,46 @@
+package analytics
+
+import (
+	"fmt"
+
+	"github.com/porter-dev/porter/internal/models"
+	segment "gopkg.in/segmentio/analytics-go.v3"
+)
+
+type SegmentIdentifier interface {
```
abelanger5, line 10: Comment all exported interfaces, structs, and methods.

internal/analytics/segment.go
```diff
@@ -0,0 +1,83 @@
...
+
+type AnalyticsSegment struct {
+	segment.Client
+	isEnabled bool
+	logger    *logger.Logger
+}
+
+//
+/*
+	Initialize the segment client and return a superset of it, the AnalyticsSegmentClient will handle cases when
```
abelanger5, line 21:
```suggestion
// Initialize the segment client and return a superset of it, the AnalyticsSegmentClient will handle cases when
// the segment client failed on initialization or not enabled
```

Can you modify these comments to keep the style consistent with the other parts of the code?

internal/analytics/tracks.go
```diff
@@ -0,0 +1,64 @@
...
+func (t *SegmentNewUserTrack) getProperties() segment.Properties {
+	return segment.NewProperties().Set("email", t.userEmail)
+}
+
+type SegmentRedeployViaWebhookTrack struct {
+	userId     string
+	repository interface{}
+}
+
+func CreateSegmentRedeployViaWebhookTrack(userId string, repository interface{}) *SegmentRedeployViaWebhookTrack {
```
abelanger5, line 47: Can you document what the `repository` argument is? Also, i'm not sure why this needs to be an `interface{}`. AFAIK the repository argument is the image repository URI, which is always a string.

internal/analytics/README.md
```diff
@@ -0,0 +1,45 @@
+# How the analytics package works
```
abelanger5, line 1: Perhaps we want to make this a `doc.go` file? This will automatically render on the `pkg.go.dev` site. This is suggested by: https://blog.golang.org/godoc, and it links to the example: https://golang.org/src/encoding/gob/doc.go. 

If we want to keep this as a markdown file, I think it should be moved to `/docs/developing` since I'm not sure we want to keep markdown files in the source folders. These might get accidentally mounted in a Docker container or added as part of the build.

abelanger5 (changes requested):
Looks great @jnfrati, that's exactly what I had in mind. 

I added some minor comments, but this definitely cleans up the handlers.

internal/analytics/tracks.go
jnfrati: Didn't exactly knew what was coming there, so I ended up setting the same type as [the repository on the usage file](https://github.com/porter-dev/porter/blob/master/server/api/release_handler.go#L1096), just to make sure I ended up using the `interface{}` but if you say it should be a string I can update that really quick 👍

internal/analytics/identifiers.go
jnfrati: As the only things used outside the package are the struct constructors and the functions on segment.go, I wasn't sure if I should leave this exported. What do you think? Either way I'll add proper comments to the constructor so we can know what those metrics are used for.

abelanger5: Should other packages call any of the methods in the `SegmentIdentifier` interface? If not, this can be unexported.

internal/analytics/tracks.go
abelanger5: Yep, sounds good. I believe it should always be a string.

abelanger5 (approved):
Awesome work @jnfrati!

Issue #892, opened 2021-07-15
Form tabs order changes on every load


sunguroku:
# Description

# Location

- [x] Browser
- [ ] CLI
- [ ] API

# Steps to reproduce

1. Open `ExpandedChart` multiple times - order of tabs changes across reloads (namely status/metrics vs. the tabs rendered via `form.yaml`)

# Additional Details

PR #908, opened 2021-07-22
`ExpandedChart` and Github App -> staging

abelanger5:


sunguroku:
**what works:**
- Checked that reverting, upgrading, persistence of values across tabs are working as before. 
- ExpandedChart loading speed is snappier, but main offender seems to be the status tab - should be resolved with eventual logs pagination. We should prob separate the revision call so that it happens only when the revision section is opened
- Tested backwards compatibility. Environment variable changes on an existing deployment properly update the github secret even after switching to the new github app.

**Issues:**
- If a user has been added to the project and tries to link up a github repository that already has the porter app installed, there is no way to do so without installing the app into another organization just to update the settings. This is a common scenario: if one person in the org links up a GH repo and invites other people, those people will not be able to launch anything.

A few issues that aren't showstoppers:
- Order of tabs on ExpandedChart still shuffles around (#892)
- Guessing it was intentional, but not sure if pulling out the github integration settings page from integrations tab and putting strictly into personal settings (even if it's scoped to personal) is optimal.

PR #943, opened 2021-07-30
Segment infra metrics

sunguroku:
## Pull request type

- [ ] Bugfix
- [ ] Feature
- [x] Other (please describe): telemetry for cluster events

## Pull request checklist

## What is the new behavior?
sends cluster provisioning, destroy, connection events to segment.

### DIFF
--- a/internal/analytics/track_events.go
+++ b/internal/analytics/track_events.go
@@ -5,4 +5,5 @@ type SegmentEvent string
 const (
 	NewUser            SegmentEvent = "New User"
 	RedeployViaWebhook SegmentEvent = "Triggered Re-deploy via Webhook"
+	NewClusterEvent			SegmentEvent = "New Cluster Event"
 )
--- a/internal/analytics/track_events.go
+++ b/internal/analytics/track_events.go
@@ -5,5 +5,5 @@ type SegmentEvent string
 const (
 	NewUser            SegmentEvent = "New User"
 	RedeployViaWebhook SegmentEvent = "Triggered Re-deploy via Webhook"
-	NewClusterEvent			SegmentEvent = "New Cluster Event"
+	NewClusterEvent    SegmentEvent = "New Cluster Event"
 )
--- a/internal/analytics/tracks.go
+++ b/internal/analytics/tracks.go
@@ -66,3 +66,35 @@ func (t *segmentRedeployViaWebhookTrack) getEvent() SegmentEvent {
 func (t *segmentRedeployViaWebhookTrack) getProperties() segment.Properties {
 	return segment.NewProperties().Set("repository", t.repository)
 }
+
+type segmentNewClusterEventTrack struct {
+	userId    string
+	projId string
+	clusterName string
+	clusterType string // EKS, DOKS, or GKE
+	eventType string // connected, provisioned, or destroyed
+}
+
+// Constructor for track of type "New Cluster Event"
+// tracks whenever a cluster is newly provisioned, connected, or destroyed.
+func CreateSegmentNewClusterEvent(userId string, projId string, clusterName string, clusterType string, eventType string) *segmentNewClusterEventTrack {
+	return &segmentNewClusterEventTrack{
+		userId:     userId,
+		projId: projId,
+		clusterName: clusterName,
+		clusterType: clusterType,
+		eventType: eventType, 
+	}
+}
+
+func (t *segmentNewClusterEventTrack) getUserId() string {
+	return t.userId
+}
+
+func (t *segmentNewClusterEventTrack) getEvent() SegmentEvent {
+	return NewClusterEvent
+}
+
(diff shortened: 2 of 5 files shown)


internal/analytics/tracks.go
```diff
@@ -66,3 +66,35 @@ func (t *segmentRedeployViaWebhookTrack) getEvent() SegmentEvent {
...
+
+type segmentNewClusterEventTrack struct {
+	userId    string
+	projId string
+	clusterName string
+	clusterType string // EKS, DOKS, or GKE
+	eventType string // connected, provisioned, or destroyed
+}
+
+// Constructor for track of type "New Cluster Event"
```
abelanger5, line 78:
```suggestion
// CreateSegmentNewClusterEvent tracks whenever...
```

Exported methods should have comments which begin with the method name.

internal/analytics/track_events.go
```diff
@@ -5,4 +5,5 @@ type SegmentEvent string
 const (
 	NewUser            SegmentEvent = "New User"
 	RedeployViaWebhook SegmentEvent = "Triggered Re-deploy via Webhook"
+	NewClusterEvent			SegmentEvent = "New Cluster Event"
```
abelanger5, line 8: Run `gofmt` on the files, there are a few places where it hasn't been run (should happen automatically on save in VSCode)

internal/analytics/tracks.go
```diff
@@ -66,3 +66,35 @@ func (t *segmentRedeployViaWebhookTrack) getEvent() SegmentEvent {
...
+func (t *segmentNewClusterEventTrack) getUserId() string {
+	return t.userId
+}
+
+func (t *segmentNewClusterEventTrack) getEvent() SegmentEvent {
+	return NewClusterEvent
+}
+
+func (t *segmentNewClusterEventTrack) getProperties() segment.Properties {
+	return segment.NewProperties().Set("Project ID", t.projId).Set("Cluster Name", t.clusterName).Set("Cluster Type", t.clusterType).Set("Event Type", t.eventType)
```
abelanger5, line 99: Split over multiple lines

server/api/cluster_handler.go
```diff
@@ -54,6 +56,7 @@ func (app *App) HandleCreateProjectCluster(w http.ResponseWriter, r *http.Reques
 	}
 
 	app.Logger.Info().Msgf("New cluster created: %d", cluster.ID)
+	app.analyticsClient.Track(analytics.CreateSegmentNewClusterEvent(fmt.Sprintf("%v", userID), fmt.Sprintf("%v", projID), cluster.Name, "", "connected"))
```
abelanger5, line 59: All of the `app.analyticsClient.Track` lines should be split over multiple lines as well

internal/analytics/tracks.go
```diff
@@ -66,3 +66,35 @@ func (t *segmentRedeployViaWebhookTrack) getEvent() SegmentEvent {
...
+	userId    string
+	projId string
+	clusterName string
+	clusterType string // EKS, DOKS, or GKE
+	eventType string // connected, provisioned, or destroyed
+}
+
+// Constructor for track of type "New Cluster Event"
+// tracks whenever a cluster is newly provisioned, connected, or destroyed.
+func CreateSegmentNewClusterEvent(userId string, projId string, clusterName string, clusterType string, eventType string) *segmentNewClusterEventTrack {
```
abelanger5, line 80: This feels pretty redundant, why not just export via `SegmentNewClusterEventTrack` and pass that in directly to the analytics function?

abelanger5 (changes requested):
Looks good! Just a few minor code-level changes.

internal/analytics/tracks.go
jnfrati: From what I remember we talked about when making this package, the idea behind the constructors was to prevent passing directly the struct to the analytics function and instead have a standardized way of using constructors that may or not have additional null safe checks. From that perspective, I think it's okay to have the constructor even if it's a little bit redundant.

abelanger5: Hmm, if that's the case the constructor should really take an `opts` object or something similar which looks like:

```
type NewClusterOpts struct {
        UserId    string
	ProjId string
	ClusterName string
	ClusterType string // EKS, DOKS, or GKE
	EventType string
}
```

Instead of passing all 5 args in directly.

internal/analytics/tracks.go
jnfrati: As I ended up messing this up on the rest of the package, we should also document this on a backend style guide. Idk if it's okay to change only this comment for now and after that, update the rest of the package comments on another PR or do all of them on the new PR and leave this one as it is for now.

internal/analytics/tracks.go
jnfrati: That could be a better option actually, we already do something similar with the User model for tracking newly registered users.

abelanger5 (approved):

PR #955, opened 2021-08-02
[0.7.0] HPA metrics graphic fix

jnfrati:
## Pull request type

- [x] Bugfix
- [ ] Feature

## Pull request checklist

When autoscaling is enabled after the application is created, the chart would have missing data and line points were not rendering properly.
image

## What is the new behavior?

Improve labels thx @sunguroku 

If there's no correspondent data for that date value, it will only show the metric data.
image

### DIFF
--- a/dashboard/src/main/home/cluster-dashboard/expanded-chart/metrics/AreaChart.tsx
+++ b/dashboard/src/main/home/cluster-dashboard/expanded-chart/metrics/AreaChart.tsx
@@ -18,6 +18,15 @@ import { LinearGradient } from "@visx/gradient";
 import { max, extent, bisector } from "d3-array";
 import { timeFormat } from "d3-time-format";
 import { NormalizedMetricsData } from "./types";
+import {
+  AnimatedAreaSeries,
+  AnimatedLineSeries,
+  Axis,
+  Grid,
+  XYChart,
+  Tooltip as XYTooltip,
+  darkTheme,
+} from "@visx/xychart";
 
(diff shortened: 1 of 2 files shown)


dashboard/src/main/home/cluster-dashboard/expanded-chart/metrics/AreaChart.tsx
```diff
@@ -194,7 +207,7 @@ const AreaChart: React.FunctionComponent<AreaProps> = ({
     (hpaEnabled &&
       tooltipData?.tooltipHpaData &&
       valueScale(getValue(tooltipData?.tooltipHpaData))) ||
-    0;
+    undefined;
```
abelanger5, line 210: Would `|| undefined` do nothing, since `undefined` has bool value `false`?

jnfrati: Well yes and no. If the previous value is false, it will default the variable to undefined, so I can perform checks like this `hpaGraphTooltipGlyphPosition !== undefined`

This is a bad practice but the idea is to replace this whole component with an XYChart so this is a temporary solution

abelanger5: Oh whoops, didn't see the extra parentheses and thought the value was getting cast to a bool. Sounds good!

abelanger5 (approved):

PR #963, opened 2021-08-03
[0.7.0] Fix WSL issues on auth login

jnfrati:
## Pull request type

- [x] Bugfix
- [ ] Feature

## Pull request checklist

If the CLI is running under WSL is common to find issues when trying to open the browser or even on the redirection from login if opening the browser is solved.

## What is the new behavior?

Added functions to retrieve if the CLI is running under WSL and if so, use correct commands for opening the browser or to assign correctly the redirection url

For checking if the user is running under WSL I used the uname -a command and check if it has "microsoft" in it. 

And for getting the host url the provided solution by wsl is using `wsl.exe hostname -I` from inside wsl to get the exact IP on where the subsystem is running.

### DIFF
--- a/cli/cmd/login/server.go
+++ b/cli/cmd/login/server.go
@@ -57,7 +57,6 @@ func Login(
 	// open browser for host login
 	var redirectHost string
 	if utils.CheckIfWsl() {
-		fmt.Println("Got till here")
 		redirectHost = fmt.Sprintf("http://%s:%d", utils.GetWslHostName(), port)
 	} else {
 		redirectHost = fmt.Sprintf("http://localhost:%d", port)
--- a/cli/cmd/login/server.go
+++ b/cli/cmd/login/server.go
@@ -55,7 +55,14 @@ func Login(
 	}()
 
 	// open browser for host login
-	redirectHost := fmt.Sprintf("http://localhost:%d", port)
+	var redirectHost string
+	if utils.CheckIfWsl() {
+		fmt.Println("Got till here")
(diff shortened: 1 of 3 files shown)


cli/cmd/login/server.go
```diff
@@ -55,7 +55,14 @@ func Login(
 	}()
 
 	// open browser for host login
-	redirectHost := fmt.Sprintf("http://localhost:%d", port)
+	var redirectHost string
+	if utils.CheckIfWsl() {
+		fmt.Println("Got till here")
```
abelanger5, line 60: Remove debug line

cli/cmd/utils/browser.go
```diff
@@ -17,8 +18,14 @@ func OpenBrowser(url string) error {
...
 	default: // "linux", "freebsd", "openbsd", "netbsd"
-		cmd = "xdg-open"
+		if CheckIfWsl() {
+			cmd = "cmd.exe"
+			args = []string{"/c", "start"}
+		} else {
+			cmd = "xdg-open"
+		}
 	}
+	fmt.Println(url)
```
abelanger5, line 28: It does actually make sense to print out the URL, but perhaps make it a message like:

"Opening your browser at <url>. If the browser does not automatically open, go to this URL to log in."

abelanger5 (changes requested):
Thanks @jnfrati -- also it works on Mac!

abelanger5 (approved):

PR #972, opened 2021-08-04
Fix for issue with icons not showing up

jlorezz:


Xetera:
Is it a good idea to be pulling things out of the `/wp-content/` of other sites? Seems like this site doesn't support serving images directly and prompt you to download them which might mean these endpoints can change at any time.

jlorezz:
> Is it a good idea to be pulling things out of the `/wp-content/` of other sites? Seems like this site doesn't support serving images directly and prompts you to download them which might mean these endpoints can change at any time.

you are right. we could just host them our own with like imgur or something.

jusrhee:
You can just put the files under `dashboard/src/assets` along with the other images

jlorezz:
> You can just put the files under `/dashboard/src/assets` along with the other images

sounds good! will do that.

jusrhee:
@jlorezz trying to review but could you pull from master - looks like your branch is nearly 1000 commits behind

jlorezz:
> @jlorezz trying to review but could you pull from master - looks like your branch is nearly 1000 commits behind

will do! one second.

jusrhee:
Great, thanks! Also, still not seeing an icon after linking up a DigitalOcean registry for some reason:
image

PR #979, opened 2021-08-05
[0.7.0] Upgrade notes and ability to upgrade add-ons

abelanger5:
## Pull request type

- [ ] Bugfix
- [X] Feature

## Pull request checklist

Users cannot update add-ons, and there are no upgrade notes rendered when wishing to upgrade. 

## What is the new behavior?

Render a modal that displays upgrade notes. If no upgrade notes are found, tell the user that the upgrade should be backwards compatible. 

- UpgradeChartModal is left as a class component as all other modals are class-based components, and they should probably be refactored together (lots of repeated styling + functionality)
- The **target chart** requires an `upgrade.yaml` file to exist with the following structure:

```yaml
upgrade_notes:
- previous: <previous-version>
  target: <target-version>
  note: |
    These are custom notes!
```

The note gets rendered as markdown (code blocks should be supported). The modal should only render the upgrade notes which are relevant to that upgrade (i.e. if upgrading `0.4.0` -> `0.5.0`, the upgrade notes for `0.3.0` -> `0.4.0` will **not** be rendered).


sunguroku:
not sure if this is expected behavior, but "addons" that do not exist in our chart repository such as cert-manager also renders the template upgrade option.
image

abelanger5:
> Also, not sure if this is expected behavior, but "addons" that do not exist in our chart repository such as cert-manager also renders the template upgrade option. Not sure what this would do, or which chart registry it's looking at.

Actually the issue here is that the cert-manager chart does exist in our addons, and so it had a difficult time comparing. As a quick fix, I added a comparison check for the versions so that it'll use the existing chart version if it's greater than the most recent version in our repo. This isn't a long-term solution, but i couldn't find a quick way to distinguish charts from our repos with other repos. 

> Another minor thing that I think could be a problem is that there is no loading indicator when the upgrade notes are being rendered, so it looks like the note is empty. The rendering time is noticeable even when the upgrade note just has one item, which I can see being an actual problem when there are multiple upgrade notes present.

I'm a little confused by this, are you not seeing the loading indicator?

sunguroku:
hm seems like loading indicator is showing now.. not sure why it didn't show before

checked that cert-manager and other charts (like ingress controller) isn't showing the upgrade template button

sunguroku (approved):

PR #996, opened 2021-08-05
[0.7.0] `porter run` should spawn ephemeral pods

abelanger5:
## Pull request type

- [ ] Bugfix
- [X] Feature

## Pull request checklist

`porter run...` connects to existing pods, which is not desired in many cases. 
 
## What is the new behavior?

`porter run...` copies the configuration for a different pod, and uses that pod to run commands in. Using the flag `--existing_pod` will provide the old functionality.


sunguroku:
Works well but a few issues I've found:
- Opening an interactive shell works, but for some reason commands that exit such as `printenv` or `echo 'hello world'` fail with the following error. I've checked that the same commands execute properly when run inside the interactive shell.
Screen Shot 2021-08-05 at 8 35 22 PM
- Before the `Running [CMD] for release...` prompt is shown, there is no visual indication that the command is in loading state, so it wasn't clear that the command is working. 
- Assuming this is nontrivial for it to be worth, but I missed the `If you don't see a command prompt, try pressing enter` prompt and thought it was stuck at a reconnect attempt.

abelanger5:
Ok, fixed a few of these issues. 

> Opening an interactive shell works, but for some reason commands that exit such as `printenv` or `echo 'hello world'` fail with the following error. I've checked that the same commands execute properly when run inside the interactive shell.

Just made the default behavior to output the logs from this pod when this happens. I also added an option `porter logs ...` while I was at it. 

> Before the `Running [CMD] for release...` prompt is shown, there is no visual indication that the command is in loading state, so it wasn't clear that the command is working.

This is because you're running with `go run main.go` first, so it needs to compile -- when the binary is already built, this is logged immediately. 

> Assuming this is nontrivial for it to be worth, but I missed the `If you don't see a command prompt, try pressing enter` prompt and thought it was stuck at a reconnect attempt.

`kubectl` also has this behavior, and it's not very clear how to get around it. I don't think it's worth tackling this, since the issue is with an [underlying package](https://pkg.go.dev/k8s.io/kubectl@v0.21.0/pkg/util/term#TTY.Safe).

sunguroku (approved):

PR #1008, opened 2021-08-09
[0.8.0] Pod events managment

jnfrati:
## Pull request type

- [ ] Bugfix
- [x] Feature

## Pull request checklist

Users don't have any historical view to see what's happening with the pods of one application.

Issue Number: N/A

## What is the new behavior?

The idea is to give users two screens to see pod events. 

The main one would be located in the cluster dashboard view as a new tab where the user can see all the logs that are currently happening all over that cluster, this includes events that may involve non-helm charts deployments (such as k8s API server) and events that correspond to their deployed apps through helm charts.

The second one would be a specific application-level tab, where you will be able to see all the pod events for that specific application.

To be able to retrieve specific application pods, the main change that we need to make would be to add an annotation for the chart name inside the application objects, the most important one would be pods, but we were talking about adding it to the other controllers as well e.g: deployments.

For the Porter API, we need an endpoint and a webhook. 
- The webhook would be used for the Porter Agent to give porter the ability to save in it's own database all the events that are happening on different clusters
- And for the endpoint, the idea would be to have something like `GET /projects/{project_id}/k8s/{namespace}/pods/events`, this endpoint will be able to filter by the next query params: 
  - `chart_name="<chartName>"` This will retrieve all the events for pods that have the given chart name into their annotations.
  - `pod_name="<podName>"` Return only given pod events, if possible, it should return also if the pod exists but not for that specific chart in case this query param is used with the previous one.


jnfrati:
The idea for the events tab, on the application level or the cluster level, would be to look like this. The only thing that changes is the list of pods that you can select for showing the events.
The little icon at the right will be used to show the last 100 logs of the event and the icon on the left will be a dynamic icon that will represent what type of event it was (warning or natural) according to the [Event type](https://pkg.go.dev/k8s.io/api/events/v1#Event) on the k8s specification.
image

jusrhee:
Looks good! Just to clarify, would showing the previous 100 logs just jump to the relevant part of the logs tab or show the logs inline?

jnfrati:
> Looks good! Just to clarify, would showing the previous 100 logs just jump to the relevant part of the logs tab or show the logs inline?

Oh, I forgot to clarify this. As it's kinda difficult to scroll to a specific part of the logs tab, it would be easier to just open a modal with the logs that we're trying to show for that specific event. What do you think @jusrhee ?

jusrhee:
Yup, makes sense - was just curious.

sunguroku:
Looks good to me, assuming this entails getting rid of the system tab in logs view? It was a bit problematic that the pod events were not conspicuous enough anyway so I think a separate events tab will work well.

In terms of future considerations, I think piping in events from HPA will be essential even for v1 - this might make a case for having an api call for chart wide events where a single endpoint returns events across all objects in the chart, instead of subscribing to the events endpoint of each object from the frontend. (e.g. 'events/OBJECT_NAME'). 

I think it really depends on the number of objects we will typically subscribe to for events - what are some controllers you were mentioning that would be included in the global cluster events tab?

If the scope of this PR is just the application level events, looks good to me - or does it include the cluster wide events view? Then it'd be nice to get a similar visualization + details on what objects we'll subscribe to cluster-wide.

abelanger5:
> this includes events that may involve non-helm charts deployments (such as k8s API server)

Just to clarify, for this PR, we are **not** implementing a deployment controller, correct? The cluster-level events tab will just be an aggregation of the pod-level events (for now)? 

> To be able to retrieve specific application pods, the main change that we need to make would be to add an annotation for the chart name inside the application objects

I believe the controller objects already have a set of annotations attached that provide the helm chart/name. The pod objects do not, but it should be possible to find by looking up the owner references. I'd like to do this, to support this being backwards-compatible. 

> And for the endpoint, the idea would be to have something like `GET /projects/{project_id}/k8s/{namespace}/pods/events`, this endpoint will be able to filter by the next query params:
> - `chart_name="<chartName>"` This will retrieve all the events for pods that have the given chart name into their annotations.
> - `pod_name="<podName>"` Return only given pod events, if possible, it should return also if the pod exists but not for that specific chart in case this query param is used with the previous one.

I believe this endpoint also needs the query params `limit` and `skip` for pagination purposes. Also, getting the pod logs will be a separate API endpoint, correct? I think the `/events` query will be too large if we attach the logs for every single pod event.

ishankhare07:
> Just to clarify, for this PR, we are not implementing a deployment controller
Since the controller is listening to all kinds of pod events, there will be pods not deployed through a helm chart.

ishankhare07:
>  webhook would be used for the Porter Agent to give porter the ability to save in it's own database

Not sure what do you mean by a webhook here, I was assuming we are just going ahead with a simple `POST` endpoint that the agent can post on

jnfrati:
@sunguroku 
> I think it really depends on the number of objects we will typically subscribe to for events - what are some controllers you were mentioning that would be included in the global cluster events tab?
I'm sorry this probably was misunderstood, the events that will be shown on the cluster events tab on the final version would be the following: 
- All HPA events that happen across the cluster
- All Pod events that happen across the cluster, including pods that may not belong to a helm chart (so they wouldn't be accessible in any other way e.g: kube api pods)
- All node events that happen across that cluster

For this version (unless the time is enough to add something else) we will include only POD events, HPA and Node events will be included in future versions of the event tabs

@abelanger5 
> I believe this endpoint also needs the query params limit and skip for pagination purposes. Also, getting the pod logs will be a separate API endpoint, correct? I think the /events query will be too large if we attach the logs for every single pod event.
I agree with you, those query params would be really useful, the only thing that worries me a little bit is how pagination will work when the server is getting new events frequently (skip 10, would skip different objects each time a new event is posted from the porter agent to the porter API) wdyt? 

@ishankhare07 
> Not sure what do you mean by a webhook here, I was assuming we are just going ahead with a simple `POST` endpoint that the agent can post on

Yep, that's exactly what I was saying, just made the difference between endpoint and webhook to clarify the one that it's going to be called from the frontend and the one that's going to be called from the porter agent.

jnfrati:
Also, for future reference, the final object that all events (Pod, HPA or Node events) will respect looks like this: 
```
{
  "resource_type": "Pod/HPA/Node",
  "name": "resource_name",
  "owner_type": "deploy/StatefulSet/DaemonSet/...",
  "owner_name": "<name of the owning object>",
  "event_type": "critical / normal",
  "namespace": "namespace_name",
  "cluster": "cluster_name",
  "message": "<message>",
  "reason": "<reason>",
  "timestamp": "2021-06-30T21:48:23Z",
  "data": [
    "in case of normal events this will be empty",
    "in case of critical events, this will have last 100 log lines"
  ]
}
```

abelanger5:
For documentation purposes, the backend implements the following endpoints:

1. `POST /api/projects/{project_id}/clusters/{cluster_id}/events`, expects the following example request body:

```
{
  "resource_type": "pod",
  "name": "pod-test-2",
  "owner_type": "deployment",
  "owner_name": "pod-test",
  "event_type": "critical",
  "namespace": "default",
  "message": "Pod crashed",
  "reason": "OOM killed",
  "timestamp": "2021-06-30T21:48:23Z",
  "data": [
    "log data 1",
    "log data 2"
  ]
}
```

2. `GET /api/projects/{project_id}/clusters/{cluster_id}/events`, expects the following potential query parameters:

```
"limit": int 
"skip": int
"type": string (i.e. "pod" or "node")
"owner_type": string
"owner_name": string
"sort_by": string (only supported is "timestamp")
```

And returns the following sample response:

```
[
  {
    "id": 1,
    "project_id": 1,
    "cluster_id": 6,
    "owner_name": "pod-test",
    "owner_type": "deployment",
    "event_type": "critical",
    "resource_type": "pod",
    "name": "pod-test-1",
    "namespace": "default",
    "message": "",
    "reason": "OOM killed",
    "timestamp": "2021-06-30T21:48:23Z"
  },
  {
    "id": 2,
    "project_id": 1,
    "cluster_id": 6,
    "owner_name": "pod-test",
    "owner_type": "deployment",
    "event_type": "critical",
    "resource_type": "pod",
    "name": "pod-test-2",
    "namespace": "default",
    "message": "",
    "reason": "OOM killed",
    "timestamp": "2021-06-30T21:48:23Z"
  }
]
```

3. `GET /api/projects/{project_id}/clusters/{cluster_id}/events/{event_id}`, returns the following sample response:

```
{
  "id": 1,
  "project_id": 1,
  "cluster_id": 6,
  "owner_name": "pod-test",
  "owner_type": "deployment",
  "event_type": "critical",
  "resource_type": "pod",
  "name": "pod-test-1",
  "namespace": "default",
  "message": "",
  "reason": "OOM killed",
  "data": "bG9nIGRhdGEgMQpsb2cgZGF0YSAy",
  "timestamp": "2021-06-30T21:48:23Z"
}
```

Note that data is **base64 encoded logs**, in this case they decode to:

```
log data 1
log data 2
```

abelanger5:
Endpoints for detecting/creating agent:

1. `POST /api/projects/{project_id}/agent/deploy`, expects `cluster_id` as a query parameter and the same parameters as the standard `/deploy` endpoint.

2. `GET /api/projects/{project_id}/agent/detect`, expects `cluster_id` as a query parameter and the same parameters as the `.../k8s/prometheus/detect` endpoint.

PR #1012, opened 2021-08-10
github app client: don't hardcode secret path

anukul:
## Pull request type

- [x] Bugfix
- [ ] Feature

## Pull request checklist

The `GITHUB_APP_SECRET_PATH` environment variable is not used in the following scenarios:
- In the Launch flow, selecting Deployment Method = Git Repository, and trying to fetch the list of repos.
- Creating a GitHub Actions workflow after deployment.

This causes both of these requests to fail if the secret is not present at `/porter/docker/github_app_private_key.pem`.

## What is the new behavior?

The aforementioned requests now use `GITHUB_APP_SECRET_PATH` and pass, as a result.


server/api/deploy_handler.go
```diff
@@ -375,6 +375,7 @@ func (app *App) HandleUninstallTemplate(w http.ResponseWriter, r *http.Request)
 					ServerURL:              app.ServerConf.ServerURL,
 					GithubOAuthIntegration: gr,
 					GithubAppID:            app.GithubAppConf.AppID,
+					GithubAppSecretPath:    app.GithubAppConf.SecretPath,
```
anukul, line 378: Not sure if `GithubAppSecretPath` should be a new struct field, or we should store a `app.GithubAppConf` to access its two fields used here.

abelanger5: I think this is fine, if we end up pulling the `GithubAppConf` out of the `api` package (which we likely will eventually) it would make sense for them to share a model.

abelanger5 (approved):
Good catch @anukul 👍

PR #1013, opened 2021-08-10
[0.8.0] GitHub Actions updates

anukul:
## Pull request type

- [ ] Bugfix
- [x] Feature

## Pull request checklist

After deploying from a launch flow with the Git Repository source, the Porter server adds a GitHub Actions [workflow](https://github.com/anukul/gin-example/commit/3f15765e8c2b7f223338862d89403dac87dd0bd2) and the associated [secrets](https://sourcegraph.com/search?q=context:global+repo:%5Egithub%5C.com/porter-dev/porter%24%40v0.7.0+file:%5Einternal/integrations/ci/actions/actions%5C.go+create+.*+new+secret&patternType=regexp&case=yes) to the repository. 

This workflow runs `porter update` on every push to the selected branch.

## What is the new behavior?

After deployment, a modal is shown with the contents of the workflow and the secrets.
It also contains a toggle "Push workflow to repository" (default: enabled), in case users want us to skip this step and manage it themselves (e.g. adding to an existing workflow, not using GitHub Actions etc.).

- Use [YamlEditor](https://sourcegraph.com/github.com/porter-dev/porter@v0.7.0/-/blob/dashboard/src/components/YamlEditor.tsx) component to show the workflow
- Add a new route [`GET /projects/{project_id}/ci/actions`](https://sourcegraph.com/github.com/porter-dev/porter@v0.7.0/-/blob/server/router/router.go?L394), call it to fetch the YAML and the secrets _after_ deployment (since this step needs [release](https://sourcegraph.com/github.com/porter-dev/porter/-/blob/server/api/git_action_handler.go?L73) info)
- Create (and publish) a new GitHub Action - "Update Porter App" that performs the update in a single step, and use it in the workflow.

### DIFF
--- a/cli/cmd/api/domain.go
+++ b/cli/cmd/api/domain.go
@@ -17,7 +17,7 @@ type CreateDNSRecordRequest struct {
 // CreateDNSRecordResponse is the DNS record that was created
 type CreateDNSRecordResponse models.DNSRecordExternal
 
-// CreateGithubAction creates a Github action with basic authentication
+// CreateDNSRecord creates a Github action with basic authentication
 func (c *Client) CreateDNSRecord(
 	ctx context.Context,
 	projectID, clusterID uint,
--- a/dashboard/src/main/home/launch/launch-flow/SourcePage.tsx
+++ b/dashboard/src/main/home/launch/launch-flow/SourcePage.tsx
@@ -28,7 +28,7 @@ type PropsType = RouteComponentProps & {
   setImageTag: (x: string) => void;
 
   actionConfig: ActionConfigType;
-  setActionConfig: (x: ActionConfigType) => void;
+  setActionConfig: React.Dispatch<React.SetStateAction<ActionConfigType>>;
   procfileProcess: string;
   setProcfileProcess: (x: string) => void;
   branch: string;
@@ -162,7 +162,10 @@ class SourcePage extends Component<PropsType, StateType> {
           actionConfig={actionConfig}
           branch={branch}
           setActionConfig={(actionConfig: ActionConfigType) => {
-            setActionConfig(actionConfig);
+            setActionConfig((currentActionConfig: ActionConfigType) => ({
+              ...currentActionConfig,
+              ...actionConfig,
+            }));
             setImageUrl(actionConfig.image_repo_uri);
             /*
             setParentState({ actionConfig }, () =>
@@ -173,10 +176,11 @@ class SourcePage extends Component<PropsType, StateType> {
           procfileProcess={procfileProcess}
           setProcfileProcess={(procfileProcess: string) => {
             setProcfileProcess(procfileProcess);
-            setValuesToOverride({
+            setValuesToOverride((v: any) => ({
+              ...v,
               "container.command": procfileProcess || "",
               showStartCommand: !procfileProcess,
-            });
+            }));
           }}
           setBranch={setBranch}
           setDockerfilePath={setDockerfilePath}
--- a/dashboard/src/main/home/launch/launch-flow/SourcePage.tsx
+++ b/dashboard/src/main/home/launch/launch-flow/SourcePage.tsx
@@ -28,7 +28,9 @@ type PropsType = RouteComponentProps & {
   setImageTag: (x: string) => void;
 
   actionConfig: ActionConfigType;
-  setActionConfig: React.Dispatch<React.SetStateAction<ActionConfigType>>;
+  setActionConfig: (
+    x: ActionConfigType | ((prevState: ActionConfigType) => ActionConfigType)
+  ) => void;
   procfileProcess: string;
   setProcfileProcess: (x: string) => void;
   branch: string;
--- a/dashboard/src/main/home/launch/launch-flow/SourcePage.tsx
+++ b/dashboard/src/main/home/launch/launch-flow/SourcePage.tsx
@@ -227,6 +227,16 @@ class SourcePage extends Component<PropsType, StateType> {
     }
   };
 
+  handleContinue = () => {
+    const { sourceType, setPage } = this.props;
+
+    if (sourceType == "repo") {
+      setPage("workflow");
+    } else if (sourceType == "registry") {
+      setPage("settings");
+    }
+  };
+
   render() {
     let { templateName, setTemplateName, setPage } = this.props;
 
@@ -270,7 +280,7 @@ class SourcePage extends Component<PropsType, StateType> {
         <SaveButton
           text="Continue"
           disabled={!this.checkSourceSelected()}
-          onClick={() => setPage("workflow")}
+          onClick={this.handleContinue}
           status={this.getButtonStatus()}
           makeFlush={true}
           helper={this.getButtonHelper()}
--- a/dashboard/src/main/home/launch/launch-flow/SourcePage.tsx
+++ b/dashboard/src/main/home/launch/launch-flow/SourcePage.tsx
@@ -230,9 +230,9 @@ class SourcePage extends Component<PropsType, StateType> {
   handleContinue = () => {
     const { sourceType, setPage } = this.props;
 
-    if (sourceType == "repo") {
+    if (sourceType === "repo") {
       setPage("workflow");
-    } else if (sourceType == "registry") {
+    } else {
       setPage("settings");
     }
   };
--- a/dashboard/src/main/home/launch/launch-flow/SourcePage.tsx
+++ b/dashboard/src/main/home/launch/launch-flow/SourcePage.tsx
@@ -270,7 +270,7 @@ class SourcePage extends Component<PropsType, StateType> {
         <SaveButton
           text="Continue"
           disabled={!this.checkSourceSelected()}
-          onClick={() => setPage("settings")}
+          onClick={() => setPage("workflow")}
           status={this.getButtonStatus()}
           makeFlush={true}
           helper={this.getButtonHelper()}
--- a/dashboard/src/main/home/launch/launch-flow/WorkflowPage.tsx
+++ b/dashboard/src/main/home/launch/launch-flow/WorkflowPage.tsx
@@ -1,0 +1,170 @@
+import React, { useContext, useEffect, useState } from "react";
+import { RouteComponentProps, withRouter } from "react-router";
+import { FullActionConfigType } from "../../../../shared/types";
+import api from "../../../../shared/api";
+import { Context } from "../../../../shared/Context";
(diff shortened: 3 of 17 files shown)


abelanger5:
Looks great! Just one minor thing:

> * Add a new route [`GET /projects/{project_id}/ci/actions`](https://sourcegraph.com/github.com/porter-dev/porter@v0.7.0/-/blob/server/router/router.go?L394), call it to fetch the YAML and the secrets _after_ deployment (since this step needs [release](https://sourcegraph.com/github.com/porter-dev/porter/-/blob/server/api/git_action_handler.go?L73) info)

It should be possible to generate this YAML without requiring release information beyond the release name. In the current user flow, the release name is changeable until the last step. However, I think it's acceptable to modify the user flow so that the release name can only be set on the first step of deployment.

I think that if we show the Github action YAML after the user has configured the rest of the template settings, there is still some initial confusion about what happens when you move beyond the "Link from Github repository" step.

anukul:
@abelanger5 wrt this
> It should be possible to generate this YAML without requiring release information beyond the release name. 

Can I create the `release.WebhookToken` on the client, or should I instead add a new endpoint for this on the server (to use the existing method [[1]](https://sourcegraph.com/github.com/porter-dev/porter@5646693316f6044ae5b2632c8b7ca5ba78891932/-/blob/server/api/deploy_handler.go?L121:27) [[2]](https://sourcegraph.com/github.com/porter-dev/porter@5646693316f6044ae5b2632c8b7ca5ba78891932/-/blob/server/api/release_handler.go?L843:27) of calling [GenerateRandomBytes](https://sourcegraph.com/github.com/porter-dev/porter@5646693316f6044ae5b2632c8b7ca5ba78891932/-/blob/internal/repository/encrypt.go?L27:6))?

abelanger5:
> Can I create the `release.WebhookToken` on the client, or should I instead add a new endpoint for this on the server (to use the existing method [[1]](https://sourcegraph.com/github.com/porter-dev/porter@5646693316f6044ae5b2632c8b7ca5ba78891932/-/blob/server/api/deploy_handler.go?L121:27) [[2]](https://sourcegraph.com/github.com/porter-dev/porter@5646693316f6044ae5b2632c8b7ca5ba78891932/-/blob/server/api/release_handler.go?L843:27) of calling [GenerateRandomBytes](https://sourcegraph.com/github.com/porter-dev/porter@5646693316f6044ae5b2632c8b7ca5ba78891932/-/blob/internal/repository/encrypt.go?L27:6))?

I'm a little confused -- the webhook token being generated isn't a requirement to show the YAML, right? You just need to generate a display-only YAML file (without running all of the steps that the `actions` package runs in the `Setup()` function.

anukul:
I was thinking to also show the secrets. So in case they choose to copy & paste the workflow by themselves, they can also add the secrets that it needs to run.

> ## What is the new behavior?
> 
> After deployment, a modal is shown with the contents of the workflow and the secrets.
>

abelanger5:
> I was thinking to also show the secrets. So in case they choose to copy & paste the workflow by themselves, they can also add the secrets that it needs to run.

Ah, I see. In that case, I think there should be separate toggles for writing the secrets and the workflow file (it may be desirable to write the secrets, but not the workflow). In this case, however, there should only be a single secret, the API token that the Porter server will use. Where are you seeing that the webhook token is a secret? (that was the case on previous versions of the server, but should no longer be a requirement) 

But you can create an endpoint for generating an API token to be used. In this case, however, we should add API tokens to the Project Settings screen, where the tokens can be revoked from.

anukul:
@abelanger5 Noting the conclusion of our discussion here, please confirm if this looks good:

Part 1
---

For the scope of this PR, the new modal will only show the workflow YAML, not the secrets.

If a user disables the "Push workflow to repository" toggle, the workflow will not be pushed, but the secrets will still be created either way.
This allows users to copy & paste the shown YAML into a workflow of their choice, while still having access to the secrets it references.

Part 2
---

Add a new `Version uint` field (default: `1`) to the [`GitActionConfig`](https://sourcegraph.com/-/editor?remote_url=git%40github.com%3Aporter-dev%2Fporter.git&branch=0.8.0-github-actions-updates&file=internal%2Fmodels%2Fgitrepo.go&editor=JetBrains&version=v1.2.0&utm_product_name=IntelliJ+IDEA&utm_product_version=2021.2&start_row=48&start_col=29&end_row=48&end_col=29) model.

This field will indicate which GitHub Actions workflow is used by a deployment (before [`1`] / after [`2`] v0.7.0 - https://github.com/porter-dev/porter/pull/935) and will be used to decide accordingly whether to create the [env secret](https://sourcegraph.com/-/editor?remote_url=git%40github.com%3Aporter-dev%2Fporter.git&branch=0.8.0-github-actions-updates&file=internal%2Fintegrations%2Fci%2Factions%2Factions.go&editor=JetBrains&version=v1.2.0&utm_product_name=IntelliJ+IDEA&utm_product_version=2021.2&start_row=309&start_col=24&end_row=309&end_col=39) during release upgrades and rollbacks.

New deployments use the updated workflow, but still create 3 [secrets](https://sourcegraph.com/search?q=context:global+repo:%5Egithub%5C.com/porter-dev/porter%24%40v0.7.0+file:%5Einternal/integrations/ci/actions/actions%5C.go+create+.*+new+secret&patternType=regexp&case=yes) as per the old workflow. This will be changed so only the API token secret is created now.

anukul:
Screenshot of GitHub Actions page (after Source Settings):

image


Walkthrough:

https://user-images.githubusercontent.com/44864521/129637861-808ace72-cd8a-4df1-a9a4-11cff036ee77.mov

cli/cmd/api/domain.go
```diff
@@ -17,7 +17,7 @@ type CreateDNSRecordRequest struct {
 // CreateDNSRecordResponse is the DNS record that was created
 type CreateDNSRecordResponse models.DNSRecordExternal
 
-// CreateGithubAction creates a Github action with basic authentication
+// CreateDNSRecord creates a Github action with basic authentication
```
abelanger5, line 20: Nice catch 🥇

server/api/git_action_handler.go
```diff
@@ -20,6 +20,48 @@ const (
...
+	name := vals["name"][0]
+
+	clusterID, err := strconv.ParseUint(vals["cluster_id"][0], 10, 64)
+
+	if err != nil {
+		app.sendExternalError(err, http.StatusInternalServerError, HTTPError{
+			Code:   ErrReleaseReadData,
+			Errors: []string{"release not found"},
+		}, w)
+	}
```
abelanger5, line 43: Add return statement here after sending error

abelanger5 (changes requested):
Working great @anukul! Some minor fixes.

abelanger5 (changes requested):
Some documentation considerations:
- Can you add a link to the update actions repo, if the user chooses **not** to write the workflow? 
- Can you add a section to the docs with some details on the workflow file, and how to add the workflow file to the repo if the user doesn't want to auto-write the workflow? i'm talking about [this doc](https://github.com/porter-dev/porter/blob/master/docs/deploy/applications/deploying-from-git-repo.md).

dashboard/src/main/home/launch/launch-flow/SourcePage.tsx
```diff
@@ -28,7 +28,7 @@ type PropsType = RouteComponentProps & {
   setImageTag: (x: string) => void;
 
   actionConfig: ActionConfigType;
-  setActionConfig: (x: ActionConfigType) => void;
+  setActionConfig: React.Dispatch<React.SetStateAction<ActionConfigType>>;
```
jnfrati, line 31: Not entirely sure about this one, why tightly couple the prop to something internal to React instead of receiving a function?

dashboard/src/main/home/launch/launch-flow/WorkflowPage.tsx
```diff
@@ -0,0 +1,170 @@
...
+        makeFlush={true}
+        disabled={hasError}
+        onClick={() => props.setPage("settings")}
+        helper={getButtonHelper()}
+      />
+    </StyledWorkflowPage>
+  );
+};
+
+export default withRouter(WorkflowPage);
```
jnfrati, line 103: Here is better to use the react router hooks, such as useHistory or useLocation

jnfrati (changes requested):
Nice work the one that you have here! Left some comments that I would like to discuss before merging 🙌

anukul:
@abelanger5 added link to porter-update-action:


https://user-images.githubusercontent.com/44864521/129776851-7cc238ab-d70f-473a-8c4a-41ba277068e8.mov

dashboard/src/main/home/launch/launch-flow/SourcePage.tsx
anukul: Since we use both forms of the `setState` function here, the alternative is:

```js
  setActionConfig: (
    x: ActionConfigType | ((prevState: ActionConfigType) => ActionConfigType)
  ) => void;
```

Should I change it to this?

jnfrati: Probably it would be better to not use the set function directly, but instead pass a handler function to the SourcePage. If this is not possible, then the typing solution that you described is better as it doesn't depend on some internal React typings 🙌

anukul: I agree with the refactoring suggestion. I haven't also attended to some other changes, so maybe we can batch them up for later.

I don't think this is an internal type, since it's part of React's ["Top-Level API"](https://reactjs.org/docs/react-api.html).

anukul: Got it!

abelanger5 (dismissed):

abelanger5 (changes requested):
Hey @anukul this should be ready for merging, but a compilation issue needs to be fixed (believe it came from the most recent commit:

```
/webpack/src/main/home/launch/launch-flow/LaunchFlow.tsx(330,10)
webpack_1      |       TS2739: Type '{ name: string; fullActionConfig: FullActionConfigType; shouldCreateWorkflow: boolean; setShouldCreateWorkflow: Dispatch<SetStateAction<boolean>>; setPage: Dispatch<...>; }' is missing the following properties from type 'RouteComponentProps<{}, StaticContext, unknown>': history, location, match
```

abelanger5 (approved):

abelanger5 (approved):

PR #1023, opened 2021-08-11
[0.8.0] - Error boundary implementation

jnfrati:
## Pull request type

- [ ] Bugfix
- [x] Feature

## Pull request checklist

## Motivation

Avoid blank screens when an un-catched error happens, usually if a react component cannot render for whatever is the reason, the application goes blank. 
If this happens, instead of a blank screen, the user will see something like this: 
image


abelanger5 (approved):
Works great @jnfrati 👍

PR #1027, opened 2021-08-11
[0.8.0] More Slack Notification Options

igalakhov:
## Pull request type

- [ ] Bugfix
- [x] Feature

## Pull request checklist

When a slack notifier is set up for a project, notifications for all deployments are sent out all the time.

## What is the new behavior?

Users will be able to select which specific notifications (if any) they want to be sent through. On the frontend, this will be a set of checkboxes under "settings" in expanded chart like:

[ ] Enable notifications for project

and then another section that appears with specific events
[ ] deploy [ ] success [ ] fail

Likely going to create a model linked to each deployment (that is, each deployment stores an id to a model since there's no need to copy things) that stores a few fields related to the notifier:
- whether or not slack notifications are on for this deployment
- boolean fields for each type of event (this is the part that needs to be done carefully since we will have events in the future)

And then the part of the code that actually sends the notification needs a function that determines if it should be sent for a specific event.


abelanger5:
Makes sense to me 👍 . A few things:

1. Just make sure to decouple the notification settings for the deployment from Slack, since we'll add more notification options (like email) pretty soon. (It seems like you're already doing this, but just wanted to state explicitly).

2. What are the default notification settings that will be written for each deployment? Will the default be **not** to notify at all?

3. Regarding this checkbox:

> [ ] Enable notifications for project

Just to clarify, is this "Enable notifications for deployment"? Or do you mean the whole project?

4. If there are no Slack channels set up (in general, no notification integrations at all), we should inform the user in the "Settings" tab as well.

server/api/notifications_handler.go
```diff
@@ -0,0 +1,137 @@
...
+	if err != nil {
+		if err == gorm.ErrRecordNotFound {
+			w.WriteHeader(http.StatusNotFound)
+			return
+		}
+
+		app.sendExternalError(err, http.StatusInternalServerError, HTTPError{
+			Code:   ErrReleaseReadData,
+			Errors: []string{"release not found"},
+		}, w)
```
abelanger5, line 45: Add return statement when sending error

server/api/notifications_handler.go
```diff
@@ -0,0 +1,137 @@
...
+	name := chi.URLParam(r, "name")
+	namespace := vals["namespace"][0]
+
+	clusterID, err := strconv.ParseUint(vals["cluster_id"][0], 10, 64)
+
+	if err != nil {
+		app.sendExternalError(err, http.StatusInternalServerError, HTTPError{
+			Code:   ErrReleaseReadData,
+			Errors: []string{"release not found"},
+		}, w)
```
abelanger5, line 99: Add return statement when sending error

abelanger5 (approved):

PR #1030, opened 2021-08-12
add porter-update-action_container

anukul:
This image will be used to run [`porter-dev/porter-update-action`](https://github.com/porter-dev/porter-update-action)

Related - #1013


anukul:
Should we tag this image with the action version, or a combination of the action version and the bundled CLI version?

anukul:
Should we tag this image with the action version, a combination of the action version and the bundled CLI version, or something else?

abelanger5 (approved):

PR #1038, opened 2021-08-13
[0.8.0] Use porter-update-action

anukul:
## Pull request type

Related: #1013

- [ ] Bugfix
- [ ] Feature
- [x] Other (please describe): Refactoring

## Pull request checklist

- Update workflow YAML is generated by server
- Webhook token and build env secrets are created in new deployments, which is unneccessary with the new workflow

## What is the new behavior?

- Server now uses [porter-update-action](https://github.com/porter-dev/porter-update-action)
- Webhook token and build env secrets are no longer created for new deployments.  
Note: Logic for the webhook token secret is completely removed, so it's also not deleted for existing deployments anymore.

---


abelanger5 (changes requested):
Code-wise this looks good @anukul. A few clarifications:
1. The webhook token secret and env variable token secret are still being called by the rollback/upgrade endpoints, correct? If that's the case, they'll still be written when applications using the Github marketplace action are updated. We still need to introduce casing to differentiate the Github actions version. 
2. Regarding this:
> Note: Logic for the webhook token secret is completely removed, so it's also not deleted for existing deployments anymore.

Why did we remove the webhook token logic completely? It still seems desirable to delete this secret, if it exists.

internal/config/config.go
```diff
@@ -71,6 +71,8 @@ type ServerConf struct {
 	ProvisionerCluster string `env:"PROVISIONER_CLUSTER"`
 	IngressCluster     string `env:"INGRESS_CLUSTER"`
 	SelfKubeconfig     string `env:"SELF_KUBECONFIG"`
+
+	UpdateAppGHAVersion string `env:"UPDATE_APP_GHA_VERSION,default=v0.1.0"`
```
anukul, line 75: Not sure if it's is the best place to put this.

internal/models/gitrepo.go
```diff
@@ -75,6 +75,8 @@ type GitActionConfig struct {
 
 	// Determines on how authentication is performed on this action
 	IsInstallation bool `json:"is_installation"`
+
+	Version string `json:"version" gorm:"default:v0.0.1"`
```
abelanger5, line 79: The default is set here to `v0.0.1`, but the constraint is set as `< 0.1.0`. Shouldn't the default be `v0.1.0` as well?

abelanger5 (changes requested):

internal/models/gitrepo.go
anukul: I set this for existing records so the constraint only passes for them.

abelanger5 (approved):

Issue #857, opened 2021-07-06
Initial push to GCR from Porter GHA appears to fail


jusrhee:
# Description

If a user provisions GCR through Porter (or has a registry with no images) the initial container image push from our GHA appears to fail.

# Location

- [ ] Browser
- [X] CLI
- [ ] API

# Steps to reproduce

1. Provision to GCP through Porter (GKE + GCR) - ensure that no images exist in the GCR instance
2. Deploy an application from a connected repo
3. GHA fails on image push

# Additional Details

abelanger5:
Closed by #1055

PR #1055, opened 2021-08-14
[0.8.0] Retry GCR pushes when first one fails

abelanger5:
## Pull request type

- [X] Bugfix
- [ ] Feature

## Pull request checklist

Issue Number: #857 

## What is the new behavior?

Retry pushes to GCR in the case that the service account temporarily does not have push access to the storage bucket.


abelanger5:
Turns out the issue here was actually with the caching: specifically, when a new token had to be generated (typically on the first push to the GCR repository), the token was retrieved properly, and was written to the cache, but the token was not retrieved properly after being written to the cache in the `GetGCRToken` method. Retries were successful because the cache was read from in a different location.

Issue #1069, opened 2021-08-18
porter run interactive shell doesnt wrap text or resize to terminal


joegilley:
# Description

![45bc8ba8e98e1e713a8a5c85a3f1cab2](https://user-images.githubusercontent.com/1476969/129817261-20a260fa-4551-4620-aeff-918bcdc1cdd3.gif)

# Location

- [ ] Browser
- [x] CLI
- [ ] API

tested on 90f60a5c766ca735a8d04e5fc848c3d53ee58429

# Steps to reproduce

1. Run `porter run anyreleasewithshell bash`
1. type a bunch of stuff

# Additional Details

## Expected behavior

<img width="517" alt="image" src="https://user-images.githubusercontent.com/1476969/129817354-586f6095-7840-40bb-811a-faf125f085bd.png">

expect content to wrap, as it does in `kubectl run`

abelanger5:
Closed by #1070 -- released in `v0.8.0`!

Issue #1032, opened 2021-08-12
Unable to get interactive prompt when running `rails c`


nikokon:
# Description
I’m trying to run a rails console but it seems to exit and delete the pod without prompting for input. Running against a running pod using -e is working as usual.

When debugging the issue further I notice that it was working in the "default" namespace but neither "development" nor "nightly" where the same app was living.

# Location

- [ ] Browser
- [x] CLI
- [ ] API

# Steps to reproduce

1. porter run --namespace <some_namespace_other_than_default> <rails_app_name> -- bin/rails c
2. await it to start pod, print the logs and exit

# Additional Details
porter logs:
```
Running bin/rails c for release plick-api
Attempting connection to the container, this may take up to 10 seconds. If you don't see a command prompt, try pressing enter.
Could not open a shell to this container. Container logs:

<the usual logs printed when starting rails console>

Sucessfully deleted ephemeral pod

```

nikokon:
I was thinking it might be a memory constraint but I just noticed that the nightly namespace actually had more memory defined than the default.

abelanger5:
Hey @nikokon, thanks for the bug report! I've just added the `--verbose` option to `porter run` in version `v0.7.2` of the CLI -- could you give it a try and see if that helps debug? This will print the events associated with the ephemeral pod, which should give us some insight. 

There could be multiple issues here, it seems that if the logs are getting printed you're most likely running into a memory constraint. 

In the case that the logs aren't getting printed, one potential limitation is when the cluster doesn't have enough resources to create the ephemeral pod with the assigned resources, the cluster may trigger a scale-up of nodes. However, this doesn't happen in 10-second window that the ephemeral pod is allocated, and so the pod gets deleted before the new node is ready.

nikokon:
Sorry for the delay, been off the computer for a few days. So, I've been running the command a few times now running CLI version 0.7.4
succeeded a couple of times but mostly failed been running only in default namespace so that's probably not an issue, however it does feel like a resource / timing issue. Here's the output from a couple of runs (I just blanked out some urls for good messure :shrug:)

```
=> porter run --verbose plick-api -- bin/rails c
Running bin/rails c for release plick-api
Attempting connection to the container, this may take up to 10 seconds. If you don't see a command prompt, try pressing enter.
Could not open a shell to this container. Container logs:

Could not get logs. Pod events:

0/21 nodes are available: 13 Insufficient cpu, 8 Insufficient memory.
[...]

PR #1070, opened 2021-08-18
fix: connect term sizing to stream io

joegilley:
## Pull request type

- [x] Bugfix
- [ ] Feature

## Pull request checklist

Tests aren't passing on master right now for me, not sure if its a setup problem or HEAD is broken.
Feel free to commandeer, add tests, change logic, make it your own, whatever you like :)

Issue Number: #1069

## What is the new behavior?

Interactive shells now use the full terminal width and overdraws correctly continue to the next line.

Also I added an event watcher to wait for the new pod to be ready instead of polling.


cli/cmd/run.go
```diff
@@ -277,28 +278,51 @@ func executeRun(config *PorterRunSharedConfig, namespace, name, container string
...
 
+	newPod, err := createPodFromExisting(config, existing, args)
 	podName := newPod.ObjectMeta.Name
 
-	t := term.TTY{
-		In:  os.Stdin,
-		Out: os.Stdout,
-		Raw: true,
-	}
+	err = waitForPod(config, newPod)
```
abelanger5, line 322: We may want to print a message above this line to the user indicating that we're waiting for the pod to be ready. In cases where the pod is unscheduleable (see #1032 for reference) the user will be waiting for up to 30 seconds.

abelanger5 (changes requested):
Hey @joegilley thanks for this PR -- this is working great! 

Could you also add terminal size queue logic to the `executeRun` function (not just `executeRunEphemeral`)? The difference is essentially the same as `kubectl`'s [`exec`](https://github.com/kubernetes/kubectl/blob/567f2b8c4b5cd2a5b8688f6436ba44ae4b09a80d/pkg/cmd/exec/exec.go#L123) and  [`attach`](https://github.com/kubernetes/kubectl/blob/567f2b8c4b5cd2a5b8688f6436ba44ae4b09a80d/pkg/cmd/attach/attach.go#L161). 

Otherwise, this should be good to include in release `0.8.0`!

joegilley:
* add helpful prints during wait
* fix multiple race conditions:
  * if the pod is ready before we start listening to events.
  * if the pod exits before we start listening to events.
  * if k8s doesnt know about the new pod yet when we ask for a listener

abelanger5 (approved):
This is great @joegilley, much cleaner implementation. 

> Tests aren't passing on master right now for me, not sure if its a setup problem or HEAD is broken.

Just to clarify, HEAD is broken and we're only testing specific packages at the moment (sorry about that). There's a large backend refactoring which will hopefully be completed soon, in which case all tests will work again.

PR #1105, opened 2021-08-24
[0.9.0] - Refactored base api

jnfrati:
## Pull request type

- [x] Other (please describe):

## Pull request checklist

There is no way to type API response without using the `as` helper

## What is the new behavior?

Modified the baseApi file so it's more clear and returns functions that will allow devs to type out the responses of the API

- Modified file extension (from tsx to ts)
- Using axios builder instead of axios.{method} for creating the axios instance


abelanger5 (approved):
Awesome @jnfrati, working well!

PR #1112, opened 2021-08-24
[0.9.0] [Frontend, Forms] Remove defaults

### LINKED ISSUE #1031: Unable to set input value back to empty on form submit
# Description

Ex: if a start command is specified for the web template, the start command can't be unset (seems like if the input is empty it's just using the last value instead of actually submitting the empty string as the value)

# Location

- [X] Browser
- [ ] CLI
- [ ] API

# Steps to reproduce

1. Deploy a web chart with a start command set
2. Attempt to unset the start command
3.

# Additional Details

### LINKED ISSUE #1060: Form field value field causing problems
# Description

The `value` field passed to each form field from the server causes problems, as it makes it so there's no single source of truth for what the variable values should be (as there's also an initial variable list)

# Location

- [x] Browser
- [ ] CLI
- [ ] API

# Steps to reproduce

This is related to issues #1057 and #1031 

# Additional Details

Right now, every form field gets a `value` parameter from the server (which is an array whose first element is the value of the field). This value is supposed to be used to keep the state of the form between reloads of the frontend. The issue right now is how this `value` is populated on the server:
* The `value` should **only be set when there is a value to be set**. That is, if we have a string input, it should not recieve `value: [""]` from the server unless its value is actually the empty string. 
* This is actually what caused the bug with deleting fields that have defaults. The frontend currently ignores `value: [""]` (since `""` evaluates to false in Javascript and every string field receives this implicitly), so when a field is erased like that it doesn't actually register due to the frontend logic.

igalakhov:
## Pull request type

- [x] Bugfix

## Pull request checklist

There is collision between `field.value` and `field.default` which causes unwanted bugs.

## What is the new behavior?

There is no more `field.default` and `field.value` is the single source of truth for the starting state of each field.

 Only issue is that this breaks the form debugger so that needs to be somehow taken care of in this PR as well (perhaps by populating values using the default originally or something).

closes #1060 
closes #1031

### DIFF
--- a/dashboard/src/components/porter-form/field-components/ArrayInput.tsx
+++ b/dashboard/src/components/porter-form/field-components/ArrayInput.tsx
@@ -6,13 +6,14 @@ import {
   GetFinalVariablesFunction,
 } from "../types";
 import useFormField from "../hooks/useFormField";
+import { hasSetValue } from "../utils";
 
 const ArrayInput: React.FC<ArrayInputField> = (props) => {
   const { state, variables, setVars } = useFormField<ArrayInputFieldState>(
     props.id,
     {
       initVars: {
-        [props.variable]: props.value && props.value[0] ? props.value[0] : [],
+        [props.variable]: hasSetValue(props) ? props.value[0] : [],
       },
     }
   );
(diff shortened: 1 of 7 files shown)


abelanger5 (approved):
Looks to be working well @igalakhov, thanks! Tested that new charts deploy properly and that fields can be unset (like the start command).

PR #1123, opened 2021-08-24
[0.9.0] Handle namespaces when deploying through GHA

abelanger5:
## Pull request type

- [X] Bugfix
- [ ] Feature

## Pull request checklist

Namespaces aren't supported when deploying through GHA at the moment. 

## What is the new behavior?

Support namespaces in GHA. 

Since the namespace isn't set by the time the GHA preview is rendered, we'll just show `default` for now in the GHA file, so that when users copy the workflow file, they can change the namespace.


dashboard/src/main/home/launch/launch-flow/LaunchFlow.tsx
```diff
@@ -329,6 +329,7 @@ const LaunchFlow: React.FC<PropsType> = (props) => {
       return (
         <WorkflowPage
           name={templateName}
+          namespace={"default"}
```
anukul, line 332: Just noticed a small inconsistency:

image

The Settings page is the first chance the user gets to select a namespace, and it shows up after the Github Actions page.

If the user selects a non-`default` namespace, we do push the right workflow, but the preview would show `default`. 🤔

abelanger5: Yep, good catch! I mentioned this up in the description of the PR:

> Since the namespace isn't set by the time the GHA preview is rendered, we'll just show default for now in the GHA file, so that when users copy the workflow file, they can change the namespace.

anukul: Ohh sorry I just jumped to the diff :s

anukul (approved):

PR #1184, opened 2021-09-07
[0.9.0] Set build context default to Dockerfile dir

anukul:
## Pull request type

- [ ] Bugfix
- [ ] Feature
- [x] Other (please describe): Spec change

## Pull request checklist

Docker build context default is set to repository root

## What is the new behavior?

Docker build context default is set to directory containing Dockerfile

TODO: Allow setting build context from dashboard

### DIFF
--- a/cli/cmd/create.go
+++ b/cli/cmd/create.go
@@ -227,10 +227,6 @@ func handleSubdomainCreate(subdomain string, err error) error {
 }
 
 func createFromGithub(createAgent *deploy.CreateAgent, overrideValues map[string]interface{}) error {
-	if localPath == "" {
-		localPath = filepath.Dir(dockerfile)
-	}
-
 	fullPath, err := filepath.Abs(localPath)
 
 	if err != nil {
--- a/cli/cmd/create.go
+++ b/cli/cmd/create.go
@@ -93,7 +93,7 @@ func init() {
 		"path",
 		"p",
 		"",
-		"if local build, the path to the build directory",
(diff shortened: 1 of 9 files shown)


cli/cmd/create.go
```diff
@@ -223,6 +227,10 @@ func handleSubdomainCreate(subdomain string, err error) error {
 }
 
 func createFromGithub(createAgent *deploy.CreateAgent, overrideValues map[string]interface{}) error {
+	if localPath == "" {
```
abelanger5, line 230: Is `createFull` the only method that calls `createFromGithub`? If so, we can remove this right?

abelanger5 (changes requested):
Looks good @anukul, just one minor thing to resolve 👍

cli/cmd/create.go
anukul: That's right.. fixed, thanks.

igalakhov:
Hi @anukul - seems like everything is working well! One slight problem is that (as mentioned in issue #1186) the auto buildpack doesn't work for folders for some reason. I get this error specifically:
> Error: invalid app path 'src/': evaluate symlink: lstat src: no such file or directory
After some investigation, it seems like this happens due to Docker - I suspect something with symlinks is done to build and the way we pull/store the repository breaks things (note: I tested with absolute paths and that has the same issue). The exact line of code that fails is this one (this is in library code):
```go
if resolvedAppPath, err = filepath.EvalSymlinks(appPath); err != nil {
		return "", errors.Wrap(err, "evaluate symlink")
	}
```

PR #1191, opened 2021-09-13
[POR-27] Informers fail to listen once token expires

igalakhov:
## Pull request type

- [x] Bugfix
- [ ] Feature

After a token expires (this happens relatively quickly on AWS), the WebSockets start failing and producing errors:

image

## What is the new behavior?

Currently investigating why this happens, probably just need some way to refresh the token on expiration.

### DIFF
--- a/server/api/k8s_handler.go
+++ b/server/api/k8s_handler.go
@@ -2,11 +2,8 @@ package api
 
 import (
 	"encoding/json"
+	"errors"
 	"fmt"
-	"net/http"
-	"net/url"
-	"strconv"
-
 	"github.com/go-chi/chi"
 	"github.com/gorilla/schema"
 	"github.com/gorilla/websocket"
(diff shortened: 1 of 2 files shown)


server/api/k8s_handler.go
```diff
@@ -1152,18 +1152,35 @@ func (app *App) HandleStreamControllerStatus(w http.ResponseWriter, r *http.Requ
...
-	// get path parameters
-	kind := chi.URLParam(r, "kind")
+	lastTime := int64(0)
 
-	selectors := ""
-	if vals["selectors"] != nil {
-		selectors = vals["selectors"][0]
-	}
-	err = agent.StreamControllerStatus(conn, kind, selectors)
+	for {
```
abelanger5, line 1157: Can we move this for loop and auth error logic to the underlying agent method, i.e. `StreamControllerStatus`? The handler should be isolated from this type of implementation-specific logic.

abelanger5 (changes requested):
@igalakhov looks great, just one requested change!

abelanger5 (approved):

PR #1192, opened 2021-09-13
[POR-30] Clone job feature

jnfrati:
## Pull request type

- [x] Feature

## Pull request checklist

## Motivation
The idea is to provide a simple way to launch a new job copying the configurations from a previous one. 

From a UI perspective, it would be placing a simple button with a description for the user to be redirected to the launch flow.
image

From the code perspective, the Clone job button will redirect the user to the launch tab with all the configurations needed on the URL as query params, the only config that will not be copied will be the environment variables as they can contain sensitive information. This will also set the basis for adding other chart redirections with pre-filled data (e.g. one link deploy for mongo db in which the user will get all the data prefilled for a fast deployment)


jnfrati:
As an update, we ended up deciding to pivot a little bit the perspective:

- First of all, we're going to retrieve the desired release to clone from the API instead of directly passing it via URL search/query params.
- Github deployed apps are not going to be supported for this feature. As we have to think about how the user will be able to customize the GHAction workflows (they may want to preserve the same workflow using the same last image, or maybe create a new workflow with customized options), until we decide that, the idea is to provide the clone feature for Docker deployments.

jusrhee (approved):

PR #1193, opened 2021-09-13
[POR-31] Fix large number of logs crashing frontend

igalakhov:
## Pull request type

- [x] Bugfix
- [ ] Feature

A large number of logs causes the frontend to crash.

## What is the new behavior?

The number of logs is limited to 1000.

A bit harder of a change than one would expect because React list key issues


jnfrati (approved):
Looks great! 👏

PR #1199, opened 2021-09-14
[POR-34] API migration + refactor

abelanger5:
## Pull request type

- [ ] Bugfix
- [ ] Feature
- [X] Improvement

## Pull request checklist

## Goals

- Strongly typed API, see here for an example: https://github.com/porter-dev/porter/blob/0.8.0-api-cleanup/api/types/user.go -- allows us to really easily implement the Golang client for the CLI, and gives us strong type agreement between the frontend and backend (i.e. typescript can just copy these API types) 

- Log to the appropriate scope -- i.e. automatically including the referenced objects in the logs, like user_id, project_id, cluster_id, etc, which we can then pipe into LogDNA and construct search queries for. So to investigate a user's error, we can just construct a search query for that user/project (this was done by abstracting API error logging to a separate package), and view a list of all operations performed on a specific cluster by a user. 

- Send detailed internal server errors to sentry + notify us. the issue with the current backend is that 90% of our errors are just reported as "internal server error," when they're really not.

- Give us the ability to automatically generate an API spec that we can publish (combination of the strong typing and each handler using APIRequestMetadata https://github.com/porter-dev/porter/blob/eab3c2951e67750d329b9140fe83418fd10c10d8/api/types/request.go#L47 ) 

- Allow backend unit tests to pass again, so that we can set up a CI workflow to test the backend before merging to master.

Consolidated API into `api` package:
- `api/types` contains the request/response types, along with API type definitions
- `api/client` contains the Golang client (this does not implement the whole API, only the subset that the CLI uses)
- `api/server` implements the server routes and handler logic


abelanger5:
- [ ] Selecting namespace `All` breaks the dashboard, because the endpoint fails

abelanger5:
- [ ] re-add analytics to handlers 
- [ ] `.../accounts` endpoint for Github app does not show anything on FE
- [ ] branch with slash throws error (again)
- [ ] `app` field not populated in GHA preview
- [ ] when getting GHA template, `should_create_workflow` incorrectly passed/populated
- [ ] namespace doesn't seem to be passed through via GHA
- [ ] `image_url` field doesn't seem to be passed correctly on deploy (passed as hello-porter images)

abelanger5:
- [ ] email verification flow has inconsistent behavior

PR #1201, opened 2021-09-15
[POR-36] Token cache slows down dashboard

igalakhov:
## Pull request type

- [x] Bugfix
- [ ] Feature

Token cache slows down the dashboard because gorm one-to-one selects more data than it should be.

## What is the new behavior?

One-to-one is now done manually, so things should be faster.


abelanger5 (changes requested):
Hey @igalakhov -- looks great, there's one minor thing (not sure if this somehow has to do with this fix, or with `gorm`). But the `created_at` time in the DB for the new token caches is `0001-01-01 00:00:00+00`, while it used to be written correctly. Not sure if you're seeing a similar thing locally?

These `created_at` fields aren't currently used, but we might want to use this field for various things in the future, so just wanted to check.

abelanger5 (approved):

PR #1405, opened 2021-11-09
Option to use Heroku as well as Paketo buildpacks + detection for Go, NodeJS, Python, and Ruby projects

mnafees:
## Pull request type

- [ ] Bugfix
- [x] Feature

## Pull request checklist

Currently, the backend handler to detect runtimes just cycles through common files to choose a runtime.

## What is the new behavior?

We now try to auto-detect 4 runtimes including Go, NodeJS, Python, and Ruby by using the detection algorithms directly from their respective Paketo buildpacks.

We also now have the option to use either Heroku or Paketo builders and their buildpacks.

- Use the GitHub project to detect the correct runtime (Go, NodeJS, Python, or Ruby) for a project
- Include this information in the releases information
- In the CLI, directly import the aforementioned build configuration information and use the specified builder and buildpacks through `pack`
- If no runtime was detected then simply use the builder that was selected


api/server/handlers/gitinstallation/get_buildpack.go
```diff
@@ -71,34 +91,23 @@ func (c *GithubGetBuildpackHandler) ServeHTTP(w http.ResponseWriter, r *http.Req
...
-		"Gemfile":          "Ruby",
-		"package.json":     "Node.js",
-		"pom.xml":          "Java",
-		"composer.json":    "PHP",
+	builderInfoMap := initBuilderInfo()
+	var wg sync.WaitGroup
+	wg.Add(len(buildpacks.Runtimes))
+	for i := range buildpacks.Runtimes {
+		go func(idx int) {
+			buildpacks.Runtimes[idx].Detect(
```
abelanger5, line 99: If the `Detect` function stalls and the handler exits, do we have a mechanism to cancel the detect methods? We might want to propagate a `context.Context` and pass `context.WithCancel` to the child processes.

api/types/release.go
```diff
@@ -45,6 +46,7 @@ type CreateReleaseRequest struct {
 
 	ImageURL           string                        `json:"image_url" form:"required"`
 	GithubActionConfig *CreateGitActionConfigRequest `json:"github_action_config,omitempty"`
+	BuildConfig        *CreateBuildConfigRequest     `json:"build_config" form:"required"`
```
abelanger5, line 49: It doesn't seem like this field should be required -- for example when deploying from a Docker image. Might want to use ``json:"build_config,omitempty"`` and case on when BuildConfig is nil in the CreateRelease handler?

abelanger5 (changes requested):

api/types/release.go
mnafees: Yes, this makes sense.

api/server/handlers/gitinstallation/get_buildpack.go
mnafees: As discussed - this will be done in a future PR with some refactoring. Will link back to this comment for visibility.

abelanger5 (approved):
Great job @mnafees 💯

PR #1424, opened 2021-11-16
[POR-11] Cert-manager applet for certificate visualization and renewal

abelanger5:
## Pull request type

- [ ] Bugfix
- [X] Feature

## Pull request checklist

There is no mechanism for user's to view or renew certificates from the Porter dashboard.

## What is the new behavior?

Provide a mechanism to view the status of or renew certificates on the dashboard. 

image

- This is implemented as a "built-in" form on the Porter dashboard. If we detect a `cert-manager` Helm chart, we render this view. This may not work for all `cert-manager` instances, but is fully supported from the clusters that Porter provisions.


dashboard/src/components/ExpandableResource.tsx
```diff
@@ -1,58 +1,105 @@
...
-    );
-  }
-}
-
-ExpandableResource.contextType = Context;
+const ExpandableResource: React.FC<Props> = (props) => {
+  const { resource, button } = props;
+  const { currentCluster, currentProject } = useContext(Context);
+
+  let onSave = () => {
```
jnfrati, line 22: Just to standardize declaration, it would be nice to use the `const` keyword here instead of let to declare internal component functions

```suggestion
  const onSave = () => {
```

dashboard/src/components/ExpandableResource.tsx
```diff
@@ -1,58 +1,105 @@
-import React, { Component } from "react";
+import React, { Component, useContext, useEffect } from "react";
 import styled from "styled-components";
 import { Context } from "shared/Context";
-
+import { useWebsockets } from "shared/hooks/useWebsockets";
```
jnfrati, line 4: Is this being used for this component?

```suggestion
```

jnfrati (commented):
Not critical but would be nice to apply! Nice job on the components refactor 👏

dashboard/src/components/ExpandableResource.tsx
abelanger5: Nope, nice catch!

PR #1458, opened 2021-11-24
[POR-231] Query pod last status on pod level events

jnfrati:
## Pull request type

- [ ] Bugfix
- [ ] Feature

## Pull request checklist

## Motivation

Sub event list may not show the actual status of a pod, that's why when we render the sub event list for a specific pod, we try to query for the pod again and check the last status that has been seen


internal/kubernetes/agent.go
```diff
@@ -518,6 +518,22 @@ func (a *Agent) GetPodsByLabel(selector string, namespace string) (*v1.PodList,
...
+	// Get pod by name
+	pod, err := a.Clientset.CoreV1().Pods(namespace).Get(
+		context.TODO(),
+		name,
+		metav1.GetOptions{},
+	)
+
+	if err != nil && errors.IsNotFound(err) {
+		return nil, IsNotFoundError
+	}
```
abelanger5, line 532: Add another case for just `err != nil` and return the error directly.

abelanger5 (changes requested):

abelanger5 (approved):

PR #1473, opened 2021-11-29
use cache from remote registry

sunguroku:
## Pull request type

- [ ] Bugfix
- [ ] Feature
- [x] Other (please describe): CLI now uses the remote registry as the source of cache to speed up build times.

## Pull request checklist

## What is the new behavior?
CLI uses docker cache during builds

### DIFF
--- a/cli/cmd/docker/builder.go
+++ b/cli/cmd/docker/builder.go
@@ -21,6 +21,7 @@ import (
 type BuildOpts struct {
 	ImageRepo         string
 	Tag               string
+	CurrentTag		  string
 	BuildContext      string
 	DockerfilePath    string
 	IsDockerfileInCtx bool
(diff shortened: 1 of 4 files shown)


cli/cmd/docker/builder.go
```diff
@@ -67,6 +68,9 @@ func (a *Agent) BuildLocal(opts *BuildOpts) error {
 		Tags: []string{
 			fmt.Sprintf("%s:%s", opts.ImageRepo, opts.Tag),
 		},
+		CacheFrom: []string{
```
abelanger5, line 71: Only add `CacheFrom` when `opts.ImageRepo` or `opts.CurrentTag` are not empty -- otherwise we're adding an unnecessary step which could error out right?

abelanger5 (changes requested):
Can you run `gofmt` on the changed files?

PR #1477, opened 2021-11-29
[POR-245] Use the build-plan buildpack when using the Paketo builders

mnafees:
## Pull request type

- [ ] Bugfix
- [x] Feature

## Pull request checklist

We currently do not support having `plan.toml` in projects. The CLI currently defaults to the `paketo` builder.

## What is the new behavior?

This PR introduces the buildpack that supports having `plan.toml` files in project root directories into the CLI whenever using the `paketo` builders. It also changes the default builder from `paketo` to `heroku`.

### DIFF
--- a/cli/cmd/pack/pack.go
+++ b/cli/cmd/pack/pack.go
@@ -9,6 +9,7 @@ import (
 	"github.com/buildpacks/pack"
 	"github.com/porter-dev/porter/api/types"
 	"github.com/porter-dev/porter/cli/cmd/docker"
+	"github.com/porter-dev/porter/internal/integrations/buildpacks"
 )
 
 type Agent struct{}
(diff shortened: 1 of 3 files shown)


cli/cmd/pack/pack.go
```diff
@@ -46,5 +48,9 @@ func (a *Agent) Build(opts *docker.BuildOpts, buildConfig *types.BuildConfig) er
 		// FIXME: use all the config vars
 	}
 
+	if strings.HasPrefix(buildOpts.Builder, "paketo") {
```
abelanger5, line 51: Why are we checking for the `paketo` prefix here?

mnafees: The buildpack that supports `plan.toml` is Paketo-specific so that is why we check for this string.

PR #1507, opened 2021-12-03
[POR-252] Initial version of Preview Apps CLI

mnafees:
## Pull request type

- [ ] Bugfix
- [x] Feature

## Pull request checklist


cli/cmd/apply.go
```diff
@@ -0,0 +1,340 @@
...
+
+func (d *Driver) ShouldApply(resource *models.Resource) bool {
+	return true
+}
+
+func (d *Driver) Apply(resource *models.Resource) (*models.Resource, error) {
+	client := GetAPIClient(config)
+
+	// TODO: use source.repo, source.version
+	config.SetProject(d.target.Project)
```
abelanger5, line 137: This will set the project/cluster in the config file which seems like a bit of an unpredictable side effect.

cli/cmd/apply.go
```diff
@@ -0,0 +1,340 @@
...
+		}
+	}
+
+	// lastly, just put in the defaults
+	if source.Name == "" || source.Repo == "" || source.Version == "" {
+		// default to these values when any one of the source values are empty
+		// this makes sense because it might save us from version mismatches and other mishaps
+		source.Name = "web"
+		source.Repo = "https://charts.getporter.dev"
+		source.Version = "v0.13.0"
```
abelanger5, line 262: I believe we can actually set the `source.Version` default to `"latest"` so we don't need to constantly update this -- might require some testing though. 

I think we should require the `source.Name` field at the moment, and just default the `source.Repo` and `source.Version` independently. So this might look more like:

```go
if source.Name == "" {
   return nil, fmt.Errorf("source name required")
}

if source.Repo == "" {
  source.Repo = "https://charts.getporter.dev"
}

if source.Version == "" {
  source.Version = "latest"
}
```

cli/cmd/apply.go
```diff
@@ -0,0 +1,340 @@
...
+	}
+
+	basePath, err := os.Getwd()
+
+	if err != nil {
+		return err
+	}
+
+	worker := worker.NewWorker()
+	worker.RegisterDriver("", NewPorterDriver) // FIXME: workaround for when driver is not present
```
abelanger5, line 64: Could this be fixed by a PR into the switchboard repo?

cli/cmd/apply.go
```diff
@@ -0,0 +1,340 @@
...
+	if err != nil {
+		// create new app
+		name = resource.Name
+
+		regList, err := client.ListRegistries(context.Background(), config.Project)
+		if err != nil {
+			return nil, err
+		}
+		if len(*regList) > 0 {
+			registryURL = (*regList)[0].URL
```
abelanger5, line 186: One mistake we made early on was the use of globals everywhere for the CLI package, which was the recommended approach in the cobra examples but in hindsight makes these variables very unpredictable. This should be fine for now, but I just want to flag that the behavior of setting a global deep inside nested functions is an antipattern. There are many other methods in the CLI guilty of this though.

cli/cmd/apply.go
```diff
@@ -0,0 +1,340 @@
...
+	"github.com/rs/zerolog"
+	"github.com/spf13/cobra"
+)
+
+// applyCmd represents the "porter apply" base command when called
+// with a porter.yaml file as an argument
+var applyCmd = &cobra.Command{
+	Use:   "apply",
+	Short: "Applies the provided porter.yaml to a project",
+	Run: func(cmd *cobra.Command, args []string) {
```
abelanger5, line 28: Make sure to add a `Long` description in the style of the `create/update` commands. `kubectl apply -h` is also a great reference here.

abelanger5 (changes requested):

jnfrati (approved):

abelanger5 (dismissed):

PR #1614, opened 2022-01-19
[POR-129] Show available new version of the CLI

mnafees:
## Pull request type

- [ ] Bugfix
- [ ] Feature
- [x] Other (please describe): Enhacement

## Pull request checklist

N/A

## What is the new behavior?

We now query for and show a message when a new CLI version is available to download.

### DIFF
--- a/cli/cmd/root.go
+++ b/cli/cmd/root.go
@@ -1,9 +1,15 @@
 package cmd
 
 import (
+	"context"
+	"fmt"
 	"os"
+	"strings"
+	"time"
 
+	"github.com/Masterminds/semver/v3"
 	"github.com/fatih/color"
+	"github.com/google/go-github/v41/github"
 	api "github.com/porter-dev/porter/api/client"
 	"github.com/spf13/cobra"
 	"k8s.io/client-go/util/homedir"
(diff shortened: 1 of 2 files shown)


cli/cmd/root.go
```diff
@@ -25,6 +31,26 @@ func Execute() {
...
+		release, _, err := ghClient.Repositories.GetLatestRelease(ctx, "porter-dev", "porter")
+		if err == nil {
+			release.GetURL()
+			// we do not care for an error here because we do not want to block the user here
+			constraint, err := semver.NewConstraint(fmt.Sprintf("> %s", strings.TrimPrefix(Version, "v")))
+			if err == nil {
+				latestRelease, err := semver.NewVersion(strings.TrimPrefix(release.GetTagName(), "v"))
+				if err == nil {
+					if constraint.Check(latestRelease) {
+						color.New(color.FgYellow).Println("A new version of the porter CLI is available to download at https://github.com/porter-dev/porter/releases/latest")
```
mnafees, line 47: This should eventually show a message asking the user to run 
```sh
$ /bin/bash -c "$(curl -fsSL https://raw.githubusercontent.com/porter-dev/porter/HEAD/install.sh)"
```
to upgrade the CLI

abelanger5 (changes requested):
The issue here is that this runs on every CLI command, not all of which are interactive. For example, this will print on `porter update config get-env`, which is meant to print out the environment variables so people can use them in workflows. Could we figure out a way to annotate certain commands and flag combinations so that this isn't printed when we mark a command as "non-interactive"?

cli/cmd/root.go
```diff
@@ -25,6 +32,32 @@ func Execute() {
...
+				latestRelease, err := semver.NewVersion(strings.TrimPrefix(release.GetTagName(), "v"))
+				if err == nil {
+					if constraint.Check(latestRelease) {
+						color.New(color.FgYellow).Fprint(os.Stderr, "A new version of the porter CLI is available. Run the following to update: ")
+						if runtime.GOOS == "darwin" {
+							color.New(color.FgYellow).Add(color.Bold).Fprintln(os.Stderr, "brew install porter-dev/porter/porter")
+						} else {
+							color.New(color.FgYellow).Add(color.Bold).Fprintln(os.Stderr, "/bin/bash -c \"$(curl -fsSL https://install.porter.run)\"")
+						}
+
```
abelanger5, line 54:
```suggestion
View CLI installation and upgrade docs here: https://docs.porter.run/cli/installation
```

Perhaps we should add a line linking to the installation docs, just in case they didn't install with `brew` on Mac or they haven't run the new install command?

abelanger5 (changes requested):

abelanger5 (approved):

PR #1637, opened 2022-01-31
[POR-315] Do not build an image if one already exists in the registry

mnafees:
## Pull request type

- [ ] Bugfix
- [ ] Feature

## Pull request checklist

Currently we build images in the registry even if we already have the same built and tagged.

## What is the new behavior?

Do not build already existing images with the same name and tag.


cli/cmd/docker/agent.go
```diff
@@ -158,6 +158,19 @@ var PullImageErrNotFound = fmt.Errorf("Requested image not found")
 
 var PullImageErrUnauthorized = fmt.Errorf("Could not pull image: unauthorized")
 
+// CheckIfImageExists checks if the image exists in the registry
+func (a *Agent) CheckIfImageExists(image string) error {
```
abelanger5, line 162: I think the signature of this should be `func CheckIfImageExists(image string) (bool, error)` where the bool indicates existence, rather than throwing an error if it does not exist. This makes it easier to case when the method is called, so the calling method doesn't need to know the "image not found" error type.

abelanger5 (changes requested):
A few broader changes/discussions:
- If the image tag is `latest`, the user likely wants it overwritten each time. 
- We should add an option to force new builds, this could be something like `--build` for the `create` and `update` commands and `forceBuild` for apply.
- We should print to the console when we're skipping a build, so if users are confused by the behavior they can see what's going on.

abelanger5 (approved):

PR #1638, opened 2022-01-31
Create separate provisioner service for infrastructure

abelanger5:
## Pull request type

- [ ] Bugfix
- [X] Feature

## Pull request checklist

Infrastructure management on Porter currently does not support:
- Upgrading existing infrastructure from the dashboard 
- Viewing the deployment history for infrastructure, and reverting to previous revisions
- Supporting programmatic and custom infrastructure via forms (just like Helm charts)

## What is the new behavior?

To resolve the following issues:
- Support upgrades of existing infrastructure from the dashboard 
- View a history of operations for each infrastructure
- Support rendering `form.yaml` for existing infra

Some preliminary screenshots (UNSTYLED):

image

image


dashboard/src/main/home/infrastructure/components/InfraResourceList.tsx
```diff
@@ -0,0 +1,145 @@
...
+  }
+
+  if (hasError) {
+    return <Placeholder>Error</Placeholder>;
+  }
+
+  console.log(infraState);
+
+  const renderContents = () => {
+    return Object.keys(infraState.resources).map((key) => {
```
jnfrati, line 60: Always make sure that you're passing an object, if by any change infraState.resources is filled with anything besides an object this will directly throw an object.
A simple solution would be to use:
```suggestion
    return Object.keys({...(infraState?.resources || {})}).map((key) => {
```

dashboard/src/main/home/infrastructure/components/InfraResourceList.tsx
```diff
@@ -0,0 +1,145 @@
...
+      return (
+        <RepoName lastItem={false} isSelected={false}>
```
jnfrati, line 62: Add a key to help react render big list items and avoid errors/warnings on console

```suggestion
        <RepoName key={key} lastItem={false} isSelected={false}>
```

dashboard/src/main/home/infrastructure/components/DeployList.tsx
```diff
@@ -0,0 +1,212 @@
...
+  }
+
+  const getOperationDescription = (type: OperationType): string => {
+    switch (type) {
+      case "create":
+        return "This infrastructure was created";
+      case "update":
+        return "The infrastructure configuration was updated";
+      default:
+        return "The infrastructure configuration was updated";
```
jnfrati, line 69: Super minor but we can avoid an unnecessary return here.

```suggestion
      case "update":
      default:
        return "The infrastructure configuration was updated";
```

dashboard/src/main/home/infrastructure/components/DeployList.tsx
```diff
@@ -0,0 +1,212 @@
...
+        setOperationList(data);
+        setIsLoading(false);
+      })
+      .catch((err) => {
+        console.error(err);
+        setHasError(true);
+        setCurrentError(err.response?.data?.error);
+        setIsLoading(false);
+      });
+  }, [currentProject]);
```
jnfrati, line 44: Remember adding the infra_id to refresh the operations when it changes 

```suggestion
  }, [currentProject, infra_id]);
```

dashboard/src/main/home/infrastructure/ExpandedInfra.tsx
```diff
@@ -0,0 +1,185 @@
...
+      case "deploys":
+        return <DeployList infra_id={infra_id} />;
+      case "resources":
+        return <InfraResourceList infra_id={infra_id} />;
+      case "settings":
+        return <div>Settings</div>;
+    }
+  };
+
+  const formData = yaml.load(initYaml);
```
jnfrati, line 79: Not sure if we should encapsulate this one directly on a useMemo as I'm not sure how expensive can be to load the yaml on each rerender.

dashboard/src/main/home/infrastructure/ExpandedInfra.tsx
```diff
@@ -0,0 +1,185 @@
...
+
+  return (
+    <StyledExpandedInfra>
+      <Header
+        last_updated={readableDate(infra.latest_operation?.last_updated)}
+        back_link={"/infrastructure"}
+        name={integrationList[infra.kind].label}
+        icon={integrationList[infra.kind].icon}
```
jnfrati, line 87: Just in case, can we add a null casing here (as well as other places using the integrationList)? Wouldn't be the first time where the component cannot render because of the index being undefined/null. 

```suggestion
        name={integrationList[infra.kind]?.label}
        icon={integrationList[infra.kind]?.icon}
```

dashboard/src/main/home/infrastructure/InfrastructureList.tsx
```diff
@@ -0,0 +1,333 @@
...
+          columns={columns}
+          data={infraList}
+          isLoading={isLoading}
+          onRowClick={(row) => {
+            let original = row.original as Infrastructure;
+            pushFiltered(
+              { history, location },
+              `/infrastructure/${original.id}`,
+              ["project_id"]
+            );
```
jnfrati, line 194: You can get a populated version of pushFiltered using the useRouting hook (this is exported from the same file as the common pushFiltered). The only reason the old pushFiltered function is still exported is for backwards compatibility with class components.

dashboard/src/main/home/infrastructure/InfrastructureRouter.tsx
```diff
@@ -0,0 +1,21 @@
...
+import InfrastructureList from "./InfrastructureList";
+import ExpandedInfra from "./ExpandedInfra";
+
+const InfrastructureRouter = () => {
+  return (
+    <Switch>
+      <Route
+        path="/infrastructure/:infra"
+        render={({ match }) => {
+          return <ExpandedInfra infra_id={parseInt(match.params["infra"])} />;
```
jnfrati, line 13: Instead of using the render function, we can use the [useParams hook](https://v5.reactrouter.com/web/api/Hooks/useparams), this will encapsulate all the logic inside the component and the router will only be responsible for rendering components.

You can find some example of how we're currently using it over [here](https://github.com/porter-dev/porter/blob/master/dashboard/src/main/home/cluster-dashboard/dashboard/node-view/ExpandedNodeView.tsx#L28)

jnfrati (changes requested):
I left some comments on super minor items, overall it looks awesomely good! :clap:

abelanger5:
Looks like this was branched from `dev`, closing to use a branch with a much cleaner commit history.

PR #1684, opened 2022-02-21
Add `--force-push` to `porter update` to force push an image

mnafees:
## Pull request type

- [ ] Bugfix
- [ ] Feature

## Pull request checklist

Right now we try to push all images even when they are not built.

## What is the new behavior?

We now only push images when either `--force-push` is unset or image tag is `latest`.


cli/cmd/deploy/deploy.go
```diff
@@ -323,7 +323,11 @@ func (d *DeployAgent) Build(overrideBuildConfig *types.BuildConfig, forceBuild b
 }
 
 // Push pushes a local image to the remote repository linked in the release
-func (d *DeployAgent) Push() error {
+func (d *DeployAgent) Push(forcePush bool) error {
+	if !forcePush && d.tag != "latest" {
```
abelanger5, line 327: I'm confused by this, shouldn't we be checking if the tag exists in the image repo? In this case, it'll just skip all pushes that don't have a `latest` tag (assuming that forcePush isn't set). But we'd still like to push if the image doesn't exist in the repo.

abelanger5 (approved):

PR #1707, opened 2022-02-28
Default to Docker build when `--dockerfile` is set

mnafees:
## Pull request type

- [ ] Bugfix
- [ ] Feature

## Pull request checklist

Method can be `pack` even when `--dockerfile` is set.

## What is the new behavior?

Default to Docker build method when `--dockerfile` is set.


mnafees:
This is not a novel idea but perhaps might lead to `porter update` breaking for a few users because we are not handling the edge cases properly. Closing this.

PR #1733, opened 2022-03-08
[hotfix] Wait for `new-release-tests` workflows to exit in prerelease

mnafees:


mnafees:
We do not want to depend on third party actions. Closing this.

PR #1758, opened 2022-03-15
WIP: Pass build secrets to update action via Github secrets API

lsnow99:
Depends on https://github.com/porter-dev/porter-update-action/pull/8

## Pull request type

- [ ] Bugfix
- [x] Feature

## Pull request checklist

Secrets need to be manually configured by users in their Github actions. There is no current integration with the Github secrets API.

## What is the new behavior?

**Not fully implemented yet**

Users can mark environment variables as build arguments. If an environment variable is marked as secret and as a build argument, then it will be stored as a secret in Github for that repository and be accessible by Github actions as `PORTERSECRET_SECRETNAME=SECRETVALUE`. Whenever a build argument changes, a new build for that release is triggered. If the build argument happens to also be a secret, then this involves writing a new Github action yaml to the repository. Note that this will overwrite the old Github action yaml, so if a user has modified the porter action on their own previously, their changes will be overwritten.

What has been implemented:
- CLI is updated to convert all environment variables of the form PORTERSECRET_SECRETNAME=SECRETVALUE to build arguments
- porter-update-action is updated to accept a comma-separated list of key=value pairs in a `build_secrets` input and will create environment variables of the form `PORTERSECRET_SECRETNAME=SECRETVALUE` from them, which are used by the CLI for building
- When writing the initial GitHub action to the chosen repository, any secrets that were provided 
- The `GithubActions` struct has been modified to include a `BuildSecrets` map. There is a new method called `UpdateRepoBuildSecrets` that will bring the repository build secrets to the desired state if the `BuildSecrets` map is non-nil. `UpdateRepoBuildSecrets` is called by `GithubActions.Setup()`
- The `GetGithubActionYAML` now generates an update app step that includes passing the comma separated list of secret key value pairs to the porter-update-action via the `build_secrets` input. An example of this in a generated workflow is `build_secrets: SECRET2="${{ secrets.PORTERSECRET_SECRET2" }},SECRET1="${{ secrets.PORTERSECRET_SECRET1" }}`

Things left to do:
- Determine when to trigger rebuilds/to update build secrets.
  - Whenever a synced environment group is modified or if the release's normal environment is modified and build arguments are changed, then it should trigger a flow to update the secrets in GitHub by calling `GithubActions.Setup()`. If only non-secret build arguments were changed, then a build should still be triggered via the GitHub API, but no new workflow should be created.
- In the create handler (and at any point that build secrets are modified) we need to actually retrieve the secret values from Kubernetes so that we can store them in GitHub. Right now this is left as a TODO
- Track what environment variables are build arguments in the database and update the frontend to support this


dashboard/src/main/home/cluster-dashboard/env-groups/EnvGroupArray.tsx
```diff
@@ -113,6 +114,14 @@ const EnvGroupArray = ({
                     spellCheck={false}
                   />
 
+                  {!disabled && (
```
lsnow99, line 117: This file contains temporary code to demonstrate a possible UI for how to mark environment variables as build arguments.

api/server/handlers/release/create.go
```diff
@@ -30,6 +31,22 @@ type CreateReleaseHandler struct {
 	authz.KubernetesAgentGetter
 }
 
+type CreateReleaseEnvValues struct {
```
lsnow99, line 34: Not sure if this struct + the json marshal and unmarshalling process is the best way to read the environment variables

PR #1829, opened 2022-04-04
[POR-385] Create new preview environment endpoints and custom Github Action workflow

mnafees:
## Pull request type

- [ ] Bugfix
- [ ] Feature

## Pull request checklist


api/server/handlers/environment/reenable_deployment.go
```diff
@@ -0,0 +1,117 @@
...
+	cluster, _ := r.Context().Value(types.ClusterScope).(*models.Cluster)
+
+	deplID, reqErr := requestutils.GetURLParamUint(r, "deployment_id")
+
+	if reqErr != nil {
+		c.HandleAPIError(w, r, reqErr)
+		return
+	}
+
+	depl, err := c.Repo().Environment().ReadDeploymentByID(deplID)
```
abelanger5, line 46: Needs verification that this deployment belongs to the project and cluster. Otherwise any user can re-enable the deployment status (will only set it to Creating at the moment).

internal/repository/environment.go
```diff
@@ -6,11 +6,14 @@ type EnvironmentRepository interface {
 	CreateEnvironment(env *models.Environment) (*models.Environment, error)
 	ReadEnvironment(projectID, clusterID, gitInstallationID uint, gitRepoOwner, gitRepoName string) (*models.Environment, error)
 	ReadEnvironmentByID(projectID, clusterID, envID uint) (*models.Environment, error)
+	ReadEnvironmentByOwnerRepoName(owner, repo string) (*models.Environment, error)
 	ListEnvironments(projectID, clusterID uint) ([]*models.Environment, error)
 	DeleteEnvironment(env *models.Environment) (*models.Environment, error)
 	CreateDeployment(deployment *models.Deployment) (*models.Deployment, error)
 	ReadDeployment(environmentID uint, namespace string) (*models.Deployment, error)
```
abelanger5, line 13: To ensure that handlers can't accidentally read deployments they don't have access to, each repo call should be querying for the `project_id` and `cluster_id` as well. This guarantees handler safety. This applies to:
- `ReadDeployment` 
- `ReadEnvironmentByOwnerRepoName`
- `ReadDeploymentByGitDetails`
- `ListDeployments`

Since there are cases where these methods are called from the incoming webhook handler, we can also consider attaching a unique webhook secret to the environment so that the webhook can identify an environment instead.

abelanger5 (changes requested):

api/server/handlers/environment/delete_deployment.go
```diff
@@ -34,32 +34,39 @@ func NewDeleteDeploymentHandler(
...
-	if ok := c.DecodeAndValidate(w, r, request); !ok {
+	if reqErr != nil {
+		c.HandleAPIError(w, r, apierrors.NewErrInternal(reqErr))
 		return
 	}
 
-	// read the environment to get the environment id
-	env, err := c.Repo().Environment().ReadEnvironment(project.ID, cluster.ID, uint(ga.InstallationID), owner, name)
+	// read the deployment
+	depl, err := c.Repo().Environment().ReadDeploymentByGitDetails(envID, owner, name, prNumber)
```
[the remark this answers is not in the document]
abelanger5: Resolved by reading the environment further up in the method.

internal/repository/environment.go
abelanger5: Resolved by:
- Adding project/cluster IDs to calls that read from the environment and deployment
- Adding a unique webhook ID to the environment model which is queried when project/cluster info is not available

abelanger5 (approved):

PR #1910, opened 2022-04-19
[hotfix] Preview environment improvements

mnafees:
- Keep the GH API as the source of truth for deployment statuses
- Workflow run URL
- fetch all open PRs for a repository

### DIFF
--- a/api/server/handlers/environment/list_deployments_by_cluster.go
+++ b/api/server/handlers/environment/list_deployments_by_cluster.go
@@ -195,7 +195,7 @@ func fetchOpenPullRequests(
 
 	var ghPRs []*github.PullRequest
 
-	for resp.NextPage != 0 && err != nil {
+	for resp.NextPage != 0 && err == nil {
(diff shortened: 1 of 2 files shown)


api/server/handlers/environment/list_deployments_by_cluster.go
```diff
@@ -147,6 +193,21 @@ func fetchOpenPullRequests(
 		return nil, err
 	}
 
+	var ghPRs []*github.PullRequest
+
+	for resp.NextPage != 0 && err != nil {
```
abelanger5, line 198:
```suggestion
	for resp.NextPage != 0 && err == nil {
```

I think this should be `err == nil`?

abelanger5:
```suggestion
	for resp.NextPage != 0 && err == nil {
```

I think this should be `err == nil`?

abelanger5 (changes requested):

abelanger5 (approved):

PR #1915, opened 2022-04-19
Fix alternative GCR hostnames

abelanger5:
## Pull request type

- [X] Bugfix
- [ ] Feature

## Pull request checklist

Alternative GCR hostnames like `eu.gcr.io` don't properly list repositories. 

## What is the new behavior?

Add casing for alternative GCR hostnames.

### DIFF
--- a/internal/registry/registry.go
+++ b/internal/registry/registry.go
@@ -131,7 +131,7 @@ func (r *Registry) listGCRRepositories(
 
 	regURL := r.URL
 
-	if !strings.Contains(regURL, "http") {
+	if !strings.HasPrefix(regURL, "http") {
 		regURL = fmt.Sprintf("https://%s", regURL)
 	}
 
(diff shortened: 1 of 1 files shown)


internal/registry/registry.go
```diff
@@ -129,9 +129,22 @@ func (r *Registry) listGCRRepositories(
 	// for oauth. This also prevents us from making more requests.
 	client := &http.Client{}
 
+	regURL := r.URL
+
+	if !strings.Contains(regURL, "http") {
```
mnafees, line 134: I think we should use `strings.HasPrefix` here. What if the URL contains `http` but not as the URL scheme? Line 138 can then parse / throw an error because of parsing ambiguities (Re: https://github.com/porter-dev/porter/pull/1915/files)

mnafees (changes requested):

internal/registry/registry.go
abelanger5: Good point, just pushed 💯

mnafees (approved):
LGTM 👍

PR #1935, opened 2022-04-21
Fix namespaces not persisting across sidebar tabs

scottwey:
## Pull request type

- [x] Bugfix
- [ ] Feature

## Pull request checklist

Currently, selected `namespace`s do not persist when navigating across sidebar links. 

## What is the new behavior?
This should fix the issue of `namespace` not persisting across sidebar links. Sadly, I don't have a good way to test this, but please give it a look. Thanks!

This issue arises because `concat` does not mutate the string in place, but returns a new string (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/concat). We need to reassign to the variable `search` after concat-ing the `namespace` string.

### DIFF
--- a/dashboard/src/main/home/sidebar/Sidebar.tsx
+++ b/dashboard/src/main/home/sidebar/Sidebar.tsx
@@ -116,7 +116,7 @@ class Sidebar extends Component<PropsType, StateType> {
               let search = `?cluster=${currentCluster.name}&project_id=${currentProject.id}`;
 
               if (pathNamespace) {
-                search.concat(`&namespace=${pathNamespace}`);
+                search = search.concat(`&namespace=${pathNamespace}`);
               }
 
               return {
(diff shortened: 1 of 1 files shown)


abelanger5:
Hey @scottwey thanks for this PR 💯  -- putting this in our integration environment to run some tests.

jnfrati (approved):
Thanks for the PR @scottwey!

scottwey:
@jnfrati Odd that this fix doesn't work. Is there any easy way for me to test this myself?

jnfrati:
Hey @scottwey! Sorry for the lack of context in the review. 

Basically, the issue, rather than the namespace being concatenated to the query params or not, is that at some point during the navigation, the namespace is being cleared so it gets lost. The bad news would be that I didn't manage to find the point where this is happening.

[This guide](https://github.com/porter-dev/porter/blob/master/docs/developing/setup.md) should guide you to make a basic local installation of porter to develop! 
Besides that you should add to the `docker/.env` file an extra env variable `DISABLE_ALLOWLIST=true` to be sure that the frontend is not going to crash. 

Let me know if this works for you and I'll be glad to help you if anything goes wrong during the local setup!

scottwey:
@jnfrati Thanks for the guide! I'll take a look if I have some time. :)

jnfrati (approved):
Hey @scottwey! I managed to find the root of the issue, we were pushing the namespace to the query parameters instead of the url params so the nav links were never going to find anything. Thanks for your contribution! Merging this up :raised_hands:

PR #1981, opened 2022-05-05
[POR-449] Initial changes for public API v1

mnafees:
## Pull request type

- [ ] Bugfix
- [x] Feature

## Pull request checklist

N/A

### DIFF
--- a/api/server/handlers/cluster/delete_namespace.go
+++ b/api/server/handlers/cluster/delete_namespace.go
@@ -8,6 +8,7 @@ import (
 	"github.com/porter-dev/porter/api/server/shared"
 	"github.com/porter-dev/porter/api/server/shared/apierrors"
 	"github.com/porter-dev/porter/api/server/shared/config"
+	"github.com/porter-dev/porter/api/server/shared/requestutils"
 	"github.com/porter-dev/porter/api/types"
 	"github.com/porter-dev/porter/internal/models"
 )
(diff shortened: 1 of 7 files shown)


api/server/router/namespace_v1.go
```diff
@@ -0,0 +1,55 @@
+package router
```
abelanger5, line 1: I think having a separate `v1` package inside of the router makes sense.

api/server/handlers/cluster/delete_namespace.go
```diff
@@ -28,10 +29,16 @@ func NewDeleteNamespaceHandler(
 }
 
 func (c *DeleteNamespaceHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
-	request := &types.DeleteNamespaceRequest{}
+	namespace, _ := requestutils.GetURLParamString(r, types.URLParamNamespace)
 
-	if ok := c.DecodeAndValidate(w, r, request); !ok {
-		return
+	if namespace == "" {
```
abelanger5, line 34: I do think it's ok to change the signature for endpoints that are called very infrequently, like delete namespace -- we can just make sure the frontend is changed as well. But if we're going to add casing like this it's best to create a separate handler (I don't think we _strictly_ need to use the same handlers -- let's just make sure any shared logic is called by both handlers).

api/server/handlers/cluster/get_namespace.go
```diff
@@ -0,0 +1,61 @@
...
+	cluster, _ := r.Context().Value(types.ClusterScope).(*models.Cluster)
+
+	agent, err := c.GetAgent(r, cluster, "")
+
+	if err != nil {
+		c.HandleAPIError(w, r, apierrors.NewErrInternal(err))
+		return
+	}
+
+	res, err := agent.GetNamespace(namespace)
```
abelanger5, line 48: I think we might want our own API type for some Kubernetes objects like namespace

api/server/handlers/cluster/get_namespace.go
```diff
@@ -0,0 +1,61 @@
...
+
+	if reqErr != nil {
+		c.HandleAPIError(w, r, apierrors.NewErrInternal(reqErr))
+	}
+
+	cluster, _ := r.Context().Value(types.ClusterScope).(*models.Cluster)
+
+	agent, err := c.GetAgent(r, cluster, "")
+
+	if err != nil {
```
abelanger5, line 43: Shouldn't we be casing on `Not Found` here as well?

PR #2018, opened 2022-05-11
[hotfix] Do not upgrade or rollback image to hello-porter images

mnafees:


api/server/handlers/release/update_rollback.go
```diff
@@ -118,7 +118,9 @@ func updateReleaseRepo(config *config.Config, release *models.Release, helmRelea
 		return fmt.Errorf("Could not find field repository in config")
 	}
 
-	if repoStr != release.ImageRepoURI {
+	if repoStr != "public.ecr.aws/o1j4x7p4/hello-porter" &&
```
abelanger5, line 121: Should throw an error to the user here

api/server/handlers/release/upgrade.go
```diff
@@ -117,6 +118,60 @@ func (c *UpgradeReleaseHandler) ServeHTTP(w http.ResponseWriter, r *http.Request
...
+						repoStr = rel.GitActionConfig.ImageRepoURI
+					}
+				} else {
+					repoStr = rel.ImageRepoURI
+				}
+
+				if repoStr == "public.ecr.aws/o1j4x7p4/hello-porter" ||
+					repoStr == "public.ecr.aws/o1j4x7p4/hello-porter-job" {
+					c.HandleAPIError(w, r, apierrors.NewErrPassThroughToClient(
+						fmt.Errorf("malformed image repository"),
```
abelanger5, line 160: Can't do this unfortunately, as we need to be able to update things like build-time env vars

abelanger5 (commented):
Going to dismiss this in favor of a different PR - as I mentioned I think the updates to the `upgrade.go` file are likely to break a few flows.

PR #2100, opened 2022-06-07
[POR-496] Add ability to locally set path to a kubeconfig

mnafees:
## Pull request type

- [ ] Bugfix
- [x] Feature

## Pull request checklist

Commands like `porter run` need to get a temporary kubeconfig from the Porter server to be able to run properly.

## What is the new behavior?

The CLI now has the option to set path to a local kubeconfig that can be used for such commands. This will be necessary for in-cluster auth mechanisms and self-hosted users.

### DIFF
--- a/cli/cmd/config.go
+++ b/cli/cmd/config.go
@@ -155,7 +155,7 @@ var configSetHostCmd = &cobra.Command{
 }
 
 var configSetKubeconfigCmd = &cobra.Command{
-	Use:   "set-kubeconfig [host]",
+	Use:   "set-kubeconfig [kubeconfig-path]",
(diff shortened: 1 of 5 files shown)


cli/cmd/config.go
```diff
@@ -154,6 +154,20 @@ var configSetHostCmd = &cobra.Command{
 	},
 }
 
+var configSetKubeconfigCmd = &cobra.Command{
+	Use:   "set-kubeconfig [host]",
```
abelanger5, line 158:
```suggestion
	Use:   "set-kubeconfig [kubeconfig-path]",
```

abelanger5 (changes requested):

PR #2108, opened 2022-06-08
[POR-508] [FEAT] Gitlab integration

jnfrati:
## Pull request type

- [ ] Bugfix
- [x] Feature

## Pull request checklist

## Motivation

Create a v1 integration where users can connect their self hosted instances of gitlab, and can deploy applications with the same features that porter already has for github.


api/server/handlers/project_integration/get_gitlab_repo_contents.go
```diff
@@ -0,0 +1,171 @@
...
+
+	request := &types.GetContentsRequest{}
+
+	ok := p.DecodeAndValidate(w, r, request)
+
+	if !ok {
+		return
+	}
+
+	integrationID, reqErr := requestutils.GetURLParamUint(r, "integration_id")
```
abelanger5, line 49: Since we're reading this in many handlers, we should pull this into the `authz` package and pass the gitlab integration through the context, as we're doing with other resources.

api/server/handlers/project_integration/get_gitlab_repo_buildpack.go
```diff
@@ -0,0 +1,219 @@
...
+		p.Config(), gi,
+	), oauth.MakeUpdateGitlabAppOAuthIntegrationFunction(project.ID, giAppOAuth, p.Repo()))
+
+	if err != nil {
+		p.HandleAPIError(w, r, apierrors.NewErrPassThroughToClient(fmt.Errorf("invalid gitlab access token"),
+			http.StatusUnauthorized))
+		return
+	}
+
+	client, err := gitlab.NewOAuthClient(accessToken, gitlab.WithBaseURL(gi.InstanceURL))
```
abelanger5, line 125: It seems like we're reusing this block a lot (getting the access token and creating a new client), we should pull this into a shared helper or middleware.

abelanger5 (changes requested):
Looking great, just a few minor cleanup items 💯

PR #2155, opened 2022-06-16
[POR-515] Allow user to change the default branch for an application

jnfrati:
## Pull request type

- [ ] Bugfix
- [ ] Feature

## Pull request checklist


api/server/handlers/release/update_git_action_config.go
```diff
@@ -0,0 +1,71 @@
...
+	name, _ := requestutils.GetURLParamString(r, types.URLParamReleaseName)
+	namespace := r.Context().Value(types.NamespaceScope).(string)
+
+	request := &types.UpdateGitActionConfigRequest{}
+
+	if ok := c.DecodeAndValidate(w, r, request); !ok {
+		return
+	}
+
+	release, err := c.Repo().Release().ReadRelease(cluster.ID, name, namespace)
```
abelanger5, line 41: This should be using the `ReleaseScope` so that we're able to check whether a user has permission to update a release.

abelanger5: Note that the release scope requires a revision passed in the url, but you can just set this to `0` to get the latest revision.

api/types/release.go
```diff
@@ -193,3 +193,7 @@ type GetReleaseAllPodsResponse []v1.Pod
 type PatchUpdateReleaseTags struct {
 	Tags []string `json:"tags"`
 }
+
+type UpdateGitActionConfigRequest struct {
+	GitActionConfig *GitActionConfig `json:"git_action_config"`
```
abelanger5, line 198: If we're only supporting updating the branch, we shouldn't provide the whole object as a parameter in the request. Just thinking about if we wanted to make this an `/api/v1` endpoint.

internal/repository/gorm/git_action_config.go
```diff
@@ -49,3 +49,7 @@ func (repo *GitActionConfigRepository) ReadGitActionConfig(id uint) (*models.Git
 
 	return ga, nil
 }
+
+func (repo *GitActionConfigRepository) UpdateGitActionConfig(ga *models.GitActionConfig) error {
```
abelanger5, line 53: Make sure you add this method in the `test` repository as well, otherwise builds will fail.

abelanger5 (changes requested):

PR #2240, opened 2022-07-06
[POR-638] Helm release revisions monitoring using a worker pool and job queue 

mnafees:
## Pull request type

- [ ] Bugfix
- [x] Feature

## Pull request checklist

Helm release revisions can grow huge in size and lead to slowdown of the dashboard, backend and infra in general.

## What is the new behavior?

The case at hand is that Helm release revisions can get in huge numbers and this leads to the slowdown of the dashboard, backend and our infra. We should have a cronjob like worker to go through clusters and their release revisions and keep onlt the 100 most recent ones. We should be backing up the deleted revisions to an S3 bucket.


internal/models/cluster.go
```diff
@@ -75,6 +75,9 @@ type Cluster struct {
 
 	// CertificateAuthorityData for the cluster, encrypted at rest
 	CertificateAuthorityData []byte `json:"certificate-authority-data,omitempty"`
+
+	// MonitorHelmReleases to trim down the number of revisions per release
+	MonitorHelmReleases bool `json:"monitor_helm_releases"`
```
abelanger5, line 80:
```suggestion
	MonitorHelmReleases bool
```

JSON tag isn't necessary since this never gets used as an API type (other JSON tags are outdated)

abelanger5 (changes requested):
This is awesome - just some minor comments on the code 💯 . 

One doc item - some diagrams detailing how the channels are constructed, where the channels should be closed from, and the interplay between the dispatcher and worker would be great. A separate markdown file inside the `./docs` folder seems necessary. 

Several items to get this deployed:
- For each Github action that pushes the provisioner image, we need to create a corresponding job that pushes the worker image.
- Make a PR into the `porter-self-hosted` repo which includes the worker and any crons that need to run if the worker is deployed.

abelanger5 (approved):

PR #2300, opened 2022-07-28
Implement stack and stack source config name update

jnfrati:
## Pull request type

- [ ] Bugfix
- [ ] Feature

## Pull request checklist

### DIFF
--- a/api/server/handlers/stack/create.go
+++ b/api/server/handlers/stack/create.go
@@ -269,10 +269,10 @@ func getSourceConfigModels(sourceConfigs []*types.CreateStackSourceConfigRequest
 			}
 
 			// If the source config had a source config ID then we need to copy it over
-			if len(sourceConfig.SourceConfigID) != 0 {
-				newSourceConfig.SourceConfigID = sourceConfig.SourceConfigID
+			if sourceConfig.StableSourceConfigID != "" {
+				newSourceConfig.StableSourceConfigID = sourceConfig.StableSourceConfigID
 			} else {
-				newSourceConfig.SourceConfigID = string(uid)
+				newSourceConfig.StableSourceConfigID = string(uid)
 			}
 
 			res = append(res, *newSourceConfig)
--- a/api/server/handlers/stack/update_stack_name.go
+++ b/api/server/handlers/stack/update_stack_name.go
@@ -1,0 +1,64 @@
+package stack
+
+import (
+	"fmt"
+	"net/http"
+
+	"github.com/porter-dev/porter/api/server/handlers"
+	"github.com/porter-dev/porter/api/server/shared"
+	"github.com/porter-dev/porter/api/server/shared/apierrors"
+	"github.com/porter-dev/porter/api/server/shared/config"
+	"github.com/porter-dev/porter/api/types"
+	"github.com/porter-dev/porter/internal/models"
+)
+
+type StackUpdateStackName struct {
+	handlers.PorterHandlerReadWriter
+}
+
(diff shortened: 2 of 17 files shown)


api/server/handlers/stack/create.go
```diff
@@ -261,12 +261,21 @@ func getSourceConfigModels(sourceConfigs []*types.CreateStackSourceConfigRequest
...
+			newSourceConfig := &models.StackSourceConfig{
 				UID:          uid,
 				Name:         sourceConfig.Name,
 				ImageRepoURI: sourceConfig.ImageRepoURI,
 				ImageTag:     sourceConfig.ImageTag,
-			})
+			}
+
+			// If the source config had a source config ID then we need to copy it over
+			if len(sourceConfig.SourceConfigID) != 0 {
```
abelanger5, line 272: I think it's more idiomatic to check for the empty string, also reads a bit better (makes the type more clear)

internal/models/stack.go
```diff
@@ -160,6 +160,8 @@ func (s StackResource) ToStackResource(stackID string, stackRevisionID uint, sou
 type StackSourceConfig struct {
 	gorm.Model
 
+	SourceConfigID string
```
abelanger5, line 163: `SourceConfigID` is a confusing name for this field (we already have an `ID` and a `UID`). Can we call this `StableSourceConfigID` or something along those lines?

abelanger5 (changes requested):
Looks pretty good - think there are a few things to clarify to make it more clear what the additional field represents

api/server/handlers/stack/create.go
jnfrati: This would be something like `sourceConfig.SourceConfigID != ""`?

internal/models/stack.go
jnfrati: Sounds good!

api/server/handlers/stack/create.go
abelanger5: Yep

api/server/handlers/stack/update_stack_name.go
```diff
@@ -0,0 +1,64 @@
+package stack
```
mnafees, line 1: I think a generalised handler to update the stack with only a few options available like the name of the stack makes sense. This way it would be more future-proof and extensible.

api/server/router/v1/stack.go
```diff
@@ -820,5 +820,63 @@ func getV1StackRoutes(
 		Router:   r,
 	})
 
+	// PATCH /api/v1/projects/{project_id}/clusters/{cluster_id}/namespaces/{namespace}/stacks/{stack_id}/update_name -> stack.NewStackAddApplicationHandler
```
mnafees, line 823: Going by the above comment, this could then possibly be 
```
PATCH /api/v1/projects/{project_id}/clusters/{cluster_id}/namespaces/{namespace}/stacks/{stack_id}
```

api/types/stacks.go
```diff
@@ -63,6 +63,10 @@ type CreateStackAppResourceRequest struct {
 	SourceConfigName string `json:"source_config_name" form:"required"`
 }
 
+type UpdateStackNameRequest struct {
```
mnafees, line 66: This can simply be `UpdateStackRequest`.

mnafees (changes requested):
Added a few comments about the update stack handler. But all else looks good to me 👍🏽

abelanger5 (approved):

PR #2305, opened 2022-08-01
[hotfix] Enable helm repo monitoring for all newly created clusters

mnafees:
## Pull request type

- [ ] Bugfix
- [x] Feature

## Pull request checklist

N/A

## What is the new behavior?

Enable helm repo monitoring for all newly created clusters.

### DIFF
--- a/api/server/handlers/cluster/create.go
+++ b/api/server/handlers/cluster/create.go
@@ -112,7 +112,6 @@ func getClusterModelFromManualRequest(
 		GCPIntegrationID:         request.GCPIntegrationID,
 		AWSIntegrationID:         request.AWSIntegrationID,
 		CertificateAuthorityData: cert,
-		MonitorHelmReleases:      true,
 	}, nil
 }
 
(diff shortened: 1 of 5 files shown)


api/server/handlers/cluster/create.go
```diff
@@ -112,6 +112,7 @@ func getClusterModelFromManualRequest(
 		GCPIntegrationID:         request.GCPIntegrationID,
 		AWSIntegrationID:         request.AWSIntegrationID,
 		CertificateAuthorityData: cert,
+		MonitorHelmReleases:      true,
```
abelanger5, line 115: Actually this probably shouldn't be set to `true` - when users bring their own cluster, rather than provisioning through us, this behavior would be extremely unexpected (especially if they already have a number of helm releases).

internal/kubernetes/resolver/resolver.go
```diff
@@ -390,6 +390,7 @@ func (rcf *CandidateResolver) buildCluster() (*models.Cluster, error) {
 		InsecureSkipTLSVerify:   kcCluster.InsecureSkipTLSVerify,
 		UserLocationOfOrigin:    kcAuthInfo.LocationOfOrigin,
 		UserImpersonate:         kcAuthInfo.Impersonate,
+		MonitorHelmReleases:     true,
```
abelanger5, line 393: Same deal here - I think this should still be `false`.

abelanger5 (changes requested):

abelanger5 (approved):

PR #2329, opened 2022-08-18
Advanced RBAC

mnafees:


api/server/handlers/project/list_test.go
```diff
@@ -1,74 +0,0 @@
...
-
-	"github.com/porter-dev/porter/api/server/handlers/project"
-	"github.com/porter-dev/porter/api/server/shared"
-	"github.com/porter-dev/porter/api/server/shared/apitest"
-	"github.com/porter-dev/porter/api/types"
-	"github.com/porter-dev/porter/internal/models"
-	"github.com/porter-dev/porter/internal/repository/test"
-)
-
-func TestListProjectsSuccessful(t *testing.T) {
```
abelanger5, line 14: Why is this deleted?

abelanger5 (changes requested):
Looks pretty good! Left some minor comments. 

We should definitely document how the policy document gets parsed somewhere - the `policy` package should have a `doc.go` file, perhaps we can add it there for now.

PR #2335, opened 2022-08-24
rfac: move useLogs hook to separate file

meehawk:
## Pull request type

- [ ] Bugfix
- [ ] Feature
- [x] Other (please describe):

This PR refactors the `dashboard/src/main/home/cluster-dashboard/expanded-chart/status/Logs.tsx` file with the following changes:

- moves `useLogs` hook to a separate file
- moves common types needed by both `useLogs` and `Logs.tsx` into a common `types.ts` folder present in the same parent directory 

## Pull request checklist

~~## What is the current behavior?~~

~~## What is the new behavior?~~


jnfrati (approved):
Hey @meehawk! Thanks for the PR! 
Looks perfect, but I have a quick question (not blocking pure curiosity) about moving the SelectedPodType to another file, this is an entirely personal preference, but when there are a small number of custom types I usually prefer to have these on quick access at the same file they're being used. 
Do you have any strong opinions against this behavior? (We do this a lot over the whole project)

meehawk:
> Hey @meehawk! Thanks for the PR! Looks perfect, but I have a quick question (not blocking pure curiosity) about moving the SelectedPodType to another file, this is an entirely personal preference, but when there are a small number of custom types I usually prefer to have these on quick access at the same file they're being used. Do you have any strong opinions against this behavior? (We do this a lot over the whole project)

Great question @jnfrati, I agree with the preferance to keep the types where they are used and thats what I'd have done as well. My reasoning for separate file was because it was a common file that was used both by `useLogs` hook as well as the `Logs.tsx` component and SelectedPodType semantically is unrelated to either of the two files use-case. We could also move the type to `useLogs` and export from there if thats preferred

jnfrati:
@meehawk That sounds like a perfect reason! 
I was just wondering what your thoughts were around this but the implementation looks perfect! Going to merge this up after the workflow run 👍

PR #2336, opened 2022-08-24
fix: status logs freeze on high log inflow

meehawk:
## Pull request type

- [ ] Bugfix
- [ ] Feature

## Pull request checklist

Currently, for pods that generate a high volume of logs, the logs section in the status page freezes. One possible cause of this could be a large number of re-renders

Steps to reproduce:

1. Launch a Web Template
2. For the docker image, select: `sohamparekh1998/heavy-logger`
3. Status page freezes due to a large number of logs after some time and becomes unusable

## What is the new behavior?

The updates to the logs frame are chunked using a buffer to prevent multiple re-renders in a short span of time

1. Add a `logsBuffer` to act as a buffer for logs
2. Introduced methods to `flush` buffer and update `logs` for the `currentContainer` in the status tabs
3. Incremented `MAX_LOGS_SIZE` to 5000


dashboard/src/main/home/cluster-dashboard/expanded-chart/status/Logs.tsx
```diff
@@ -257,6 +258,7 @@ const useLogs = (
   currentPod: SelectedPodType,
   scroll?: (smooth: boolean) => void
 ) => {
+  let logsBuffer: Anser.AnserJsonEntry[][] = [];
```
jnfrati, line 261: The implementation of the logsBuffer looks nice! Although afaik this might get cleared out (resulting on lost logs) by a rerender on the parent. Maybe a useRef could solve this issue?

dashboard/src/main/home/cluster-dashboard/expanded-chart/status/Logs.tsx
```diff
@@ -340,6 +342,51 @@ const useLogs = (
...
+    setLogs((logs) => {
+      const tmpLogs = { ...logs };
+      let containerLogs = tmpLogs[containerName] || [];
+
+      containerLogs.push(...newLogs);
+      // this is technically not as efficient as things could be
+      // if there are performance issues, a deque can be used in place of a list
+      // for storing logs
+      if (containerLogs.length > MAX_LOGS) {
+        containerLogs.shift();
```
jnfrati, line 363: Not fully sure if this could cause any issue, but maybe here would be nice to remove the first `N=newLogs.length` items of the queue instead of just one to avoid overflowing the containerLogs too much.

jnfrati (changes requested):
Hey! This looks awesome! Left some comments to see what you think.

dashboard/src/main/home/cluster-dashboard/expanded-chart/status/Logs.tsx
meehawk: @jnfrati awesome catch, just implemented this.

dashboard/src/main/home/cluster-dashboard/expanded-chart/status/Logs.tsx
meehawk: Nice, update as per suggestion in the latest commit

dashboard/src/main/home/cluster-dashboard/expanded-chart/status/useLogs.ts
```diff
@@ -94,6 +96,51 @@ export const useLogs = (
...
+    setLogs((logs) => {
+      const tmpLogs = { ...logs };
+      let containerLogs = tmpLogs[containerName] || [];
+
+      containerLogs.push(...newLogs);
+      // this is technically not as efficient as things could be
+      // if there are performance issues, a deque can be used in place of a list
+      // for storing logs
+      if (containerLogs.length > MAX_LOGS) {
+        containerLogs.slice(MAX_LOGS);
```
jnfrati, line 117: Hey @meehawk two small items here! Let me know what you think!
 
I think a better approach would be to do:
```suggestion
        const logsToBeRemoved = newLogs.length < LOGS_BUFFER_SIZE ? newLogs.length : LOGS_BUFFER_SIZE;
        containerLogs = containerLogs.slice(logsToBeRemoved);
```

This way we would only be deleting the necessary amount of logs required for the new logs to not pass the  MAX_LOGS size. Although this would leave us on the edge each time, so open to hearing other ideas here! 

Also added on the second line the re-assignation of the containerLogs variable as if we don't do this, the containerLogs will not be updated with the reduced array, [reference docs here](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/slice)

meehawk: i think the reason we dont reassign the container array is so that we can still keep the "rendered output" corresponding to the previous logs in the DOM. If we change the reference, we also lose older logs from the UI

meehawk: Hey @jnfrati thanks for the suggestion.

1. Agree with the suggestion. Done!
2. Nice catch! Updated the code

dashboard/src/main/home/cluster-dashboard/expanded-chart/status/useLogs.ts
```diff
@@ -199,6 +251,26 @@ export const useLogs = (
...
+  }, []);
+
+  /**
+   * In some situations, we might never hit the limit for the max buffer size.
+   * An example is if the total logs for the pod < LOGS_BUFFER_SIZE.
+   *
+   * For handling situations like this, we would want to force a flush operation
+   * on the buffer so that we dont have any stale logs
+   */
+  useEffect(() => {
```
jnfrati, line 265: While I was testing I noticed that the system logs are not being populated properly. After digging a little bit around the code I noticed that the issue comes from this useEffect, as the logs buffer doesn't differentiate between containerNames, when a user switches tabs to the system logs the flushLogsBuffer ends up overriding the system logs. This might also happen on pods that have more than two containers so it would be nice to consider this case in wich a pod might have N+2 containers.

I think modifying the logsBuffer to be a map of arrays in which each key represents a containerName would be better. What do you think?

dashboard/src/main/home/cluster-dashboard/expanded-chart/status/Logs.tsx
```diff
@@ -108,15 +109,16 @@ const LogsFC: React.FC<{
 
     return logs?.map((log, i) => {
       return (
-        <Log key={i}>
-          {log.map((ansi, j) => {
+        <Log key={[log.lineNumber, i].join(".")}>
+          <span className="line-number">{log.lineNumber}</span>
```
jnfrati, line 113: This log line number works pretty good! But we noticed that it breaks ansi-based parsing, could you take a look at it? 

You can replicate this using the supertest2014/nyan container image 🙌 

Before log line number:
Screen Shot 2022-08-30 at 11 25 01

After log line number:
image

jnfrati (changes requested):
Hey @meehawk!

We ended up testing a little bit more today and found a couple of issues that I described here. Let me know if you need any more clarification around these issues or if you need any help!

dashboard/src/main/home/cluster-dashboard/expanded-chart/status/Logs.tsx
meehawk: @jnfrati resolved the issues.Thanks for the catch!

dashboard/src/main/home/cluster-dashboard/expanded-chart/status/useLogs.ts
meehawk: Done 🚀

jnfrati:
Looks awesome! 
Thanks for all the work around this issue!

PR #2340, opened 2022-08-26
fix: failed pods deletion + include metadata in tooltip for failed pods

meehawk:
## Pull request type

- [x] Bugfix
- [ ] Feature

## Pull request checklist

Currently, there are two issues with handling failed pods:

1. We indicate that a particular pod is failing but that alone doesnt help to make it clear why a particular pod is failing
2. Deleting a failed pod leads to an error due to api calls being made with `undefined` and `undefined` for pod name and namespace

## What is the new behavior?

This PR:
1. Adds additional metadata to the tooltip for a failing pod to help user get more insight into why a aprticular pod might be failing
Screenshot 2022-08-26 at 4 19 13 PM

2. Fixes bug for deleting failed pods


jnfrati:
Looks perfect!

PR #2348, opened 2022-08-31
fix(docker): fix docker dev environments

dustinblackman:
## Pull request type

- [x] Bugfix
- [ ] Feature

## Pull request checklist

- [x] Docs have been reviewed and added / updated if needed

## What is the current behaviour?

The docker dev environments do not build or start.

## What is the new behaviour?

The docker dev  environments are now usable again.

- Fix `npm install` command in dashboard dev image to work with resolving issues currently bundled in `package-lock.json` with Node 16+.
- Drop running missing `migrate` command before air as it's done automatically when the porter API starts.
- Fix `air` not being installed when building docker image for API.

### DIFF
--- a/dashboard/docker/dev.Dockerfile
+++ b/dashboard/docker/dev.Dockerfile
@@ -7,7 +7,7 @@ COPY package*.json ./
 
 ENV NODE_ENV=development
 
-RUN npm install
+RUN npm ci --legacy-peer-deps
 RUN npm i -g http-parser-js
 
 COPY . ./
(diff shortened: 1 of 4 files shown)


abelanger5 (approved):
Thanks for the PR @dustinblackman! 

One thing to note is that the main Porter process only executes gorm's `AutoMigrate` method but the`/migrate` binary will soon be handling more complex schema migrations. That said we'll probably implement this as a docker-compose init container rather than including it in the docker start command, so this works well 👍

dustinblackman:
>One thing to note is that the main Porter process only executes gorm's AutoMigrate method but the/migrate binary will soon be handling more complex schema migrations. 

Ah! Good to note. Thank you!

Issue #2346, opened 2022-08-30
Welcome API fails when running in docker local


dustinblackman:
# Description

The welcome view that is shown after logging in to Porter for the first time when running locally fails to successfully create a project. The `/api/welcome` endpoint fails with a 500 error, with logs stating the following:

```
{"level":"warn","internal_error":"Get \"?company=https%3A%2F%2Fporter.run&email=test%40test.com&isCompany=true&name=Bob&role=founder\": unsupported protocol scheme \"\"","external_error":"An internal error occurred.","user_id":1,"method":"POST","url":"/api/welcome","time":"2022-08-30T14:57:07Z"}
{"level":"info","latency":0.137051,"status":500,"user_id":1,"method":"POST","url":"/api/welcome","time":"2022-08-30T14:57:07Z"}
```

# Location

- [X] Browser
- [ ] CLI
- [X] API

# Steps to reproduce

1. Run Porter locally with Docker using the following commands

```sh
docker volume create porter_sqlite                                                                                                                                                          ✘ 130 tags/v0.35.0 ✱
docker run \
  --mount type=volume,source=porter_sqlite,target=/sqlite,readonly=false \
  -e REDIS_ENABLED=false \
  -e SQL_LITE_PATH=/sqlite/porter.db \
  -p 8080:8080 \
  porter1/porter:v0.36.0
```

2. Open `https://127.0.0.1:8080`, click signup, and create an account.
3. When the welcome page loads, fill in the form and hit submit.
4. Verify in network tools and in logs that the request has failed.

# Additional Details

## Screenshots

<img width="1784" alt="image" src="https://user-images.githubusercontent.com/5246169/187471591-303303cc-7409-4c12-a1ec-87b15bca8e38.png">

## Logs

```
{"level":"info","time":"2022-08-30T15:02:40Z","message":"running migrations"}
{"level":"info","time":"2022-08-30T15:02:41Z","message":"Starting server :8080"}
{"level":"warn","internal_error":"stored cookie was not authenticated","external_error":"Forbidden","method":"GET","url":"/api/users/current","time":"2022-08-30T15:02:50Z"}
{"level":"info","latency":0.096336,"status":200,"method":"GET","url":"/api/metadata","time":"2022-08-30T15:02:50Z"}
{"level":"info","latency":0.025529,"status":200,"method":"GET","url":"/api/metadata","time":"2022-08-30T15:02:50Z"}
[...]

PR #2354, opened 2022-09-02
fix(api): handle when welcome hook is not configured

dustinblackman:
## Pull request type
- [x] Bugfix
- [ ] Feature

## Pull request checklist

Issue Number: Fixes https://github.com/porter-dev/porter/issues/2346

If `WELCOME_FORM_WEBHOOK` is not configured, first time signup will fail.

## What is the new behavior?

The welcome hook endpoint verifies if `WELCOME_FORM_WEBHOOK` is set before attempting to call it, making the parameter optional.


abelanger5 (approved):
Thanks for the PR @dustinblackman! Just checked this out locally and it works well, great work on the test cases too 👍. 

This will be added to `v0.37.0`, which should be built by the end of next week.

PR #2418, opened 2022-09-22
[hotfix] Do not make context copies of DB sessions for cluster operations

mnafees:
## Pull request type

- [x] Bugfix
- [ ] Feature

## Pull request checklist

For DB operations related to the cluster table (and its associates), we currently create a new DB session and add to the connection pool.

## What is the new behavior?

We no longer create new DB sessions in cluster operations.

### DIFF
--- a/workers/jobs/helm_revisions_count_tracker.go
+++ b/workers/jobs/helm_revisions_count_tracker.go
@@ -175,6 +175,7 @@ func (t *helmRevisionsCountTracker) Run() error {
 					Repo:                      t.repo,
 					DigitalOceanOAuth:         t.doConf,
 					AllowInClusterConnections: false,
+					Timeout:                   5 * time.Second,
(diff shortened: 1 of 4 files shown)


abelanger5 (dismissed):

workers/jobs/helm_revisions_count_tracker.go
```diff
@@ -46,7 +46,7 @@ import (
 	"helm.sh/helm/v3/pkg/releaseutil"
 )
 
-var stepSize int = 100
+var stepSize int = 20
```
abelanger5, line 49: Along with changing the step size, we need to reduce the kubernetes client connection timeout from 30 seconds to 5 seconds, otherwise this will only process 20 clusters every 30 seconds in the worst case.

abelanger5 (approved):

PR #2544, opened 2022-11-03
[POR-779] Enable custom namespaces and namespace annotations for preview environments

mnafees:
## Pull request type

- [ ] Bugfix
- [x] Feature

## Pull request checklist

We currently depend on our own generated namespace for a preview environment. There is no option for namespace annotations and preview envs are project-scoped.

## What is the new behavior?

This PR 
- adds support for custom namespaces for preview deployments
- adds support for namespace annotations
- adds a setting in cluster settings to enable preview env for a cluster
  - the parent project should have preview envs enabled for this to happen

### DIFF
--- a/api/server/handlers/cluster/update.go
+++ b/api/server/handlers/cluster/update.go
@@ -65,6 +65,10 @@ func (c *ClusterUpdateHandler) ServeHTTP(w http.ResponseWriter, r *http.Request)
 		cluster.AgentIntegrationEnabled = *request.AgentIntegrationEnabled
 	}
 
+	if request.PreviewEnvsEnabled != nil {
+		cluster.PreviewEnvsEnabled = *request.PreviewEnvsEnabled
+	}
+
 	if request.Name != "" && cluster.Name != request.Name {
 		cluster.Name = request.Name
 	}
--- a/api/server/handlers/environment/create.go
+++ b/api/server/handlers/environment/create.go
@@ -73,6 +73,17 @@ func (c *CreateEnvironmentHandler) ServeHTTP(w http.ResponseWriter, r *http.Requ
 		Mode:                request.Mode,
 		WebhookID:           string(webhookUID),
 		NewCommentsDisabled: false,
+		CustomNamespace:     request.CustomNamespace,
+	}
+
+	if len(request.NamespaceAnnotations) > 0 {
+		var annotations []string
+
+		for k, v := range request.NamespaceAnnotations {
+			annotations = append(annotations, fmt.Sprintf("%s=%s", k, v))
+		}
+
+		env.NamespaceAnnotations = []byte(strings.Join(annotations, ","))
 	}
 
 	// write Github actions files to the repo
@@ -179,6 +190,7 @@ func (c *CreateEnvironmentHandler) ServeHTTP(w http.ResponseWriter, r *http.Requ
 		GitInstallationID: uint(ga.InstallationID),
 		EnvironmentName:   request.Name,
 		InstanceName:      c.Config().ServerConf.InstanceName,
+		CustomNamespace:   env.CustomNamespace,
 	})
 
 	if err != nil {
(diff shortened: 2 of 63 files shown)


api/server/handlers/cluster/update.go
```diff
@@ -65,6 +65,10 @@ func (c *ClusterUpdateHandler) ServeHTTP(w http.ResponseWriter, r *http.Request)
 		cluster.AgentIntegrationEnabled = *request.AgentIntegrationEnabled
 	}
 
+	if request.PreviewEnvsEnabled != nil {
```
abelanger5, line 68: I thought we added a check to make sure the project had preview envs enabled here? Is that cased elsewhere?

api/server/handlers/environment/create.go
```diff
@@ -43,6 +41,14 @@ func (c *CreateEnvironmentHandler) ServeHTTP(w http.ResponseWriter, r *http.Requ
 	project, _ := r.Context().Value(types.ProjectScope).(*models.Project)
 	cluster, _ := r.Context().Value(types.ClusterScope).(*models.Cluster)
 
+	if !project.PreviewEnvsEnabled {
```
abelanger5, line 44: I think we should move this to shared auth middleware that gets registered for preview-specific endpoints.

cmd/migrate/main.go
```diff
@@ -111,3 +120,22 @@ func shouldPopulateSourceConfigDisplayName() bool {
 
 	return c.PopulateSourceConfigDisplayName
 }
+
+type EnableClusterPreviewEnvsConf struct {
+	// we add a dummy field to avoid empty struct issue with envdecode
+	DummyField string `env:"ASDF,default=asdf"`
+
+	// if true, will mark all clusters to have preview envs enabled whose parent project has it enabled
+	EnableClusterPreviewEnvs bool `env:"ENABLE_CLUSTER_PREVIEW_ENVS"`
```
abelanger5, line 129: I think we should cherry-pick the migration mechanism from Advanced RBAC and require this migration to be run. For all self-hosted instances, it'll be difficult to toggle on this environment variable.

abelanger5 (changes requested):
Looks great @mnafees 💯 

Just a few smaller comments/suggestions to address.

cmd/migrate/main.go
```diff
@@ -48,6 +52,39 @@ func main() {
...
+			return
+		}
+	}
+
+	latestMigrationVersion := startup_migrations.LatestMigrationVersion
+
+	if dbMigration.Version < latestMigrationVersion {
+		for ver, fn := range startup_migrations.StartupMigrations {
+			if ver > dbMigration.Version {
+				err := fn(db, logger)
```
abelanger5, line 71: I think we'll want to add a `lock` field to the `DbMigration` model to prevent concurrent migrations from running. This wasn't as crucial when we were just doing `AutoMigrate`, but is important if we're running one-off migration scripts.

abelanger5 (changes requested):

cmd/migrate/main.go
mnafees: Aha, nice catch! Will fix this.

cmd/migrate/main.go
```diff
@@ -52,12 +52,33 @@ func main() {
...
+	switch db.Dialector.Name() {
+	case "sqlite":
+		if err := tx.Exec("PRAGMA schema.locking_mode = EXCLUSIVE").Error; err != nil {
+			tx.Rollback()
+
+			logger.Fatal().Err(err).Msg("error acquiring lock on db_migrations")
+			return
+		}
+	case "postgres":
+		if err := db.Raw("LOCK TABLE db_migrations IN SHARE ROW EXCLUSIVE MODE").Error; err != nil {
```
abelanger5, line 66: Should this be `tx.Raw` or is that not supported via `gorm`?

abelanger5 (approved):

PR #2549, opened 2022-11-09
[POR-290] Add --wait flag to `porter update` to wait for successful deployment

mnafees:
## Pull request type

- [ ] Bugfix
- [x] Feature

## Pull request checklist

N/A

## What is the new behavior?

`porter update` now has a new `--wait` flag which when enabled will wait for the new version of the app to be ready in the cluster, or timeout in 30 minutes

### DIFF
--- a/cli/cmd/deploy.go
+++ b/cli/cmd/deploy.go
@@ -1184,7 +1184,7 @@ func checkDeploymentStatus(client *api.Client) error {
 			return fmt.Errorf("error fetching new replicaset: %w", err)
 		}
 
-		if minUnavailable <= rs.Status.ReadyReplicas {
+		if minAvailable <= rs.Status.ReadyReplicas {
 			success = true
 		}
 
--- a/cli/cmd/deploy.go
+++ b/cli/cmd/deploy.go
@@ -1075,8 +1075,8 @@ func checkDeploymentStatus(client *api.Client) error {
 		return fmt.Errorf("could not retrieve kubernetes credentials: %w", err)
 	}
 
-	timeWait := time.Now().Add(30 * time.Minute)
 	prevRefresh := time.Now()
+	timeWait := prevRefresh.Add(30 * time.Minute)
 	success := false
 
 	for time.Now().Before(timeWait) {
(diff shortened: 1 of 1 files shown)


cli/cmd/deploy.go
```diff
@@ -1029,3 +1061,76 @@ func updateUpgradeWithAgent(updateAgent *deploy.DeployAgent) error {
...
+		}
+
+		sort.Slice(depls.Items, func(i, j int) bool {
+			return depls.Items[i].CreationTimestamp.After(depls.Items[j].CreationTimestamp.Time)
+		})
+
+		depl := depls.Items[0]
+
+		// determine if the deployment has an appropriate number of ready replicas
+		minUnavailable := *(depl.Spec.Replicas) - getMaxUnavailable(depl)
```
abelanger5, line 1116: The `minUnavailable` threshold will be respected throughout the rolling deploy, so this condition would be true immediately. 

What we should check for is whether the number of replicas in the **new** replica set (which we can check via the `revision` annotation`) is greater than the max unavailable. For example, if there are 10 desired replicas, and max unavailable is 2, then we should check that there are 8 ready replicas **that belong to the new replicaset**.

abelanger5 (changes requested):

cli/cmd/deploy.go
```diff
@@ -1029,3 +1061,145 @@ func updateUpgradeWithAgent(updateAgent *deploy.DeployAgent) error {
...
+	}
+
+	sort.Slice(depls.Items, func(i, j int) bool {
+		return depls.Items[i].CreationTimestamp.After(depls.Items[j].CreationTimestamp.Time)
+	})
+
+	depl := depls.Items[0]
+
+	// determine if the deployment has an appropriate number of ready replicas
+	minUnavailable := *(depl.Spec.Replicas) - getMaxUnavailable(depl)
```
abelanger5, line 1104:
```suggestion
	minAvailable := *(depl.Spec.Replicas) - getMaxUnavailable(depl)
```

Isn't this the minimum available, rather than minimum unavailable?

abelanger5 (approved):

PR #2551, opened 2022-11-09
Change all template paths to be project-scoped

abelanger5:
## Pull request type

- [ ] Bugfix
- [ ] Feature
- [X] Other (please describe): improve scoping of template endpoints

## Pull request checklist

Endpoints to retrieve templates are user-scoped, rather than project-scoped. 

## What is the new behavior?

Make all template endpoints project-scoped. We've moved the project-scoped endpoints to `/api/v1` and we'll keep the old user-scoped template endpoints around for a few releases to prevent immediate breaking changes.

### DIFF
--- a/api/client/template.go
+++ b/api/client/template.go
@@ -35,7 +35,7 @@ func (c *Client) GetTemplate(
 
 	err := c.getRequest(
 		fmt.Sprintf(
-			"/v1/projects/%d/templates/%s/%s",
+			"/v1/projects/%d/templates/%s/versions/%s",
 			projectID,
 			name, version,
 		),
--- a/api/client/template.go
+++ b/api/client/template.go
@@ -9,13 +9,14 @@ import (
 
 func (c *Client) ListTemplates(
 	ctx context.Context,
+	projectID uint,
 	req *types.ListTemplatesRequest,
 ) (*types.ListTemplatesResponse, error) {
 	resp := &types.ListTemplatesResponse{}
 
 	err := c.getRequest(
 		fmt.Sprintf(
-			"/templates",
+			"/v1/projects/%d/templates", projectID,
 		),
 		req,
 		resp,
@@ -26,6 +27,7 @@ func (c *Client) ListTemplates(
 
 func (c *Client) GetTemplate(
 	ctx context.Context,
+	projectID uint,
 	name, version string,
 	req *types.GetTemplateRequest,
 ) (*types.GetTemplateResponse, error) {
@@ -33,7 +35,8 @@ func (c *Client) GetTemplate(
 
 	err := c.getRequest(
 		fmt.Sprintf(
-			"/templates/%s/%s",
+			"/v1/projects/%d/templates/%s/%s",
+			projectID,
 			name, version,
 		),
 		req,
--- a/api/server/handlers/v1/template/get.go
+++ b/api/server/handlers/v1/template/get.go
@@ -1,0 +1,105 @@
+package template
+
+import (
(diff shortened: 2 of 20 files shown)


api/client/template.go
```diff
@@ -26,14 +27,16 @@ func (c *Client) ListTemplates(
...
+	projectID uint,
 	name, version string,
 	req *types.GetTemplateRequest,
 ) (*types.GetTemplateResponse, error) {
 	resp := &types.GetTemplateResponse{}
 
 	err := c.getRequest(
 		fmt.Sprintf(
-			"/templates/%s/%s",
+			"/v1/projects/%d/templates/%s/%s",
```
mnafees, line 38: Can we keep this consistent with the `GET /api/v1/projects/{project_id}/clusters/{cluster_id}/namespaces/{namespace}/env_groups/{name}/versions/{version}` v1 API endpoint and scope it with the `/versions/` prefix?

api/server/handlers/v1/template/get.go
```diff
@@ -0,0 +1,105 @@
...
+	if !isValid {
+		t.HandleAPIError(w, r, apierrors.NewErrPassThroughToClient(
+			fmt.Errorf("invalid repo_url parameter"),
+			http.StatusBadRequest,
+		))
+
+		return
+	}
+
+	name, _ := requestutils.GetURLParamString(r, types.URLParamTemplateName)
```
mnafees, line 63: Should we be checking for the existence of this URL param? If it is empty the `LoadChartPublic()` might throw an error, right?

api/server/handlers/v1/template/get.go
```diff
@@ -0,0 +1,105 @@
...
+		return
+	}
+
+	parserDef := &parser.ClientConfigDefault{
+		HelmChart: chart,
+	}
+
+	res := &types.GetTemplateResponse{}
+	res.Metadata = chart.Metadata
+	res.Values = chart.Values
```
mnafees, line 88: We should set the `RepoURL` field as well here for more transparency.

api/server/handlers/v1/template/get_upgrade_notes.go
```diff
@@ -0,0 +1,105 @@
...
+	if !isValid {
+		t.HandleAPIError(w, r, apierrors.NewErrPassThroughToClient(
+			fmt.Errorf("invalid repo_url parameter"),
+			http.StatusBadRequest,
+		))
+
+		return
+	}
+
+	name, _ := requestutils.GetURLParamString(r, types.URLParamTemplateName)
```
mnafees, line 63: Same validation as the above comment.

api/server/router/v1/project.go
```diff
@@ -61,5 +86,185 @@ func getV1ProjectRoutes(
...
+	//   '403':
+	//     description: Forbidden
+	getTemplateEndpoint := factory.NewAPIEndpoint(
+		&types.APIRequestMetadata{
+			Verb:   types.APIVerbGet,
+			Method: types.HTTPVerbGet,
+			Path: &types.Path{
+				Parent: basePath,
+				RelativePath: fmt.Sprintf(
+					"%s/templates/{%s}/{%s}",
```
mnafees, line 179: Same as above comment in `api/client/template.go` about the `/versions/` prefix.

mnafees (changes requested):
Just a few comments, but looks great overall 👍🏽

mnafees (approved):

meehawk (approved):

PR #2553, opened 2022-11-09
Handle blue-green deployments from `LogSection` and fallback to live pods when `pod_values` fails

abelanger5:
## Pull request type

- [X] Bugfix
- [ ] Feature

## Pull request checklist

- Blue-green deployments don't return results for `pod_values` endpoints, because they don't have revision annotations 
- When `pod_values` returns `[]`, there are still cases where the live pods have logs in loki (this frequently happens when loki/promtail have been recently installed). 

## What is the new behavior?

- Modify query when deployment is using `bluegreen` to query for the correct pod prefix, as the image tags are part of the deployment names
- Fallback to live pods list when `pod_values` returns `[]`

### DIFF
--- a/dashboard/src/main/home/cluster-dashboard/expanded-chart/logs-section/LogsSection.tsx
+++ b/dashboard/src/main/home/cluster-dashboard/expanded-chart/logs-section/LogsSection.tsx
@@ -143,19 +143,24 @@ const LogsSection: React.FC<Props> = ({
       return;
     }
 
+    var filters = {
+      namespace: currentChart?.namespace,
+      revision: initData.revision ?? currentChart.version.toString(),
+      match_prefix: currentChart.name,
+    };
+
+    // if the current chart is set to a blue-green deployment, we don't set a revision, but instead
+    // we set the match prefix to the current chart and the active image tag.
+    if (currentChart.config.bluegreen?.enabled) {
+      filters.revision = null;
+      filters.match_prefix = `${currentChart.name}-${currentChart.config.bluegreen?.activeImageTag}`;
+    }
+
     api
-      .getLogPodValues(
-        "<TOKEN>",
-        {
-          namespace: currentChart?.namespace,
-          revision: initData.revision ?? currentChart.version.toString(),
-          match_prefix: currentChart.name,
-        },
-        {
-          project_id: currentProject.id,
-          cluster_id: currentCluster.id,
-        }
-      )
+      .getLogPodValues("<TOKEN>", filters, {
+        project_id: currentProject.id,
+        cluster_id: currentCluster.id,
+      })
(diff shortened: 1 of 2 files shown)


dashboard/src/main/home/cluster-dashboard/expanded-chart/logs-section/LogsSection.tsx
```diff
@@ -143,25 +143,64 @@ const LogsSection: React.FC<Props> = ({
       return;
     }
 
+    var filters = {
```
meehawk, line 146: Let's use const here

dashboard/src/main/home/cluster-dashboard/expanded-chart/logs-section/LogsSection.tsx
```diff
@@ -143,25 +143,64 @@ const LogsSection: React.FC<Props> = ({
...
+      namespace: currentChart?.namespace,
```
meehawk, line 147: Is there a situation where current chart won't be defined? If no then probably we don't need the ?. since I see we don't do this in the next line with currentChart.version

dashboard/src/main/home/cluster-dashboard/expanded-chart/logs-section/LogsSection.tsx
```diff
@@ -143,25 +143,64 @@ const LogsSection: React.FC<Props> = ({
...
+              "<TOKEN>",
+              {},
+              {
+                id: currentProject.id,
+                name: currentChart.name,
+                namespace: currentChart.namespace,
+                cluster_id: currentCluster.id,
+              }
+            )
+            .then((res: any) => {
```
meehawk, line 186: Can we move logic to this into an online function call and use async await here. Something like


(async function() {
  const logPodValues = await api.getLogPodValues.
  ...
  const getAllReleasePods = await...

} )()

dashboard/src/shared/api.tsx
```diff
@@ -1032,6 +1032,20 @@ const getMatchingPods = baseApi<
...
+const getAllReleasePods = baseApi<
+  {},
+  {
+    id: number;
+    name: string;
+    namespace: string;
+    cluster_id: number;
+  }
+>("GET", (pathParams) => {
+  let { id, name, cluster_id, namespace } = pathParams;
```
meehawk, line 1044: Let's use const here since variables aren't being reused

meehawk (changes requested):
Left some comments @abelanger5 but overall looks good. Is there a way I can test this locally with the blue green flag?

meehawk (commented):
Looks awesome 😎

meehawk (approved):

PR #2557, opened 2022-11-14
[POR-791] Use correct GAR image URI during launch flow

mnafees:
## Pull request type

- [x] Bugfix
- [ ] Feature

## Pull request checklist

When using a GAR registry, the new release image should be of the form `<region>-docker.pkg.dev/<GCP project ID>/<repo><image>`. The `porter create` command cases for this but the dashboard launch flow does not.

## What is the new behavior?

The dashboard launch flow cases on the image repo URI and if it belongs to a GAR registry, then it appends the `<image` part as mentioned above.

### DIFF
--- a/dashboard/src/main/home/launch/launch-flow/LaunchFlow.tsx
+++ b/dashboard/src/main/home/launch/launch-flow/LaunchFlow.tsx
@@ -71,16 +71,16 @@ const LaunchFlow: React.FC<PropsType> = (props) => {
   };
 
   const getFullActionConfig = (): FullActionConfigType => {
-    let imageRepoUri = `${selectedRegistry?.url}/${templateName}-${selectedNamespace}`;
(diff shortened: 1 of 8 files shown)


dashboard/src/main/home/launch/launch-flow/LaunchFlow.tsx
```diff
@@ -78,6 +78,11 @@ const LaunchFlow: React.FC<PropsType> = (props) => {
       imageRepoUri = selectedRegistry?.url;
     }
 
+    // Customize image repo URI for GAR
+    if (imageRepoUri.includes("pkg.dev")) {
```
meehawk, line 82: minor but can we rename this to `imageRepoURI` since I think in camelCasing we spell abbreviations with all caps.

meehawk (changes requested):

abelanger5 (approved):

meehawk (approved):

PR #2564, opened 2022-11-21
[POR-809] feat: support opening app details in new tab

meehawk:
## Pull request type

- [ ] Bugfix
- [x] Feature

## Pull request checklist

Currently the application details tab uses react-router to direct users to application details page. However, because of this, it is not possible to view the application details in a new tab

Issue Number: N/A

## What is the new behavior?

This PR allows users to open application details in new tab using a `CTRL/CMD + Click`. That way, they will be able to navigate to and open multiple tabs together>


portersupport:
@meehawk this looks great! One thing before merging this in is that if you select "Back" after expanding an application it doesn't route successfully the first time. We should fix this so that you don't need to select "Back" twice after expanding the application

PR #2579, opened 2022-12-01
[POR-817] feat: support checking updates for non porter charts

meehawk:
## Pull request type

- [ ] Bugfix
- [x] Feature

## Pull request checklist

- [] If it's a frontend change, Prettier has been run

Currently, we do not check `template updates` for non-porter charts or charts not found within the "cache" of porter supported chart repositories. This behaviour is limiting for a lost of users who would like to be notified when a user a template update is available for a user

## What is the new behavior?

This PR enables checking for non-porter charts. For this, the users have modify their helm charts to have the value of the `chart repo url` be the first url in `chart.metadata.sources` field.

- We try to fetch a index file from the first URL in the `chart.metadata.sources` object
- if this fails, we try to fetch the chart url from our cache
- failing to find a url in cache, we do not check for template updates


portersupport:
@meehawk looks like the template upgrade notes aren't working at all even with the default local charts in my dev environment: 
image

PR #2597, opened 2022-12-13
OpenTelemetry support

stefanmcshane:
# Description
In order to enable better telemetry for Porter Server (initially), we need to import OpenTelemetry (otel) libraries into Porter server. For the most part, we can use otel to auto-instrument our code. This will _not_ include context propagation as that will touch a lot of files.

# Summary of work
- Add telemetry package
- Create shared `InitTracer` function
- Add auto-instrumentation middleware for chi
- Add convenience functions and docs for general usecases of OpenTelemetry within Porter


portersupport:
Focus on first:

1. events tab
2. preview environments
3. application list components

stefanmcshane:
Merged in #3077 instead

PR #2656, opened 2023-01-24
[POR-815] Branch deployments should be auto deployed

meehawk:
## Pull request type

- [ ] Bugfix
- [x] Feature

## Pull request checklist

Currently,branch previews are not automatically created when the user adds a branch to the "Deploy branches' array for the environment.

This behaviour could be a drift from what is expected behaviour, that is, the branch to automatically create a preview deployment once tracked

## What is the new behavior?

This PR enables autocreation of preview deployments on creating a branch. The default for the environment variable is set to true which could be configured should we decide to turn this behaviour off in the future. We could also expose this as a setting at the environment level (which feels like the right way to do it)


api/server/handlers/environment/create.go
```diff
@@ -209,9 +211,115 @@ func (c *CreateEnvironmentHandler) ServeHTTP(w http.ResponseWriter, r *http.Requ
...
+		wg   sync.WaitGroup
+	)
+
+	for _, branch := range branches {
+		wg.Add(1)
+
+		go func(branch string) {
+			defer wg.Done()
+
+			namespace := fmt.Sprintf("previewbranch-%s-%s-%s", branch, strings.ReplaceAll(
```
mnafees, line 257: We no longer need this with custom namespaces in place for branch deployments.

api/server/handlers/environment/create.go
```diff
@@ -209,9 +211,115 @@ func (c *CreateEnvironmentHandler) ServeHTTP(w http.ResponseWriter, r *http.Requ
...
+			if err != nil {
+				errs = append(errs, err)
+			}
+		}(branch)
+	}
+
+	wg.Wait()
+
+	return errs
+}
```
mnafees, line 325: Needs a newline at the end of the file.

mnafees (changes requested):

api/server/handlers/environment/create.go
meehawk: Got it. How does it work since I might have missed the discussion on custom namespaces

meehawk: as in how do we fetch namespace for this?

stefanmcshane: Ive seen similar logic in a few places just before the new year. Do we have a function that auto calculates the namespace name based only on the passed-in parameters? .i.e doesnt rely on environment variables

The reason Im asking with the specific addendum on environment variables is because we should read those in, and parse those as close to startup time as possible, then not rely on them again where possible

mnafees: So when this was last implemented, we had hardcoded namespaces for branch deployments. But a recent fix has been pushed that now makes branch deployments namespace-agnostic. So we no longer need this and can completely ignore it.

mnafees: @stefanmcshane we got rid of that logic completely and now the namespace is totally dependent on the user to be chosen from.

api/server/handlers/environment/create.go
```diff
@@ -209,9 +211,115 @@ func (c *CreateEnvironmentHandler) ServeHTTP(w http.ResponseWriter, r *http.Requ
...
+				if err != nil {
+					errs = append(errs, fmt.Errorf("error creating deployment for branch %s: %w", branch, err))
+					return
+				}
+
+				deplID = depl.ID
+			} else if err != nil {
+				errs = append(errs, fmt.Errorf("error reading deployment for branch %s: %w", branch, err))
+				return
+			} else if onlyNewDeployments {
```
stefanmcshane, line 297: Nit pick here, but prioritising code readability is a good idea. A good way to do that is to "keep code left" as much as possible, reducing the number of potential branches that can be read.

This overall section could be written as
```go
if err != nil{
    if errors.Is(err, gorm.ErrRecordNotFound){
        ...gorm code handling
    }
}
if onlyNewDeployments {
    return
}

if deplID = 0 {
    return errors.New("no deployment ID")
}
```

api/server/handlers/environment/create.go
```diff
@@ -209,9 +211,115 @@ func (c *CreateEnvironmentHandler) ServeHTTP(w http.ResponseWriter, r *http.Requ
...
+						"pr_branch_from": branch,
+						"pr_branch_into": branch,
+					},
+				},
+			)
+
+			if err != nil {
+				errs = append(errs, err)
+			}
+		}(branch)
```
stefanmcshane, line 319: By now, the context of understanding why this random `(branch)` exists has left my mind. My personal rule of thumb is that if all the context to understand a function doesnt exist on one screen, refactor slightly and pull relevant pieces into separate files or functions. What's your thoughts on this?

api/server/handlers/environment/update_environment_settings.go
```diff
@@ -133,6 +133,23 @@ func (c *UpdateEnvironmentSettingsHandler) ServeHTTP(w http.ResponseWriter, r *h
...
 
 		env.GitDeployBranches = strings.Join(request.GitDeployBranches, ",")
+
+		if len(request.GitDeployBranches) > 0 && c.Config().ServerConf.EnableAutoPreviewBranchDeploy {
+			errs := autoDeployBranch(env, c.Config(), request.GitDeployBranches, true)
+
+			if len(errs) > 0 {
+				errString := errs[0].Error()
+
+				for _, e := range errs {
```
stefanmcshane, line 143: Could you use `errString := strings.Join(errs, ":")` for this?

api/server/handlers/environment/create.go
mnafees: I am not sure if I understand but this is passing a loop variable into a goroutine by convention and design.

stefanmcshane: I understand that -- my point here was moreso that a lot of our code's logic goes for 100s of lines and has a high cyclomatic complexity at almost every point. What I was intending to highlight here was that we could pull the core logic into another function, and leave the sync and wait in a much easier to understand format going forward .i.e refactor the 70 line difference between instantiating, and passing variables through to a few lines

```go
go func(branch string){
    wg.Add(1)
    defer wg.Done()
    functionWithLogic()
}(branch)
```

This way, we can drop all wg and closure related thinking from our mental ram, and focus only on the function with logic. It would also make the code _slightly_ more testable

api/server/handlers/environment/create.go
```diff
@@ -209,9 +211,127 @@ func (c *CreateEnvironmentHandler) ServeHTTP(w http.ResponseWriter, r *http.Requ
...
+		errs []error
+		wg   sync.WaitGroup
+	)
+
+	for _, branch := range branches {
+		wg.Add(1)
+
+		go func(branch string) {
+			defer wg.Done()
+			errs = append(errs, createWorkflowDispatchForBranch(env, config, onlyNewDeployments, branch)...)
```
stefanmcshane, line 256: This wont work as expected - the slice is being accessed across multiple routines. Pass in the slice to the `go func` to reduce the number of issues we will hit here

PR #2677, opened 2023-01-30
[POR-903] Add update helm repo handler

mnafees:
## Pull request type

- [ ] Bugfix
- [x] Feature

## Pull request checklist

N/A

## What is the new behavior?

This PR adds a new endpoint to update a Helm repo in a project.

### DIFF
--- a/api/server/handlers/helmrepo/update.go
+++ b/api/server/handlers/helmrepo/update.go
@@ -9,27 +9,47 @@ import (
 	"github.com/porter-dev/porter/api/server/shared"
 	"github.com/porter-dev/porter/api/server/shared/apierrors"
 	"github.com/porter-dev/porter/api/server/shared/config"
+	"github.com/porter-dev/porter/api/server/shared/requestutils"
 	"github.com/porter-dev/porter/api/types"
 	"github.com/porter-dev/porter/internal/models"
 	"gorm.io/gorm"
 )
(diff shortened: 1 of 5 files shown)


api/server/handlers/helmrepo/update.go
```diff
@@ -0,0 +1,92 @@
...
+	decoderValidator shared.RequestDecoderValidator,
+	writer shared.ResultWriter,
+) *HelmRepoUpdateHandler {
+	return &HelmRepoUpdateHandler{
+		PorterHandlerReadWriter: handlers.NewDefaultPorterHandler(config, decoderValidator, writer),
+	}
+}
+
+func (p *HelmRepoUpdateHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
+	proj, _ := r.Context().Value(types.ProjectScope).(*models.Project)
```
stefanmcshane, line 33: Do we worry that this will cause panics if an unsupported scope is passed through?

stefanmcshane (approved):
One question, not a blocker

api/server/handlers/helmrepo/update.go
mnafees: Nope, we have a middleware that checks for the correct scopes and errors out prematurely if the scope doesn't match.

PR #2699, opened 2023-02-15
Support AWS Assume Role for CLI

stefanmcshane:


stefanmcshane:
Closing in favour of cluster control plane changes, which will support assume role

PR #2711, opened 2023-02-27
Feroze compare chart versions

jusrhee:
## Pull request type

- [ ] Bugfix
- [X] Feature

## Pull request checklist

No ability to diff revisions

## What is the new behavior?

Diff config changes between revisions


dashboard/src/components/YamlEditor.tsx
```diff
@@ -4,6 +4,7 @@ import AceEditor from "react-ace";
 
 import "shared/ace-porter-theme";
 import "ace-builds/src-noconflict/mode-yaml";
+import 'brace/ext/searchbox';
```
ferozemohideen, line 7: this adds functionality to search in the helm values editor

dashboard/src/main/home/cluster-dashboard/expanded-chart/ExpandedChart.tsx
```diff
@@ -612,6 +615,7 @@ const ExpandedChart: React.FC<Props> = (props) => {
 
     setIsPreview(!isCurrent);
     getChartData(chart);
+    setCurrentChart(chart);
```
ferozemohideen, line 618: This is a temporary fix to an issue I was seeing where the correct version wasn't being selected in the version history when a user picked a different one. I saw that the previous method calls this method also, so in theory we shouldn't need this, but this is a quick hack

PR #2812, opened 2023-04-03
[hotfix] Add additional checks for authn handler support email

mnafees:


stefanmcshane:
We should remove this check altogether, instead of patching it as the functionality is no longer required.

stefanmcshane (approved):

PR #2817, opened 2023-04-04
[POR-981] Add option to choose registry for `porter.yaml`

mnafees:
## Pull request type

- [ ] Bugfix
- [x] Feature

## Pull request checklist

`porter.yaml` will always choose the first available registry in a project when creating a new app/addon.

## What is the new behavior?

This PR adds the functionality to choose a registry for an app/addon defined in a `porter.yaml` file in the following manner:
- A `resource`'s `target` can now have a `registry_url` param to define the registry URL 
- If the above does not exist, then we check if `porter config set-registry` was set and use the registry URL from there
- Lastly, we default to the previous behavior of listing the registries in the project and choosing the first one available to us

### DIFF
--- a/api/client/registry.go
+++ b/api/client/registry.go
@@ -100,6 +100,26 @@ func (c *Client) ListRegistries(
 	return resp, err
 }
 
+// GetRegistry returns a registry given a project id and registry id
+func (c *Client) GetRegistry(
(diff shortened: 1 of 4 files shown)


rudimk (approved):
Lookin' killer

PR #2839, opened 2023-04-11
Setup support for multiple deployment environments

stefanmcshane:
- Rename cloudformation flow role to the same as the credentials flow role name 
- Added support for finding image tags with Assume Role clusters
- Added `GITHUB_APP_SECRET_BASE64` for injecting github app secret without needing volumes
- Update default CIDR range to be in `10.0.0.0/8`
- Hide results from Assume role chains, to only show customer accounts for ease of use
- Setup support for new production cluster, and internal tooling cluster


.github/workflows/porter_production.yml
```diff
@@ -1,25 +1,88 @@
 "on":
   push:
     branches:
-    - master
+      - master
+      - rename-porter-role
```
stefanmcshane, line 5: will remove before merging - allows merging to the other cluster on this branch

jusrhee (approved):

PR #3066, opened 2023-05-26
add telemetry to preview environment deployment handlers

d-g-town:
Revision 2:

- Moved `InitTracer` to `main.go`
- Only swallow `telemetry.Error` if it is not used in the `HandleAPIError`/used in a super interpolated fashion
- add zone and offset to span if time attribute
____

Adding telemetry to preview environment creation handlers so we can start to get a sense for errors that are blocking deployments.

This is an initial introduction of telemetry that captures the request parameters and errors that cause us to exit.  If we exit for a reason that is not our fault (like a PR being closed), then I add the attribute `unsuccessful-exit-reason` to the span.  Future commits should extend the telemetry into the functions being called.

### DIFF
--- a/api/server/handlers/environment/create_deployment_by_cluster.go
+++ b/api/server/handlers/environment/create_deployment_by_cluster.go
@@ -34,14 +34,11 @@ func NewCreateDeploymentByClusterHandler(
 }
 
 func (c *CreateDeploymentByClusterHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
-	tracer, _ := telemetry.InitTracer(r.Context(), c.Config().TelemetryConfig)
-	defer tracer.Shutdown()
-
 	ctx, span := telemetry.NewSpan(r.Context(), "serve-create-deployment-by-cluster")
 	defer span.End()
 
-	project, _ := r.Context().Value(types.ProjectScope).(*models.Project)
-	cluster, _ := r.Context().Value(types.ClusterScope).(*models.Cluster)
+	project, _ := ctx.Value(types.ProjectScope).(*models.Project)
+	cluster, _ := ctx.Value(types.ClusterScope).(*models.Cluster)
 
 	telemetry.WithAttributes(span,
 		telemetry.AttributeKV{Key: "project-id", Value: project.ID},
@@ -72,7 +69,7 @@ func (c *CreateDeploymentByClusterHandler) ServeHTTP(w http.ResponseWriter, r *h
 		project.ID, cluster.ID, request.RepoOwner, request.RepoName,
 	)
 	if err != nil {
-		_ = telemetry.Error(ctx, span, err, "error reading environment by owner repo name")
+		err = telemetry.Error(ctx, span, err, "error reading environment by owner repo name")
 
 		if errors.Is(err, gorm.ErrRecordNotFound) {
 			c.HandleAPIError(w, r, apierrors.NewErrNotFound(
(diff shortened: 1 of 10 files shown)


api/server/handlers/environment/create_deployment_by_cluster.go
```diff
@@ -33,21 +34,46 @@ func NewCreateDeploymentByClusterHandler(
 }
 
 func (c *CreateDeploymentByClusterHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
+	tracer, _ := telemetry.InitTracer(r.Context(), c.Config().TelemetryConfig)
```
portersupport, line 37: This should be setup once globally when the server is started. Bit of an antipattern imo, but thats the recommendations atm

stefanmcshane: Wrong account, my bad.

api/server/handlers/environment/create_deployment_by_cluster.go
```diff
@@ -33,21 +34,46 @@ func NewCreateDeploymentByClusterHandler(
...
+	defer tracer.Shutdown()
+
+	ctx, span := telemetry.NewSpan(r.Context(), "serve-create-deployment-by-cluster")
+	defer span.End()
+
 	project, _ := r.Context().Value(types.ProjectScope).(*models.Project)
```
stefanmcshane, line 43: Changes these to also use the `ctx` that you created above, or you will 🥁 lose context 🥁

api/server/handlers/environment/enable_pull_request.go
```diff
@@ -34,17 +35,42 @@ func NewEnablePullRequestHandler(
 }
 
 func (c *EnablePullRequestHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
+	tracer, _ := telemetry.InitTracer(r.Context(), c.Config().TelemetryConfig)
+	defer tracer.Shutdown()
```
stefanmcshane, line 39: Same as above - you will cause spans to be lost by shutting down their tracer on a per-request basis

api/server/handlers/environment/enable_pull_request.go
```diff
@@ -124,9 +160,16 @@ func (c *EnablePullRequestHandler) ServeHTTP(w http.ResponseWriter, r *http.Requ
 			},
 		},
 	)
+	if err != nil {
+		_ = telemetry.Error(ctx, span, err, "error creating workflow dispatch event")
+		c.HandleAPIError(w, r, apierrors.NewErrInternal(err))
```
stefanmcshane, line 165: Any reason not to use the returned error from the line above, as the internal error here?

internal/telemetry/span.go
```diff
@@ -80,6 +84,8 @@ func WithAttributes(span trace.Span, attrs ...AttributeKV) {
 				span.SetAttributes(attribute.Float64(prefixSpanKey(string(attr.Key)), val))
 			case bool:
 				span.SetAttributes(attribute.Bool(prefixSpanKey(string(attr.Key)), val))
+			case time.Time:
```
stefanmcshane, line 87: It might be worthwhile to not force to UTC here, as it could remove context that we may need. Time is a tough cookie to manage. Instead, consider forcing to string, and adding the zone as a secondary attribute.

Its definitely good practice to force to UTC when you control the code however

api/server/handlers/environment/enable_pull_request.go
d-g-town: Just wanted to show that I was explicitly swallowing the errors in cases where the HandleAPIError inputs were super interpolated, but I'll add them back in where there is no interpolation.

api/server/handlers/environment/create_deployment_by_cluster.go
d-g-town: Ahhh, got it

stefanmcshane (approved):

PR #3073, opened 2023-05-30
Fix Email Overflow

sdess09:
## Pull request type

- [ ] Bugfix
- [ ] Feature

## Pull request checklist


jusrhee:
if we're going to animate this we should stop scrolling once it reaches the end of the text to avoid this kind of state:
image

if that's too much trouble tooltip is probably the move

PR #3077, opened 2023-05-30
OpenTelemetry on all endpoints

stefanmcshane:
- Bump chi to chi/v5 per https://github.com/go-chi/chi/blob/master/CHANGELOG.md#v500-2021-02-27
- Added `riandyrn/otelchi` to all endpoints
- Protected `telemetry.Shutdown()` in cases of errors

### DIFF
--- a/api/server/handlers/stacks/list_porter_app_events.go
+++ b/api/server/handlers/stacks/list_porter_app_events.go
@@ -38,12 +38,19 @@ func NewPorterAppEventListHandler(
 }
 
 func (p *PorterAppEventListHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
-	ctx := r.Context()
+	ctx, span := telemetry.NewSpan(r.Context(), "serve-list-porter-app-events")
+	defer span.End()
+
 	cluster, _ := ctx.Value(types.ClusterScope).(*models.Cluster)
+	telemetry.WithAttributes(span,
+		telemetry.AttributeKV{Key: "cluster-id", Value: int(cluster.ID)},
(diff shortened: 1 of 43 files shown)


api/server/handlers/stacks/list_porter_app_events.go
```diff
@@ -38,20 +38,28 @@ func NewPorterAppEventListHandler(
...
 	cluster, _ := ctx.Value(types.ClusterScope).(*models.Cluster)
+	telemetry.WithAttributes(span,
+		telemetry.AttributeKV{Key: "cluster-id", Value: int(cluster.ID)},
+		telemetry.AttributeKV{Key: "project-id", Value: int(cluster.ProjectID)},
+	)
 
 	stackName, reqErr := requestutils.GetURLParamString(r, types.URLParamStackName)
 	if reqErr != nil {
-		p.HandleAPIError(w, r, apierrors.NewErrPassThroughToClient(reqErr, http.StatusBadRequest))
+		e := telemetry.Error(ctx, span, nil, "error parsing stack name from url")
```
d-g-town, line 52: do we want to pass in `reqErr`?

d-g-town: Same question for `err` in the other telemetry.Error calls below

d-g-town (approved):

api/server/handlers/stacks/list_porter_app_events.go
stefanmcshane: good catch

stefanmcshane: reqErr is a specific type, so it's only valid in this one case. The rest are all of type `error`

PR #3090, opened 2023-06-01
telemetry on resource usage

d-g-town:
Add telemetry on usage quotas.

### DIFF
--- a/api/server/router/middleware/usage.go
+++ b/api/server/router/middleware/usage.go
@@ -72,22 +72,7 @@ func (b *UsageMiddleware) Middleware(next http.Handler) http.Handler {
 
 		r = r.Clone(ctx)
 
-		// if allowed {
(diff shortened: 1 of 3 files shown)


api/server/router/middleware/usage.go
```diff
@@ -45,25 +54,40 @@ func (b *UsageMiddleware) Middleware(next http.Handler) http.Handler {
...
-				apierrors.NewErrPassThroughToClient(
-					fmt.Errorf(UsageErrFmt, b.metric, limit, curr),
-					http.StatusBadRequest,
-				),
-				true,
-			)
-		}
+		r = r.Clone(ctx)
+
+		// if allowed {
```
stefanmcshane, line 75: Remove instead of commenting out. The old code is accessible in the git history

PR #3095, opened 2023-06-01
add project, cluster, and user-id to all request endpoints automatically

stefanmcshane:
- Create middleware for hydrating all spans conditionally, dependent on their context values. Support for user-id, cluster-id, and project-id added initially
- Add the same id's as above to all newly created spans throughout, without using baggage

### DIFF
--- a/internal/telemetry/span.go
+++ b/internal/telemetry/span.go
@@ -10,6 +10,8 @@ import (
 	"time"
 
 	"github.com/google/uuid"
+	"github.com/porter-dev/porter/api/types"
+	"github.com/porter-dev/porter/internal/models"
 	"go.opentelemetry.io/otel"
 	"go.opentelemetry.io/otel/attribute"
 	"go.opentelemetry.io/otel/codes"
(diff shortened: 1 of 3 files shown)


internal/telemetry/span.go
```diff
@@ -19,33 +21,29 @@ import (
...
-	// 	AttributeKV{Key: AttributeKeyCluster, Value: stringFromContext(ctx, types.ClusterScope)},
-	// 	AttributeKV{Key: AttributeKeyProject, Value: stringFromContext(ctx, types.ProjectScope)},
-	// )
+	cluster, ok := ctx.Value(types.ClusterScope).(*models.Cluster)
+	if ok {
+		WithAttributes(span, AttributeKV{Key: "cluster-id", Value: cluster.ID})
+	}
 
-	return ctx, span
+	if project, ok := ctx.Value(types.ProjectScope).(*models.Project); ok {
```
d-g-town, line 44: why not do this `; ok {` syntax on all of them?

d-g-town (approved):

internal/telemetry/span.go
stefanmcshane: 1. consistency is boring
2. i made a mistake

Choose your fighter

stefanmcshane: 1. consistency is boring
2. i made a mistake

Choose your fighter

PR #3101, opened 2023-06-02
Greatly simplify the amount of work the front-end has to do to find logs; push all that logic to the backend

ferozemohideen:
And because of the refactors, we can show timestamps now too! neato
image

## Pull request type

- [ ] Bugfix
- [ ] Feature

## Pull request checklist

### DIFF
--- a/api/server/handlers/porter_app/get_logs_within_time_range.go
+++ b/api/server/handlers/porter_app/get_logs_within_time_range.go
@@ -13,6 +13,7 @@ import (
 	"github.com/porter-dev/porter/api/types"
 	porter_agent "github.com/porter-dev/porter/internal/kubernetes/porter_agent/v2"
 	"github.com/porter-dev/porter/internal/models"
+	"github.com/porter-dev/porter/internal/telemetry"
 	v1 "k8s.io/api/core/v1"
 )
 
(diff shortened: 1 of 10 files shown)


api/server/handlers/porter_app/get_logs_within_time_range.go
```diff
@@ -0,0 +1,129 @@
...
+			var latestPod *v1.Pod
+			for _, v := range podVals {
+				name := strings.Split(v, "-hook")[0] + "-hook"
+				pods, err := agent.GetJobPods(request.Namespace, name)
+				if err != nil {
+					c.HandleAPIError(w, r, apierrors.NewErrInternal(err))
+					return
+				}
+				for _, pod := range pods {
+					if pod.GetCreationTimestamp().Time.After(*request.StartRange) && pod.GetCreationTimestamp().Time.Before(*request.EndRange) {
```
stefanmcshane, line 98: Do we crash if a request is made where start range or end range are not provided?

api/types/incident.go
```diff
@@ -112,6 +112,19 @@ type GetLogRequest struct {
 	Direction   string     `schema:"direction"`
 }
 
+// You may either provide the pod selector directly, or the chart name,
+// in which case we will attempt to find the correct pod within the timeframe.
+type GetChartLogsWithinTimeRangeRequest struct {
+	ChartName   string     `schema:"chart_name"`
+	Limit       uint       `schema:"limit"`
+	StartRange  *time.Time `schema:"start_range"`
```
stefanmcshane, line 120: use time.Time, with omitempty in the tag, unless there is a specific reason not to. Lets avoid those nil pointers where possible

stefanmcshane (changes requested):
None of the new endpoints return useful errors to the user, and a few nil pointer potentials in there also

api/server/handlers/porter_app/get_logs_within_time_range.go
```diff
@@ -33,30 +34,40 @@ func NewGetLogsWithinTimeRangeHandler(
...
+	if request.StartRange.IsZero() || request.EndRange.IsZero() {
+		err := telemetry.Error(ctx, span, nil, "must provide start and end range")
+		c.HandleAPIError(w, r, apierrors.NewErrPassThroughToClient(err, http.StatusBadRequest))
+		return
+	}
+
 	agent, err := c.GetAgent(r, cluster, "")
 	if err != nil {
-		c.HandleAPIError(w, r, apierrors.NewErrInternal(err))
+		_ = telemetry.Error(ctx, span, err, "unable to get agent")
```
stefanmcshane, line 55: you should be able to assign this to err, then pass it in as the `NewErrPassThroughToClient` for the same results

api/server/handlers/porter_app/get_logs_within_time_range.go
```diff
@@ -33,30 +34,40 @@ func NewGetLogsWithinTimeRangeHandler(
 }
 
 func (c *GetLogsWithinTimeRangeHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
-	cluster, _ := r.Context().Value(types.ClusterScope).(*models.Cluster)
+	ctx, span := telemetry.NewSpan(r.Context(), "serve-get-logs-within-time-range")
+	defer span.End()
+	r = r.Clone(ctx)
```
stefanmcshane, line 39: You dont need to do this, unless you are chaining requests like in middleware

stefanmcshane (approved):

PR #3104, opened 2023-06-02
Basic deploy event

jusrhee:
## Pull request type

- [ ] Bugfix
- [ ] Feature

## Pull request checklist


stefanmcshane (approved):

dashboard/src/main/home/app-dashboard/expanded-app/activity-feed/events/DeployEventCard.tsx
```diff
@@ -25,9 +25,9 @@ const DeployEventCard: React.FC<Props> = ({ event, appData }) => {
   const renderStatusText = (event: PorterAppEvent) => {
     switch (event.status) {
       case "SUCCESS":
-        return <Text color="#68BF8B">Deployment succeeded.</Text>;
+        return <Text color="#68BF8B">Deployment succeeded</Text>;
```
rgbmdq, line 28: It would be beneficial to introduce constants for colors instead of using the string value directly ('#68BF8B') in multiple places throughout the code. By defining color constants, you can achieve better consistency, enhance maintainability, and make it easier to update or modify the colors in the future.

Consider creating a separate module or file specifically for color constants, where you can define and export the color values. For example, you could define a constant named SUCCESS_COLOR with a value of '#68BF8B' in that module.

Then, in the code snippet you provided, you can import the color constant and use it instead of the hard-coded string value. This approach ensures that the same color is consistently used across different instances and can be easily updated if needed.

PR #3139, opened 2023-06-08
Enable prefix delegation, and larger EKS subnets

stefanmcshane:
- Enable /21 subnets by default
- Allow toggling of aws cni prefix delegation


stefanmcshane:
No longer continuing with this support

PR #3145, opened 2023-06-09
Adding basic telemetry to cli for porter apply

ferozemohideen:
## Pull request type

- [ ] Bugfix
- [ ] Feature

## Pull request checklist


stefanmcshane:
We currently do something similar with Sentry for the CLI. Take some notes from that around what we enable by default, and what we do not enable as some users ran straight into issues with that once we enabled it.

Overall, I think that the first PR into the CLI should include:
- `ENABLE_TELEMETRY` which defaults to or false, and make sure everything is gated there
- The CLI will have no access to the open telemetry collector which runs on our production cluster. We can either create a new endpoint for ingesting traces and forwarding them to the collector, or else forward them from the CLI directly to honeycomb. Pros and cons to all approaches here.

PR #3168, opened 2023-06-20
more azure support

d-g-town:
Three changes:

1. CCP CreateAzureConnection api call now returns a credential id, which needs to be passed the contract on the frontend
2. new CCP ListRepositories, TokenForRegistry, and DockerConfigFileForRegistry api calls replace the old integration id methods (for the DockerConfigFile, this means I need to pass the CCP client all the way down to the call, which I do in part by adding it to the Upgrade/InstallChart configs)
3. unnecessary formatting courtesy of goland 

This PR should not be merged until these two are merged: 
https://github.com/porter-dev/api-contracts/pull/4
https://github.com/porter-dev/cluster-control-plane/pull/104

### DIFF
--- a/api/server/handlers/registry/get_token.go
+++ b/api/server/handlers/registry/get_token.go
@@ -409,6 +409,20 @@ func (c *RegistryGetACRTokenHandler) ServeHTTP(w http.ResponseWriter, r *http.Re
 
 	telemetry.WithAttributes(span, telemetry.AttributeKV{Key: "project-id", Value: proj.ID})
 
+	request := &types.GetRegistryACRTokenRequest{}
+
+	if ok := c.DecodeAndValidate(w, r, request); !ok {
+		err := telemetry.Error(ctx, span, nil, "error decoding request")
+		c.HandleAPIError(w, r, apierrors.NewErrPassThroughToClient(err, http.StatusInternalServerError))
+		return
+	}
+
+	if request.ServerURL == "" {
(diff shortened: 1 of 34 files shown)


go.mod
```diff
@@ -2,6 +2,8 @@ module github.com/porter-dev/porter
 
 go 1.20
 
+replace github.com/porter-dev/api-contracts v0.0.63 => ../api-contracts
```
d-g-town, line 5: I'll update this to the new `api-contracts` version when that is released

api/server/handlers/registry/get_token.go
```diff
@@ -402,35 +402,77 @@ func NewRegistryGetACRTokenHandler(
...
+				})
+				tokenResp, err := c.Config().ClusterControlPlaneClient.TokenForRegistry(ctx, tokenReq)
+				if err != nil {
+					err := telemetry.Error(ctx, span, err, "error getting token response from ccp")
+					c.HandleAPIError(w, r, apierrors.NewErrPassThroughToClient(err, http.StatusInternalServerError))
+					return
+				}
+
+				if tokenResp.Msg == nil || tokenResp.Msg.Token == "" {
+					err := telemetry.Error(ctx, span, err, "no token found in response")
```
ferozemohideen, line 442: watch out for this spans where err isn't explicitly set 
```suggestion
					err := telemetry.Error(ctx, span, nil, "no token found in response")
```

PR #3204, opened 2023-06-27
fixing bug where buildpacks aren't deleted properly and adding end time for build event

ferozemohideen:
Note that this will not correct build times for workflow runs that are retried, since the updated_at attribute is updated for each try. If the user doesn't list the events before every retry, the build time will be incorrect. This is a minor bug tho

## Pull request type

- [x] Bugfix
- [ ] Feature

## Pull request checklist


api/server/handlers/porter_app/create.go
```diff
@@ -362,42 +362,14 @@ func (c *CreatePorterAppHandler) ServeHTTP(w http.ResponseWriter, r *http.Reques
 			return
 		}
 
-		if request.RepoName != "" {
```
stefanmcshane, line 365: Whats the benefit in removing the checks? How does the github API deal with default values?

stefanmcshane (approved):
Approving , but reconsider not sanitising your inputs at an API boundary. Check every value unless there is a strong reason not to

api/server/handlers/porter_app/create.go
ferozemohideen: @stefanmcshane originally, I only removed the check around buildpacks, because the bug was that if a user decided to remove all buildpacks, then request.Buildpacks would be "", and the DB model would not be updated. We had a hacky workaround such that the frontend would pass in "null" as request.Buildpacks if we wanted to remove them all, but I didn't like that approach (what if a user had a single buildpack named "null", for example?)

Do you see a better way to handle the operation of resetting buildpacks (or any other attribute) to "" in the DB?

stefanmcshane: Have you tried setting `omitempty` in the struct tag?

stefanmcshane: My rule of thumb for databases is to avoid nulls where possible. 
This removes a stupid amount of errors in application and database code. Use a default "unspecified" value instead, or a default value such as "" which is set in the database code, not the application code.

Other than that, split up the API struct from the DB struct. Convert between the 2, and avoid pointers unless you absolutely need them. We can talk about performance once we pull profiling graphs which are actually affecting the product

PR #3255, opened 2023-07-11
handle missing resource providers and quota limit preflight checks

d-g-town:
Display troubleshooting tips for missing resource providers and quota limit preflight errors during Azure provisioning.


dashboard/src/components/porter/Error.tsx
```diff
@@ -18,15 +18,15 @@ const Error: React.FC<Props> = ({
...
+
   return (
     <>
       <StyledError>
         <i className="material-icons">error_outline</i>
         <Block>
         <Bold>Error:</Bold>
         <Text>{message}</Text>
-        {ctaText && (
+        {ctaText && (errorModalContents != null || ctaOnClick != null) && (
```
d-g-town, line 29: Changing this because my code was erroring out when I accidentally passed in a null `errorModalContents`, and it tried to call `ctaOnClick` without checking if it was not null.

ferozemohideen (approved):
lgtm

PR #3265, opened 2023-07-12
auth management server

d-g-town:
- setup auth-management-server; add to tilt
- fix config restart


cmd/authmanagement/main.go
```diff
@@ -0,0 +1,27 @@
...
+	version := "v0.0.0-dgtown"
+
+	cl := loader.NewEnvLoader(version)
+
+	config, err := cl.LoadConfig()
+	if err != nil {
+		log.Fatal("Config loading failed: ", err)
+	}
+
+	authManagementServer := authmanagement.AuthManagementServer{Port: config.ServerConf.AuthManagementServerPort, Config: config}
```
stefanmcshane, line 21: It might seem counter-intuitive, but separate out the config that the new grpc server uses for running, from the config that the Porter API uses to consume or connect to that server.
The benefit is that the 2 become completely decoupled.

PR #3269, opened 2023-07-13
setup auth-management-server; add to tilt

d-g-town:
Adding `--auth` flag to optionally start up AuthManagement grpc server in addition to Porter API.


api/authmanagement/server.go
```diff
@@ -0,0 +1,81 @@
...
+	"github.com/joeshaw/envdecode"
+
+	"github.com/bufbuild/connect-go"
+	otelconnect "github.com/bufbuild/connect-opentelemetry-go"
+	"github.com/porter-dev/api-contracts/generated/go/porter/v1/porterv1connect"
+	"golang.org/x/net/http2"
+	"golang.org/x/net/http2/h2c"
+)
+
+type Config struct {
```
stefanmcshane, line 18: Add comments to these to explain what they do. Config and Port are generic and Ive no idea what they are actually for

api/authmanagement/server.go
```diff
@@ -0,0 +1,81 @@
...
+	Port                 int    `env:"AUTH_MANAGEMENT_SERVER_PORT,default=8090"`
+	TokenGeneratorSecret string `env:"TOKEN_GENERATOR_SECRET,default=secret"`
+}
+
+// AuthManagementServer contains all configuration options for the AuthManagementServer, implementing the
+// gRPC server's interface
+type AuthManagementServer struct {
+	Config *Config
```
stefanmcshane, line 26: Why does this need to be a pointer?

api/server/server.go
```diff
@@ -0,0 +1,51 @@
...
+	"net/http"
+
+	"github.com/go-chi/chi/v5"
+	"github.com/porter-dev/porter/api/server/shared/config/env"
+)
+
+// PorterAPIServer contains the routing and configuration options for starting the PorterAPIServer
+type PorterAPIServer struct {
+	Port       int
+	Handler    *chi.Mux
```
stefanmcshane, line 15: Does handler make sense here?

cmd/app/main.go
```diff
@@ -48,30 +61,68 @@ func main() {
 	}
 	config.Logger.Info().Msg("Initialed data")
 
+	// ignore error so that telemetry is not required
+	tracer, err := telemetry.InitTracer(ctx, config.TelemetryConfig)
```
stefanmcshane, line 65: Can this error out in other ways? They arent handled here.

cmd/app/main.go
```diff
@@ -48,30 +61,68 @@ func main() {
...
+
+			config.Logger.Info().Msgf("Starting AuthManagement server on port %d", a.Config.Port)
+			if err := a.ListenAndServe(ctx); err != nil && err != http.ErrServerClosed {
+				return fmt.Errorf("AuthManagement server failed: %s", err.Error())
+			}
+			config.Logger.Info().Msg("Shutting down AuthManagement server")
+			return nil
+		})
+	} else {
+		config.Logger.Info().Msgf("Not starting AuthManagement server")
```
stefanmcshane, line 104: Move this into the above if statement. No need for an else

stefanmcshane (approved):
Approving, but needs some renaming (and comments) so that the code is easier to understand. A lot of assumed knowledge here

PR #3270, opened 2023-07-14
Updated Prometheus Service LabelSelectors

rudimk:
Modified the GetPrometheusService() function to search for Prometheus services based on updated chart label selectors.

## Pull request type

- [X] Bugfix
- [ ] Feature

## Pull request checklist

At the moment, the backend looks for a Prometheus `Service` it can query, using these selectors: `app=prometheus,component=server,heritage=Helm`. This works for the Prometheus chart @v15.5.3, but more updated versions use different labels. 

## What is the new behavior?

Modified the `GetPrometheusService()` function to search for Prometheus services based on updated chart label selectors - it now searches for both old and new selectors, and returns whatever's on the cluster.

### DIFF
--- a/internal/kubernetes/prometheus/metrics.go
+++ b/internal/kubernetes/prometheus/metrics.go
@@ -14,18 +14,34 @@ import (
 
 // returns the prometheus service name
 func GetPrometheusService(clientset kubernetes.Interface) (*v1.Service, bool, error) {
-	services, err := clientset.CoreV1().Services("").List(context.TODO(), metav1.ListOptions{
+	// The prometheus-community/prometheus chart @ v15.5.3 uses non-FQDN labels.
+	redundantServices, err := clientset.CoreV1().Services("").List(context.TODO(), metav1.ListOptions{
 		LabelSelector: "app=prometheus,component=server,heritage=Helm",
 	})
 	if err != nil {
 		return nil, false, err
 	}
 
-	if len(services.Items) == 0 {
+	// OTOH the same chart @ v22.6.2 uses more standardised labels.
+	upgradedServices, err := clientset.CoreV1().Services("").List(context.TODO(), metav1.ListOptions{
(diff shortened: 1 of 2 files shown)


stefanmcshane:
Overall this looks good. If there is a preferred label that will be the most commonly selected one, consider performing the entire check for that one first, before going to the second one. If you have `selectorA` and `selectorB`, the flow could look like:
- Make `selectorA` query
- Check `selectorA` results and return if found
- Make `selectorB` query
- Check `selectorB` results and return if found

internal/kubernetes/prometheus/metrics.go
```diff
@@ -14,18 +14,34 @@ import (
...
 	}
 
-	return &services.Items[0], true, nil
+	// Now that we know there is a compatible Prometheus in here somewhere, we'll send the one that's availble.
+	if len(redundantServices.Items) > 0 {
+		return &redundantServices.Items[0], true, nil
+	}
+	if len(upgradedServices.Items) > 0 {
+		return &upgradedServices.Items[0], true, nil
+	}
```
stefanmcshane, line 44: The linter is complaining here because there is no return statement. Running `GOWORK=off go vet ./...` will find these problems locally.

internal/kubernetes/prometheus/metrics.go
```diff
@@ -14,18 +14,34 @@ import (
...
+	// The prometheus-community/prometheus chart @ v15.5.3 uses non-FQDN labels.
+	redundantServices, err := clientset.CoreV1().Services("").List(context.TODO(), metav1.ListOptions{
 		LabelSelector: "app=prometheus,component=server,heritage=Helm",
 	})
 	if err != nil {
 		return nil, false, err
 	}
 
-	if len(services.Items) == 0 {
+	// OTOH the same chart @ v22.6.2 uses more standardised labels.
```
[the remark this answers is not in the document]
rudimk: 10-4, roger. Pushed a change.

internal/kubernetes/prometheus/metrics.go
rudimk: Aye, fixed this too.

PR #3314, opened 2023-07-25
SyncedEnvGroup Revamp

sdess09:
Environment groups are stored in `porter-env-group` namespace with a the name `environment-group-name.environment-group-version` and a corresponding label `porter.run/environment-group-name` and `porter.run/environment-group-version` which are used for querying.

An envgroup consists of a configmap which contains all non-secret values, and a secret which has all secret values. If an env group only has non-secret values, a secret will be created anyway for parity, and vice-versa.

Applications are linked to preview groups by adding a label called `porter.run/linked-environment-group`. This uses a dot separator which refers to the non-versioned environment-group-name. 
At deploy-time, we check the label on the applications, and ensure that the `porter-env-group` environment group parts are synced into the target environment. These can then be referenced as a configMapRef and a secretRef. We chose this pattern as we can use a CSI in future to move the secrets and configs to a 3rd party provider of the customer's choice. This is not on the roadmap, but just given for context.

To list all applications that are linked to an application, we just search all deployments and cronjobs for the label `porter.run/linked-environment-group` which should have a value which equals a value in the `porter-env-group` namespace with the label `porter.run/environment-group-name`.

To override an environment variable in an application which has an env group attached, we just inject this into the k8s manifest as a raw value. 

**Testing**:
Every step below should end in the following steps:
1. Ensure the label is in the helm values, and the configMapRef and secretMapRef exists
2. Ensure pod spins up and contains all values from the env group, and expected overrides

Scenarios:
------
1. Create an env group
2. Create an application and attach the env groups to a web chart
3. Create an application and attach the env groups to a job chart
4. Create an application and attach the env groups to a worker chart
------
1. Create another env group
2. Create another application attaching both env groups to it
------

1. Create another application with a web, worker, and job service
2. Attach both env groups to this service
------

1. Create another application with a web, worker, and job service
2. Attach both env groups to this service
3. Add an override environment variable to the application
------

1. Push an update to the environment group which was not overridden
2. Ensure that all above applications were updated and that the override stays
------

1. Push an update to the environment group which was overridden
2. Ensure that all above applications were updated and that the override stays
------

1. Create an application with only a job
2. Attach an env group
3. Add a new service to the application


stefanmcshane (approved):

ferozemohideen (changes requested):
Haven't taken a full look yet but can we remove all commented code? it bloats the codebase

api/server/handlers/environment_groups/delete.go
```diff
@@ -0,0 +1,71 @@
...
+	)
+
+	agent, err := c.GetAgent(r, cluster, "")
+	if err != nil {
+		err := telemetry.Error(ctx, span, err, "unable to connect to kubernetes cluster")
+		c.HandleAPIError(w, r, apierrors.NewErrPassThroughToClient(err, http.StatusInternalServerError))
+		return
+	}
+
+	err = environment_groups.DeleteEnvironmentGroup(ctx, agent, request.Name)
```
ferozemohideen, line 65: why are these methods defined in environment_groups and not just agent methods? wouldn't it be cleaner to call something like agent.DeleteEnvironmentGroup(ctx, request.Name)?

api/server/handlers/environment_groups/delete.go
```diff
@@ -0,0 +1,71 @@
...
+	}
+
+	telemetry.WithAttributes(span,
+		telemetry.AttributeKV{Key: "environment-group-name", Value: request.Name},
+	)
+
+	agent, err := c.GetAgent(r, cluster, "")
+	if err != nil {
+		err := telemetry.Error(ctx, span, err, "unable to connect to kubernetes cluster")
+		c.HandleAPIError(w, r, apierrors.NewErrPassThroughToClient(err, http.StatusInternalServerError))
```
ferozemohideen, line 61: `http.StatusServiceUnavailable`

internal/kubernetes/environment_groups/list.go
```diff
@@ -0,0 +1,227 @@
...
+}
+
+// ListEnvironmentGroups returns all environment groups stored in the provided namespace. If none is set, it will use the namespace "porter-env-group"
+func ListEnvironmentGroups(ctx context.Context, a *kubernetes.Agent, listOpts ...EnvironmentGroupOption) ([]EnvironmentGroup, error) {
+	ctx, span := telemetry.NewSpan(ctx, "list-environment-groups")
+	defer span.End()
+
+	var opts environmentGroupOptions
+	for _, opt := range listOpts {
+		opt(&opts)
```
ferozemohideen, line 76: is this a common pattern? why is this preferred over just sending a struct of opts?

internal/kubernetes/environment_groups/list.go
```diff
@@ -0,0 +1,227 @@
...
+	configMapListResp, err := a.Clientset.CoreV1().ConfigMaps(opts.namespace).List(ctx, listOptions)
+	if err != nil {
+		return nil, telemetry.Error(ctx, span, err, "unable to list environment group variables")
+	}
+	secretListResp, err := a.Clientset.CoreV1().Secrets(opts.namespace).List(ctx, listOptions)
+	if err != nil {
+		return nil, telemetry.Error(ctx, span, err, "unable to list environment groups secret varialbes")
+	}
+
+	allEnvGroupNames := make(map[string]EnvironmentGroup)
```
ferozemohideen, line 108: nit: rename this since it's a map

internal/kubernetes/environment_groups/list.go
```diff
@@ -0,0 +1,227 @@
...
+		}
+		if _, ok := allEnvGroupNames[secret.Name]; !ok {
+			allEnvGroupNames[secret.Name] = EnvironmentGroup{}
+		}
+		allEnvGroupNames[secret.Name] = EnvironmentGroup{
+			Name:            name,
+			Version:         version,
+			SecretVariables: secret.Data,
+			Variables:       allEnvGroupNames[secret.Name].Variables,
+			CreatedAtUTC:    secret.CreationTimestamp.Time.UTC(),
```
ferozemohideen, line 156: 1. bob creates an env group with only secrets
2. 4 years later, bob adds non-secrets to the env group
3. when bob lists the env group, what will be its creation timestamp?

internal/kubernetes/environment_groups/get.go
```diff
@@ -0,0 +1,88 @@
...
+
+import (
+	"context"
+
+	"github.com/porter-dev/porter/internal/kubernetes"
+	"github.com/porter-dev/porter/internal/telemetry"
+)
+
+// LatestBaseEnvironmentGroup returns the most recent version of an environment group stored in the porter-env-group namespace
+func LatestBaseEnvironmentGroup(ctx context.Context, a *kubernetes.Agent, environmentGroupName string) (EnvironmentGroup, error) {
```
ferozemohideen, line 11: nit: GetLatesBaseEnvironmentGroup

internal/kubernetes/environment_groups/list.go
```diff
@@ -0,0 +1,227 @@
...
+}
+
+// LinkedPorterApplication represents an application which was linked to an environment group
+type LinkedPorterApplication struct {
+	Name      string
+	Namespace string
+}
+
+// LinkedApplications lists all applications that are linked to a given environment group. Since there can be multiple linked environment groups we must check by the presence of a label on the deployment and job
+func LinkedApplications(ctx context.Context, a *kubernetes.Agent, environmentGroupName string) ([]LinkedPorterApplication, error) {
```
ferozemohideen, line 175: nit: ListLinkedApplicationsForEnvironmentGroup

api/server/handlers/environment_groups/list.go
```diff
@@ -0,0 +1,124 @@
...
+			if app.Namespace == "" {
+				continue
+			}
+			if _, ok := applicationSetForEnvGroup[app.Namespace]; !ok {
+				applicationSetForEnvGroup[app.Namespace] = struct{}{}
+			}
+		}
+		var linkedApplications []string
+		for appNamespace := range applicationSetForEnvGroup {
+			porterAppName := strings.TrimPrefix(appNamespace, "porter-stack-")
```
ferozemohideen, line 105: I'm confused - isn't the app name included in the LinkedPorterApplication response? why re-construct it here?

api/server/handlers/porter_app/create.go
```diff
@@ -155,13 +155,15 @@ func (c *CreatePorterAppHandler) ServeHTTP(w http.ResponseWriter, r *http.Reques
...
 	chart, values, preDeployJobValues, err := parse(
+		ctx,
 		ParseConf{
 			PorterYaml:                porterYaml,
 			ImageInfo:                 imageInfo,
 			ServerConfig:              c.Config(),
 			ProjectID:                 cluster.ProjectID,
 			UserUpdate:                request.UserUpdate,
 			EnvGroups:                 request.EnvGroups,
+			EnvironmentGroups:         request.EnvironmentGroups,
```
ferozemohideen, line 166: lol

api/server/handlers/porter_app/parse.go
```diff
@@ -168,20 +172,45 @@ func parse(conf ParseConf) (*chart.Chart, map[string]interface{}, map[string]int
...
+					if val == "" {
+						services[serviceName].Config["labels"] = map[string]string{
+							environment_groups.LabelKey_LinkedEnvironmentGroup: strings.Join(conf.EnvironmentGroups, "."),
+						}
+					}
+				}
+			}
+		}
+	}
+
```
ferozemohideen, line 199: nit: can we add this logic to a helper function? I'd like to keep the high-level parse function as concise as possible

api/server/handlers/porter_app/parse.go
```diff
@@ -168,20 +172,45 @@ func parse(conf ParseConf) (*chart.Chart, map[string]interface{}, map[string]int
 		services = parsed.Services
 	}
 
+	for serviceName := range services {
+		if len(conf.EnvironmentGroups) != 0 {
+			if _, ok := services[serviceName].Config["labels"]; !ok {
+				services[serviceName].Config["labels"] = make(map[string]string)
+			}
+			if _, ok := services[serviceName].Config["labels"].(map[string]any); ok {
+				delete(services[serviceName].Config["labels"].(map[string]any), environment_groups.LabelKey_LinkedEnvironmentGroup)
```
ferozemohideen, line 181: i'm assuming this won't panic if the key isn't in the map?

api/server/handlers/porter_app/parse.go
```diff
@@ -788,12 +882,40 @@ func convertHelmValuesToPorterYaml(helmValues string) (*PorterStackYAML, error)
...
+		// 	services[serviceName].Config["labels"] = globalLabels
+		// }
 	}
+
 	return &PorterStackYAML{
 		Services: services,
 	}, nil
 }
+
+func extractGlobalLabels(helmValues string) (map[string]string, error) {
```
ferozemohideen, line 904: remove if unused

api/server/handlers/release/get.go
```diff
@@ -33,16 +34,20 @@ func NewReleaseGetHandler(
 }
 
 func (c *ReleaseGetHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
-	helmRelease, _ := r.Context().Value(types.ReleaseScope).(*release.Release)
+	ctx, span := telemetry.NewSpan(r.Context(), "serve-get-release")
```
ferozemohideen, line 37: thanks for adding telemetry to this function <3

buuuut you know that means you have to add it everywhere in this function <3

internal/kubernetes/environment_groups/create.go
```diff
@@ -0,0 +1,183 @@
...
+		if !k8serror.IsNotFound(err) {
+			return telemetry.Error(ctx, span, err, "unable to check if global environment group exists")
+		}
+
+		_, err = a.Clientset.CoreV1().Namespaces().Create(ctx, &v1.Namespace{ObjectMeta: metav1.ObjectMeta{Name: Namespace_EnvironmentGroups}}, metav1.CreateOptions{})
+		if err != nil {
+			return telemetry.Error(ctx, span, err, "unable to create global environment group")
+		}
+	}
+	telemetry.WithAttributes(span, telemetry.AttributeKV{Key: "environment-group-namespace", Value: Namespace_EnvironmentGroups})
```
ferozemohideen, line 39: nit: would prefer to make the Value here the namespace of the response object directly

internal/kubernetes/environment_groups/delete.go
```diff
@@ -0,0 +1,72 @@
...
+	if name == "" {
+		return telemetry.Error(ctx, span, nil, "environment group name cannot be empty")
+	}
+
+	environmentGroups, err := ListEnvironmentGroups(ctx, a, WithEnvironmentGroupName(name), WithNamespace(Namespace_EnvironmentGroups))
+	if err != nil {
+		return telemetry.Error(ctx, span, err, "unable to list environment groups")
+	}
+
+	for _, environmentGroup := range environmentGroups {
```
ferozemohideen, line 29: you check for linked applications in the base namespace, but then you look through all namespaces when you delete. is there an edgecase where there can be linked applications in non-base namespace?

internal/kubernetes/environment_groups/get.go
```diff
@@ -0,0 +1,88 @@
...
+type EnvironmentGroupInTargetNamespaceInput struct {
+	// Name is the environment group name which can be found on the configmap label
+	Name      string
+	Version   int
+	Namespace string
+}
+
+// EnvironmentGroupInTargetNamespace checks if an environment group of a specific name and version exists in a target namespace.
+// If an environment group exists, it will be returned
+func EnvironmentGroupInTargetNamespace(ctx context.Context, a *kubernetes.Agent, inp EnvironmentGroupInTargetNamespaceInput) (EnvironmentGroup, error) {
```
ferozemohideen, line 49: nit: GetEnvironmentGroupInTargetNamespace

ferozemohideen (changes requested):

api/server/handlers/environment_groups/delete.go
stefanmcshane: environment groups are a product level declaration which use kubernetes primitives under the scenes. We should start delineating based on API layers, business/product layers, and the foundational/3rd party layers that we use. There is no such thing as an environment group in kubernetes, so why is it cleaner to associate it there?

internal/kubernetes/environment_groups/list.go
stefanmcshane: https://dave.cheney.net/2014/10/17/functional-options-for-friendly-apis

api/server/handlers/release/get.go
stefanmcshane: Touche

api/types/porter_app.go
```diff
@@ -48,8 +48,10 @@ type CreatePorterAppRequest struct {
 	ImageInfo        ImageInfo `json:"image_info" form:"omitempty"`
 	OverrideRelease  bool      `json:"override_release"`
 	EnvGroups        []string  `json:"env_groups"`
-	UserUpdate       bool      `json:"user_update"`
-	FullHelmValues   string    `json:"full_helm_values"`
+	// EnvironmentGroups are the list of environment groups that this app is linked to. This should be used instead of EnvGroups.
```
ferozemohideen, line 51: add telemetry to EnvGroups so we can get rid of it in a month or so?

dashboard/src/shared/api.tsx
```diff
@@ -1651,6 +1661,20 @@ const createEnvGroup = baseApi<
   return `/api/projects/${pathParams.id}/clusters/${pathParams.cluster_id}/namespaces/${pathParams.namespace}/envgroup/create`;
 });
 
+const createEnvironmentGroups = baseApi<
```
ferozemohideen, line 1664: *createEnvironmentGroup

dashboard/src/shared/api.tsx
```diff
@@ -1586,6 +1586,16 @@ const upgradeChartValues = baseApi<
   return `/api/projects/${id}/clusters/${cluster_id}/namespaces/${namespace}/releases/${name}/0/upgrade`;
 });
 
+const getAllEnvGroups = baseApi<
```
ferozemohideen, line 1589: we should be consistent about Env vs Environment. I would rename this to getAllEnvironmentGroups

ferozemohideen: or listEnvironmentGroupsForCluster

dashboard/src/shared/api.tsx
```diff
@@ -1651,6 +1661,20 @@ const createEnvGroup = baseApi<
...
 });
 
+const createEnvironmentGroups = baseApi<
+  {
+    name: string;
+    variables: Record<string, string>;
+    secret_variables?: Record<string, string>;
+  },
+  {
+    id: number;
```
ferozemohideen, line 1671: rename to project_id

dashboard/src/shared/api.tsx
```diff
@@ -1760,6 +1784,19 @@ const deleteEnvGroup = baseApi<
   return `/api/projects/${pathParams.id}/clusters/${pathParams.cluster_id}/namespaces/${pathParams.namespace}/envgroup`;
 });
 
+const deleteNewEnvGroup = baseApi<
```
ferozemohideen, line 1787: deleteEnvironmentGroup*

dashboard/src/shared/api.tsx
```diff
@@ -1760,6 +1784,19 @@ const deleteEnvGroup = baseApi<
...
+  {
+    name: string;
+  },
+  {
+    id: number;
```
ferozemohideen, line 1792: project_id

dashboard/src/shared/api.tsx
```diff
@@ -2692,6 +2729,7 @@ export default {
   createGitlabIntegration,
   createEmailVerification,
   createEnvironment,
+  createEnvironmentGroups,
```
ferozemohideen, line 2732: maybe group the create,get and delete environment group endpoints in this list

dashboard/src/shared/api.tsx
ferozemohideen, line 1664: these methods should be next to each other since they're all hitting the same endpoint

dashboard/src/main/home/app-dashboard/expanded-app/ExpandedApp.tsx
```diff
@@ -202,38 +203,33 @@ const ExpandedApp: React.FC<Props> = ({ ...props }) => {
...
-              namespace: envGroup.namespace,
-              version: envGroup.version,
-            }
-          )
-          .then((res) => res.data)
-      );
-      const populatedEnvGroups = await Promise.all(populateEnvGroupsPromises);
-      const filteredEnvGroups = populatedEnvGroups.filter(envGroup => envGroup.applications.includes(newAppData.chart.name));
-      setSyncedEnvGroups(filteredEnvGroups)
+        .then((res) => res?.data?.environment_groups)
```
ferozemohideen, line 216:
```suggestion
        .then((res) => res?.data?.environment_groups) ?? []
```
then envGroups will always be defined

dashboard/src/main/home/app-dashboard/expanded-app/ExpandedApp.tsx
```diff
@@ -202,38 +203,33 @@ const ExpandedApp: React.FC<Props> = ({ ...props }) => {
...
+        .catch((error) => {
+          console.error("Failed to fetch environment groups:", error);
+          return [];
+        });
+      let filteredEnvGroups: NewPopulatedEnvGroup[] = [];
+
+      if (envGroups) {
```
ferozemohideen, line 223: when envGroups is always defined after above change, don't need this check

dashboard/src/main/home/app-dashboard/expanded-app/ExpandedApp.tsx
```diff
@@ -202,38 +203,33 @@ const ExpandedApp: React.FC<Props> = ({ ...props }) => {
...
+        filteredEnvGroups = envGroups?.filter(envGroup =>
+          envGroup?.linked_applications?.length > 0 && envGroup?.linked_applications?.includes(appName)
+        );
+      }
+
+      setSyncedEnvGroups(filteredEnvGroups || []);
```
ferozemohideen, line 229: don't need || [] with the above check

dashboard/src/main/home/app-dashboard/expanded-app/ExpandedApp.tsx
```diff
@@ -306,6 +302,7 @@ const ExpandedApp: React.FC<Props> = ({ ...props }) => {
         }
       }
     } catch (err) {
+      console.log(err)
```
ferozemohideen, line 305: remove

dashboard/src/main/home/app-dashboard/expanded-app/ExpandedApp.tsx
```diff
@@ -480,6 +403,7 @@ const ExpandedApp: React.FC<Props> = ({ ...props }) => {
         );
         const yamlString = yaml.dump(finalPorterYaml);
         const base64Encoded = btoa(yamlString);
+        // console.log(syncedEnvGroups?.map((env) => env.name))
```
ferozemohideen, line 406: remove

dashboard/src/main/home/app-dashboard/expanded-app/ExpandedApp.tsx
```diff
@@ -488,7 +412,8 @@ const ExpandedApp: React.FC<Props> = ({ ...props }) => {
           repo_name: tempPorterApp.repo_name,
           git_branch: tempPorterApp.git_branch,
           buildpacks: "",
-          env_groups: syncedEnvGroups?.map((env) => env.name),
+          //full_helm_values: yaml.dump(values),
```
ferozemohideen, line 415: remove

dashboard/src/main/home/app-dashboard/expanded-app/env-vars/EnvGroupModal.tsx
```diff
@@ -61,40 +59,25 @@ const EnvGroupModal: React.FC<Props> = ({
   const [cloneSuccess, setCloneSuccess] = useState(false);
 
   const updateEnvGroups = async () => {
-    let envGroups: PartialEnvGroup[] = [];
+    let populateEnvGroupsPromises: any[] = [];
     try {
-      envGroups = await api
-        .listEnvGroups<PartialEnvGroup[]>(
+      populateEnvGroupsPromises = await api
```
ferozemohideen, line 64: is this still a group of promises? isn't this just an array now? no need to Promise.await() later

dashboard/src/main/home/app-dashboard/expanded-app/env-vars/EnvGroupModal.tsx
```diff
@@ -354,3 +343,19 @@ const GroupModalSections = styled.div`
   grid-template-columns: 1fr 1fr;
   max-height: 365px;
 `;
+const ColumnContainer = styled.div`
+  display: flex;
+  flex-direction: column;
+  align-items: stretch; /* Ensure child elements stretch to full width */
```
ferozemohideen, line 349: remove these css comments

dashboard/src/main/home/app-dashboard/expanded-app/env-vars/EnvVariablesTab.tsx
```diff
@@ -50,62 +54,44 @@ export const EnvVariablesTab: React.FC<EnvVariablesTabProps> = ({
   }, []);
 
   const updateEnvGroups = async () => {
```
ferozemohideen, line 56: this is the same method as in the modal, can you re-use it?

dashboard/src/main/home/cluster-dashboard/env-groups/ExpandedEnvGroupDashboard.tsx
```diff
@@ -14,6 +14,7 @@ import { useQuery } from "@tanstack/react-query";
 import api from "shared/api";
 import Loading from "components/Loading";
 import Placeholder from "components/Placeholder";
+import { any } from "zod";
```
ferozemohideen, line 17: remove

dashboard/src/main/home/cluster-dashboard/env-groups/ExpandedEnvGroupDashboard.tsx
```diff
@@ -72,7 +84,7 @@ const EnvGroupDashboard = (props: PropsType) => {
     }
 
     const envGroup = envGroups.find((envGroup) => envGroup.name === name);
-
+    console.log(envGroup)
```
ferozemohideen, line 87: remove

dashboard/src/main/home/app-dashboard/expanded-app/ExpandedApp.tsx
```diff
@@ -41,9 +41,9 @@ import ActivityFeed from "./activity-feed/ActivityFeed";
 import MetricsSection from "./MetricsSection";
 import StatusSectionFC from "./status/StatusSection";
 import ExpandedJob from "./expanded-job/ExpandedJob";
-import _ from "lodash";
+import _, { flatMapDepth } from "lodash";
```
ferozemohideen, line 44: remove

dashboard/src/main/home/app-dashboard/expanded-app/ExpandedApp.tsx
```diff
@@ -129,7 +130,7 @@ const ExpandedApp: React.FC<Props> = ({ ...props }) => {
   }
   const eventId = queryParams.get('event_id');
   const selectedTab: ValidTab = tab != null && validTabs.includes(tab) ? tab : DEFAULT_TAB;
-
+  const [values, setValues] = React.useState<string>();
```
ferozemohideen, line 133: remove

dashboard/src/main/home/app-dashboard/expanded-app/ExpandedApp.tsx
```diff
@@ -202,38 +203,33 @@ const ExpandedApp: React.FC<Props> = ({ ...props }) => {
...
+      if (envGroups) {
+        filteredEnvGroups = envGroups?.filter(envGroup =>
+          envGroup?.linked_applications?.length > 0 && envGroup?.linked_applications?.includes(appName)
+        );
+      }
+
+      setSyncedEnvGroups(filteredEnvGroups || []);
       setPorterJson(porterJson);
       setAppData(newAppData);
+      setValues(newAppData?.chart)
```
ferozemohideen, line 232: remove

ferozemohideen (changes requested):

api/server/handlers/environment_groups/list.go
stefanmcshane: Forcefully dropping the porter-stack to ensure that we arent relying on it

api/server/handlers/porter_app/parse.go
stefanmcshane: Its a no-op if the map exists

api/types/porter_app.go
stefanmcshane: Will do in a follow-up when we move to porter.yaml consolidation. Great point though

api/server/handlers/environment_groups/delete.go
ferozemohideen: fair, I was modeling off of other agent-level operations we do but those should probably be refactored into the 3 layers like in ccp

internal/kubernetes/environment_groups/delete.go
stefanmcshane: LinkedApplications returns the linked applications in all namespaces. The logic here is that we list the environmentgroups from the base (source of truth) based on the envgroup name, then we make sure that there are no linked applications.
We then go through all namespaces again and delete the configmaps and the secrets that have the base name, so that we delete all versions

internal/kubernetes/environment_groups/get.go
stefanmcshane: https://groups.google.com/g/golang-nuts/c/6joQwydzNyg

internal/kubernetes/environment_groups/list.go
stefanmcshane: The creation of the new version will be the creation stamp for that version, and that version only. Both the secret and the configmap will have the same timestamp since its set at the start of the request

ferozemohideen (approved):

PR #3328, opened 2023-07-28
Pretty sure env groups don't work for pre-deploy jobs - now they do

ferozemohideen:
## Pull request type

- [ ] Bugfix
- [ ] Feature

## Pull request checklist

### DIFF
--- a/api/server/handlers/porter_app/parse.go
+++ b/api/server/handlers/porter_app/parse.go
@@ -175,28 +175,7 @@ func parse(ctx context.Context, conf ParseConf) (*chart.Chart, map[string]interf
 	}
 
 	for serviceName := range services {
-		if len(conf.EnvironmentGroups) != 0 {
-			if _, ok := services[serviceName].Config["labels"]; !ok {
-				services[serviceName].Config["labels"] = make(map[string]string)
-			}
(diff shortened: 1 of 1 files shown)


sdess09 (approved):

internal/kubernetes/porter_app/create.go
```diff
@@ -0,0 +1,8 @@
+package porter_app
+
+// TODO: migrate all kubernetes-level operations from /api/server/handlers/porter_app/{create,parse}.go to this file
+
+const (
```
ferozemohideen, line 5: @stefanmcshane what do I do here if I don't care about the actual value? I plan on just checking if the key exists in the labels later

stefanmcshane: Instead of setting the label selector as `LabelKey_PorterApplication = MY_VALUE` just set it to `LabelKey_PorterApplication`

stefanmcshane: Example here:
https://github.com/porter-dev/porter/blob/0e690cf37a4b76d8e8a310d405410334b7146935/internal/kubernetes/environment_groups/list.go#L185

Get all environment groups, regardless of their value

PR #3350, opened 2023-08-03
Add GHA step to notify to slack when deploying to prod

ferozemohideen:
## Pull request type
I followed https://github.com/marketplace/actions/slack-notify

- [ ] Bugfix
- [x] Feature

## Pull request checklist


.github/workflows/production.yml
```diff
@@ -4,6 +4,20 @@
       - production
 name: Deploy Porter to Production
 jobs:
+  slack-notification:
```
jose-fully-ported, line 7: I think this should be a `needs` for each of the build steps, and we probably want to add slack job that always runs at the end and notifies of success/failure.

Link [here](https://stackoverflow.com/a/70959885).

jose-fully-ported: This is the pattern: https://github.com/marketplace/actions/slack-notify-build

d-g-town (approved):

PR #3351, opened 2023-08-04
feat: pin node/npm versions and run npm build in CI

jose-fully-ported:
## Pull request type

- [ ] Bugfix
- [ ] Feature
- [x] Other (please describe): Tests

## Pull request checklist

Previously, local development did not pin Node/NPM, resulting in developer environments drifting from what is run in CI.

Additionally, CI never ran `npm run build`, meaning a user's change could not build in the relevant node version but still get merged by virtue of running on a developer's machine.

## What is the new behavior?

We want to ensure our builds always succeed in CI prior to merging changes. This change adds that step to CI.

Additionally, we now pin npm and node in the dashboard's package.json.

N/A

### DIFF
--- a/.github/workflows/pr_push_checks.yaml
+++ b/.github/workflows/pr_push_checks.yaml
@@ -1,6 +1,8 @@
 name: PR Checks
+
 on:
   - pull_request
+
 jobs:
   testing_matrix:
     strategy:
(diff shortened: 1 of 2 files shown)


.github/workflows/pr_push_checks.yaml
```diff
@@ -29,3 +31,37 @@ jobs:
...
+          echo "Bootstrapping npm $version (replacing $npm_version)..."
+          npm install --unsafe-perm -g --quiet "npm@$version"
+
+          # Verify npm works before capturing and ensure its stderr is inspectable later
+          npm --version 2>&1 1>/dev/null
+          echo "npm $(npm --version) installed"
+      - name: Install NPM Dependencies
+        working-directory: dashboard
+        run: |
+          npm i --legacy-peer-deps
```
ianedwards, line 63: Would using `npm ci` also work here?

ianedwards (approved):

.github/workflows/pr_push_checks.yaml
jose-fully-ported: Maybe, but this copies what we do in every other part of the codebase, so I'd rather just follow the pattern.

PR #3352, opened 2023-08-04
Add extra information to deploy events, supplemented by updates from porter-agent

ferozemohideen:
## Pull request type

Currently, when a porter app is deployed, a deploy event with status SUCCESS is sent to the server, which includes only the image tag that was deployed and no additional metadata.

With these changes, a deploy event with status PROGRESSING is sent on application deploy; that event also has metadata on the services that are deployed in that revision. The porter-agent is now responsible for listening to the deployment status that is associated with each service deploy, and updating the event if the deployment reaches a successful or failed state.

This feature is only enabled for users with updated porter-agent (code changes to enable the updating behavior are tracked here: https://github.com/porter-dev/porter-agent/pull/52) and, currrently, only porter.run emails. Once this feature is ready to go, we can drop that secondary dependency.

- [ ] Bugfix
- [x] Feature

## Pull request checklist

image

## What is the new behavior?

image


api/server/handlers/cluster/detect_agent_installed.go
```diff
@@ -32,31 +33,29 @@ func NewDetectAgentInstalledHandler(
...
 	if err != nil {
-		c.HandleAPIError(w, r, apierrors.NewErrInternal(err))
+		err = telemetry.Error(ctx, span, err, "failed to get k8s agent")
+		c.HandleAPIError(w, r, apierrors.NewErrPassThroughToClient(err, http.StatusInternalServerError))
 		return
 	}
 
-	depl, err := agent.GetPorterAgent()
-
+	res, err := GetAgentVersionResponse(agent)
```
jose-fully-ported, line 48: Should we be handling the error here?

api/server/handlers/porter_app/create.go
```diff
@@ -297,7 +301,14 @@ func (c *CreatePorterAppHandler) ServeHTTP(w http.ResponseWriter, r *http.Reques
 			return
 		}
 
-		_, err = createPorterAppEvent(ctx, "SUCCESS", porterApp.ID, 1, imageInfo.Tag, c.Repo().PorterAppEvent())
+		// Only create the PROGRESSING event if the cluster's agent is updated, because only the updated agent can update the status
+		// TODO: remove dependence on porter email once we are ready to release this feature
+		if isPorterAgentUpdated(k8sAgent, 3, 1, 6) && strings.HasSuffix(user.Email, "porter.run") {
```
jose-fully-ported, line 306: Can we change:

```
strings.HasSuffix(user.Email, "porter.run")
```

to

```
isPorterEmail(user.Email)
```

So we don't re-implement that logic everywhere?

api/server/handlers/porter_app/create.go
```diff
@@ -203,7 +207,7 @@ func (c *CreatePorterAppHandler) ServeHTTP(w http.ResponseWriter, r *http.Reques
 		// create the release job chart if it does not exist (only done by front-end currently, where we set overrideRelease=true)
 		if request.OverrideRelease && preDeployJobValues != nil {
 			telemetry.WithAttributes(span, telemetry.AttributeKV{Key: "installing-pre-deploy-job", Value: true})
-			conf, err := createReleaseJobChart(
+			conf, err := createPreDeployJobChart(
```
jose-fully-ported, line 210: Why this change (same question below)?

api/server/handlers/porter_app/create.go
jose-fully-ported: We might also want to wrap this if logic in a helper like:

```
areProgressEventsEnabled()
```

Or something. That centralizes the logic and makes it so we can enable it more widely later on without trying to determine if a particular `if` guard should be updated.

I'd create a `features.go` to centralize these sorts of functions :)

api/server/handlers/porter_app/create.go
```diff
@@ -519,7 +541,117 @@ func createPorterAppEvent(ctx context.Context, status string, appID uint, revisi
...
+	if revision <= 1 {
+		return
+	}
+	revisionFloat64 := float64(revision - 1)
+	matchEvent, err := repo.ReadDeployEventByRevision(ctx, appID, revisionFloat64)
+	if err != nil {
+		_ = telemetry.Error(ctx, span, err, "error reading deploy event by revision")
+		return
+	}
+	if matchEvent.ID == uuid.Nil {
```
jose-fully-ported, line 605: I'm not familiar with `uuid.Nil`, why not just compare to `nil`?

api/server/handlers/porter_app/create.go
```diff
@@ -297,7 +301,14 @@ func (c *CreatePorterAppHandler) ServeHTTP(w http.ResponseWriter, r *http.Reques
 			return
 		}
 
-		_, err = createPorterAppEvent(ctx, "SUCCESS", porterApp.ID, 1, imageInfo.Tag, c.Repo().PorterAppEvent())
+		// Only create the PROGRESSING event if the cluster's agent is updated, because only the updated agent can update the status
+		// TODO: remove dependence on porter email once we are ready to release this feature
+		if isPorterAgentUpdated(k8sAgent, 3, 1, 6) && strings.HasSuffix(user.Email, "porter.run") {
+			serviceDeploymentStatusMap := getServiceDeploymentMetadataFromValues(values, "PROGRESSING")
```
jose-fully-ported, line 307: Non-blocking, but the status `PROGRESSING` and any other similar statuses (like `SUCCESS`) should probably be a `const`.

api/server/handlers/porter_app/create.go
```diff
@@ -606,3 +738,41 @@ func cloneEnvGroup(c *CreatePorterAppHandler, w http.ResponseWriter, r *http.Req
...
+	}
+
+	parsedTag := strings.Split(tag, ".")
+	if len(parsedTag) != 3 {
+		return false
+	}
+
+	parsedMajor, _ := strconv.Atoi(parsedTag[0])
+	parsedMinor, _ := strconv.Atoi(parsedTag[1])
+	parsedPatch, _ := strconv.Atoi(parsedTag[2])
```
jose-fully-ported, line 767: Will the patch version ever have a tag or something? Like `1+beta3`?

api/server/handlers/porter_app/parse.go
```diff
@@ -941,3 +941,49 @@ func addLabelsToService(service *Service, envGroups []string, defaultLabelKey st
...
+						}
+					}
+				}
+
+				if porterHosts, ok := ingressMap["porter_hosts"].([]interface{}); ok && len(porterHosts) == 1 {
+					return porterHosts[0].(string)
+				}
+			}
+		}
+	}
```
jose-fully-ported, line 986: Non-blocking, but nested if statements like this are... well not great to read :D. I wonder if there is a cleaner way to do this other than exiting early.

api/server/handlers/cluster/detect_agent_installed.go
ferozemohideen: oh I'm not sure if I added this after you commented but we handle the error right below

api/server/handlers/porter_app/create.go
ferozemohideen: good call! made the file and added the logic there

api/server/handlers/porter_app/create.go
ferozemohideen: ah well when I initially wrote this code, I made it so that the internal name for a pre-deploy job was a 'release' job, modeled after Heroku's 'release' keyword for similar functionality. I only realized later that the word 'release' is pretty overloaded in the repo, referring to Helm releases - this is my attempt to correct the ambiguity I created :/

api/server/handlers/porter_app/create.go
ferozemohideen: I believe this is a gorm intricacy, I get a compile error when I try to compare to `nil` because of mismatched types. I think with the gorm find method we're using, it won't return an error if the event is not found, but rather an empty row with uuid.Nil 🤷

api/server/handlers/porter_app/create.go
ferozemohideen: good point, created a type for the status and made each one a const!

api/server/handlers/porter_app/create.go
ferozemohideen: Hmm currently we only release agent versions that have numerical patch versions, but I see what you're saying here. Ideally we can remove this dependency entirely once we go about upgrading everyone's agent versions (I added a note about monitoring telemetry and removing this once we see 100% adoption of the new version)

api/server/handlers/cluster/detect_agent_installed.go
jose-fully-ported: Ah gotcha, the `targetErr` thing confused me.

api/server/handlers/cluster/detect_agent_installed.go
```diff
@@ -32,31 +33,29 @@ func NewDetectAgentInstalledHandler(
...
 	if targetErr := kubernetes.IsNotFoundError; err != nil && errors.Is(err, targetErr) {
```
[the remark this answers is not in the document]
ferozemohideen: haha yep that's clearer, fixed

dashboard/src/components/porter/Link.tsx
```diff
@@ -50,28 +54,46 @@ const Svg = styled.svg`
...
   height: 1px;
   width: 100%;
-  background: #ffffff;
+  background: ${(props) => props.color ?? "#ffffff"};
 `;
 
-const LinkWrapper = styled.span`
-  position: relative;
+const StyledLink = styled(DynamicLink) <{ hasunderline?: boolean, color?: string }>`
+  color: ${(props) => props.color ?? "#ffffff"};
```
jose-fully-ported, line 67: React noob here. Is there a way to specify the defaults in the arguments vs inline in code?

dashboard/src/components/porter/Link.tsx
```diff
@@ -50,28 +54,46 @@ const Svg = styled.svg`
...
-const StyledLink = styled(DynamicLink)<{ hasunderline?: boolean }>`
-  color: #ffffff;
+const LinkWrapper = styled.span<{ hoverColor?: string, color?: string }>`
+  position: relative;
   display: inline-flex;
-  font-size: 13px;
-  cursor: pointer;
+  align-items: center;
+  :hover {
+    ${StyledLink} {
```
jose-fully-ported, line 87: React noob: What does this `${Name}` thing do?

dashboard/src/main/home/app-dashboard/expanded-app/activity-feed/events/cards/ServiceStatusDetail.tsx
```diff
@@ -0,0 +1,126 @@
...
+            case "SUCCESS":
+                return "DEPLOYED";
+            case "FAILED":
+                return "FAILED";
+            case "CANCELED":
+                return "CANCELED";
+            default:
+                return "UNKNOWN";
+        }
+    };
```
jose-fully-ported, line 35: For future reference, we'll want to ensure these are properly set in the api in the future vs translated here in the frontend.

dashboard/src/components/porter/Link.tsx
ferozemohideen: good call, specified the default in the component props

dashboard/src/components/porter/Link.tsx
ferozemohideen: that applies the style to every child styled component. I tried adding the color and background-color changes there without specifying the child component, but for some reason they weren't propagating to the children :/ this was the only way that worked

i am also very much a css noob lol

jose-fully-ported (approved):

PR #3353, opened 2023-08-04
feat: convert yaml to proto type for validation

ianedwards:
## Pull request type

- [ ] Bugfix
- [ ] Feature

## Pull request checklist


api/server/handlers/porter_app/validation/porter_yaml.go
```diff
@@ -0,0 +1,229 @@
+package validation
```
d-g-town, line 1: Mainly naming things:

I would reorder the methods from higher-level to lower-level:
1. ConvertAppsFromYaml
2. getAppsFromApplicationGroup
3. castServiceConfig
4. getType

Advice from Stefan: no method names should start with get - just remove it. Maybe also rename 

I would also rename this package from validation to conversion, and rename the method AppsFromYaml (so that the method call is conversion.AppsFromYaml).

Other naming suggestions: maybe actually call the main method AppProtoFromYaml, getType -> protoEnumFromType, castServiceConfig -> serviceProtoFromConfig

api/server/handlers/porter_app/validation/types.go
```diff
@@ -0,0 +1,56 @@
+package validation
```
d-g-town, line 1: I would throw all this types in the same file as the methods... doesn't seem like there so much code that they can't stay together for now

api/server/handlers/porter_app/validation/types.go
```diff
@@ -0,0 +1,56 @@
...
+	Name         string                  `yaml:"name" validate:"required_without=Applications"`
+	Version      *string                 `yaml:"version"`
+	Image        *Image                  `yaml:"image"`
+	Build        *Build                  `yaml:"build"`
+	Env          map[string]string       `yaml:"env"`
+	SyncedEnv    []*SyncedEnvSection     `yaml:"synced_env"`
+	Apps         map[string]*Service      `yaml:"apps" validate:"required_without=Applications Services"`
+	Services     map[string]*Service      `yaml:"services" validate:"required_without=Applications Apps"`
+
+	Release *Service `yaml:"release"`
```
d-g-town, line 14: Does these all have to be pointers? I imagine we should only use pointers if we want to signify that a struct should be nil (but for maps at least that is taken care of by the key being present or not).

d-g-town (approved):

api/server/handlers/porter_app/validation/porter_yaml.go
```diff
@@ -0,0 +1,229 @@
...
+	}
+
+	if strings.Contains(name, "wkr") {
+		return porterv1.ServiceType_SERVICE_TYPE_WORKER, nil
+	}
+
+	return porterv1.ServiceType_SERVICE_TYPE_JOB, nil
+}
+
+func ConvertAppsFromYaml(file []byte) (map[string]*porterv1.PorterApp, error) {
```
d-g-town, line 47: add a comment for what this does

PR #3371, opened 2023-08-14
feat: add support for exposing nginx response codes via metrics endpoint

jose-fully-ported:
## Pull request type

- [x] Feature

## Pull request checklist

This is new functionality being introduced, so there is no "current" behavior.

Issue Number: POR-1474

## What is the new behavior?

The metrics endpoint now supports an 'nginx:status' metric with a required nginx_status_level (int 1-5) querystring argument and 'kind=Ingress'. This will return nginx_ingress_controller_requests scoped to the specified app, which can then be graphed in the frontend.

Example api endpoint (note that date ranges should be updated as appropriate):

```
/api/projects/1/clusters/1/metrics?metric=nginx:status&nginx_status_level=2&shouldsum=false&kind=Ingress&name=nginx-my-app-web-web&namespace=porter-stack-nginx&startrange=1692038386&endrange=1692041986&resolution=1s
```

I also simulated traffic using the following [`k6`](https://k6.io/docs/get-started/running-k6/) script against my sample app:

```js
import http from 'k6/http';
import { check, sleep } from 'k6';
export const options = {
  stages: [
    { duration: '30s', target: 20 },
    { duration: '1m30s', target: 10 },
    { duration: '20s', target: 0 },
    { duration: '3m10s', target: 20 },
    { duration: '5m', target: 10 },
  ],
};

export default function () {
  for (let id = 1; id <= 100; id++) {
    const res = http.get(`https://app.example.com/id-${id}`);
    check(res, { 'status was 200': (r) => r.status == 200 });
    sleep(1);
  }
}
```

### DIFF
--- a/internal/kubernetes/prometheus/metrics.go
+++ b/internal/kubernetes/prometheus/metrics.go
@@ -109,17 +109,21 @@ func GetIngressesWithNGINXAnnotation(clientset kubernetes.Interface) ([]SimpleIn
 }
 
 type QueryOpts struct {
-	Metric           string   `schema:"metric"`
-	ShouldSum        bool     `schema:"shouldsum"`
-	Kind             string   `schema:"kind"`
-	PodList          []string `schema:"pods"`
-	Name             string   `schema:"name"`
-	Namespace        string   `schema:"namespace"`
-	NginxStatusLevel uint     `schema:"nginx_status_level"`
-	StartRange       uint     `schema:"startrange"`
-	EndRange         uint     `schema:"endrange"`
-	Resolution       string   `schema:"resolution"`
-	Percentile       float64  `schema:"percentile"`
+	// the name of the metric being queried for
+	Metric    string   `schema:"metric"`
+	ShouldSum bool     `schema:"shouldsum"`
+	Kind      string   `schema:"kind"`
+	PodList   []string `schema:"pods"`
+	Name      string   `schema:"name"`
+	Namespace string   `schema:"namespace"`
+	// a prefix [1,2,3,4,5] used to scope nginx requests by when querying for status code responses
+	NginxStatusLevel uint `schema:"nginx_status_level"`
+	// start time (in unix timestamp) for prometheus results
+	StartRange uint `schema:"startrange"`
+	// end time time (in unix timestamp) for prometheus results
+	EndRange   uint    `schema:"endrange"`
+	Resolution string  `schema:"resolution"`
+	Percentile float64 `schema:"percentile"`
 }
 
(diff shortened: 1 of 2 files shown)


internal/kubernetes/prometheus/metrics_test.go
```diff
@@ -0,0 +1,84 @@
...
+
+	var tests = []struct {
+		name     string
+		input    input
+		expected struct {
+			query string
+			err   error
+		}
+	}{
+		// the table itself
```
stefanmcshane, line 28: Is this comment needed?

internal/kubernetes/prometheus/metrics_test.go
```diff
@@ -0,0 +1,84 @@
...
+			output{
+				`round(sum by (ingress)(irate(nginx_ingress_controller_requests{exported_namespace=~"app-namespace",ingress="process-app-web",service="process-app-web",status=~"2.."}[5m])), 0.001)`,
+				nil,
+			},
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			query, err := getNginxStatusQuery(tt.input.opts, tt.input.selectionRegex)
+			if query != tt.expected.query {
```
stefanmcshane, line 72: Any reason not to handle/check error conditions before checking the response types?

internal/kubernetes/prometheus/metrics.go
```diff
@@ -108,16 +109,17 @@ func GetIngressesWithNGINXAnnotation(clientset kubernetes.Interface) ([]SimpleIn
...
-	ShouldSum  bool     `schema:"shouldsum"`
-	Kind       string   `schema:"kind"`
-	PodList    []string `schema:"pods"`
-	Name       string   `schema:"name"`
-	Namespace  string   `schema:"namespace"`
-	StartRange uint     `schema:"startrange"`
-	EndRange   uint     `schema:"endrange"`
-	Resolution string   `schema:"resolution"`
-	Percentile float64  `schema:"percentile"`
+	Metric           string   `schema:"metric"`
```
stefanmcshane, line 112: It seems that none of these exported fields had comments before this change. Can you add some comments to the parts that you added at the very least?

stefanmcshane (approved):
Big fan of k6 for testing. Left a few comments, but none are blockers

internal/kubernetes/prometheus/metrics_test.go
jose-fully-ported: None tbh. I'll reverse the check.

PR #3376, opened 2023-08-14
check if validate apply v2 is enabled before running commands involving application deployments and releases

d-g-town:


ianedwards (approved):
Out of curiosity - were you able to determine that this is the exhaustive set of commands that'll be impacted?

d-g-town:
> Out of curiosity - were you able to determine that this is the exhaustive set of commands that'll be impacted?

As best I can tell this is exhaustive: all the CLI commands use a helper function `checkLoginAndRun`, so I went through all the instances of this and checked whether the function works with application releases or not (if so, we will have to use new logic to do new deployments/get revision history).

stefanmcshane (approved):
Consider using the logger, instead of fmt.PrintX when writing outputs for a user

d-g-town:
> Consider using the logger, instead of fmt.PrintX when writing outputs for a user

Looks like most of the existing CLI commands use color.New().  I'll look into refactoring all the print statements into a unified logger to pass around in coming PR.

PR #3378, opened 2023-08-15
types and conversions for porter app def

ianedwards:
## Pull request type

- [ ] Bugfix
- [ ] Feature

## Pull request checklist


dashboard/src/lib/porter-apps/services.ts
```diff
@@ -0,0 +1,335 @@
...
+  autoscalingValidator,
+  healthcheckValidator,
+  deserializeAutoscaling,
+  deserializeHealthCheck,
+  serializeAutoscaling,
+  serializeHealth,
+} from "./values";
+import { Service, ServiceType } from "@porter-dev/api-contracts";
+
+export const serviceStringValidator = z.object({
```
d-g-town, line 15: worth adding comments to all these functions saying something like "this object allows us to block users from changing service values that are present in the porter.yaml"

dashboard/src/lib/porter-apps/services.ts
```diff
@@ -0,0 +1,335 @@
...
+          cron: config.cron.value,
+        },
+      })
+    )
+    .exhaustive();
+}
+
+export function deserializeService(
+  service: SerializedService,
+  existing?: SerializedService
```
d-g-town, line 174: maybe rename existing to override? If I'm understanding correctly, this implements overrides for the Porter yaml? So deserialize service would be called with `deserializeService(frontEndChanges, porterYaml)`?

dashboard/src/lib/porter-apps/index.ts
```diff
@@ -0,0 +1,112 @@
...
+
+export const porterAppFormValidator = z.object({
+  app: porterAppValidator,
+  source: sourceValidator,
+});
+export type PorterAppFormData = z.infer<typeof porterAppFormValidator>;
+
+export function porterClientAppFromProto(
+  proto: PorterApp,
+  buildpacks?: Buildpack[]
```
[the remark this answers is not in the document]
ianedwards: I might be able to remove this, but for now the problem is that the proto will only give the buildpack names, but I want to be able to use the full buildpack object we get on detection. Don't really want to have to make the detectBuildpacks API call in this function.

If we don't need all the information that detectBuildpacks gives us when calling apply, then I'll remove it.

d-g-town:
Generally looks good, but would be really helpful to start commenting the front-end code like we do the backend. Maybe to start, could you add comments to all the fleshier methods explaining what they are used for?  And distinguishing between a serialized vs service service? 

Would be very helpful for someone like me who has less familiarity with typescript to get a handle on what the logic should be accomplishing.

PR #3387, opened 2023-08-16
Use average data correctly on app metric graphs

jose-fully-ported:
## Pull request type

- [x] Bugfix

## Pull request checklist

Previously, when showing "all" data for a given metric, we would show the summed version of the data on the chart. This is confusing to users, as they would assume that the value is the average utilization of whatever the metric was.

## What is the new behavior?

Now, for "all" metrics, we show the average instead and remove the old average data line (with min/max still shown separately). Pod-specific charts are still treated normally.

Aggregated (All) view

Screenshot 2023-08-16 at 1 00 26 PM

Pod-specific view

Screenshot 2023-08-16 at 1 00 44 PM

### DIFF
--- a/dashboard/src/main/home/app-dashboard/expanded-app/MetricsSection.tsx
+++ b/dashboard/src/main/home/app-dashboard/expanded-app/MetricsSection.tsx
@@ -10,7 +10,7 @@ import { ChartTypeWithExtendedConfig, StorageType } from "shared/types";
 import TabSelector from "components/TabSelector";
 import Loading from "components/Loading";
 import SelectRow from "components/form-components/SelectRow";
-import AreaChart from "../../cluster-dashboard/expanded-chart/metrics/AreaChart";
+import AreaChart from "./metrics/AreaChart";
 import { MetricNormalizer } from "../../cluster-dashboard/expanded-chart/metrics/MetricNormalizer";
 import {
   AvailableMetrics,
@@ -54,6 +54,7 @@ const MetricsSection: React.FunctionComponent<PropsType> = ({
   );
   const [dropdownExpanded, setDropdownExpanded] = useState(false);
   const [data, setData] = useState<NormalizedMetricsData[]>([]);
+  const [isAggregated, setIsAggregated] = useState<boolean>(false);
   const [aggregatedData, setAggregatedData] = useState<
     Record<string, NormalizedMetricsData[]>
   >({});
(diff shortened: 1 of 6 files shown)


dashboard/src/main/home/cluster-dashboard/expanded-chart/metrics/AggregatedDataLegend.tsx
```diff
@@ -7,18 +7,27 @@ import { AggregatedDataColors } from "./utils";
...
+      min,
+      max,
+    }
+  } else {
+    aggregatedData = {
+      min,
+      avg,
+      max,
+    }
+  }
```
ianedwards, line 30: Will aggregated data have a type here?

dashboard/src/main/home/app-dashboard/expanded-app/MetricsSection.tsx
```diff
@@ -325,59 +327,65 @@ const MetricsSection: React.FunctionComponent<PropsType> = ({
       const allPodsMetrics = allPodsData.flatMap((d) => d.results);
       const allPodsMetricsNormalized = new MetricNormalizer(
         [{ results: allPodsMetrics }],
-        selectedMetric as AvailableMetrics
+        selectedMetric as AvailableMetrics,
       );
-      setAggregatedData(allPodsMetricsNormalized.getAggregatedData());
-      //
+      const allPodsAggregatedData = allPodsMetricsNormalized.getAggregatedData()
+      let data: NormalizedMetricsData[] = []
```
ianedwards, line 333: is this variable being used?

dashboard/src/main/home/app-dashboard/expanded-app/MetricsSection.tsx
```diff
@@ -325,59 +327,65 @@ const MetricsSection: React.FunctionComponent<PropsType> = ({
...
+        if (shouldsum && isHpaEnabled) {
+          if (selectedMetric === "cpu") {
+            await getAutoscalingThreshold(
+              "cpu_hpa_threshold",
+              shouldsum,
+              namespace,
+              start,
+              end
+            );
+          } else if (selectedMetric === "memory") {
```
ianedwards, line 371: In cases where we switch on these string values, it'd be nice to ensure that the type on the useState reflects those options so that we can match on them exhaustively using a switch statement or a library like ts-pattern:

`const [selectedMetric, setSelectedMetric] = useState<"cpu" | "memory">("cpu");`

ianedwards (approved):

PR #3388, opened 2023-08-16
scaffold create app form and use ff

ianedwards:
## Pull request type

- [ ] Bugfix
- [ ] Feature

## Pull request checklist


dashboard/src/main/home/app-dashboard/create-app/CreateApp.tsx
```diff
@@ -0,0 +1,213 @@
...
+                        source={source}
+                        projectId={currentProject.id}
+                      />
+                    ) : (
+                      <div></div>
+                    )
+                  ) : null} */}
+                </>,
+                <>
+                  <ErrorText hasError={true}>
```
d-g-town, line 143: nit: I don't love the look of this error text

dashboard/src/main/home/app-dashboard/create-app/CreateApp.tsx
```diff
@@ -0,0 +1,213 @@
...
+          <Spacer y={3} />
+        </StyledConfigureTemplate>
+      </Div>
+    </CenterWrapper>
+  );
+};
+
+export default withRouter(CreateApp);
+
+const ErrorText = styled.span`
```
d-g-town, line 170: Can/should we reuse any of these objects from the other app creation flow?

dashboard/src/components/porter/ControlledInput.tsx
```diff
@@ -0,0 +1,150 @@
+import React from "react";
+import styled from "styled-components";
+import Tooltip from "./Tooltip";
+
+export const ControlledInput = React.forwardRef<
+  HTMLInputElement,
+  {
+    id: string;
+    name: string;
+    label?: string;
```
d-g-town, line 10: perhaps add a comment to any potentially undefined fields describing what happens if they are/aren't included

d-g-town (approved):

dashboard/src/main/home/app-dashboard/create-app/CreateApp.tsx
ianedwards: I'll be removing this in my next PR, but will update it if we need to bring it back before releasing

dashboard/src/main/home/app-dashboard/create-app/CreateApp.tsx
ianedwards: My expectation is that we'll be deleting the other file at a certain point, so I think it might be better to have separation of concerns in this case

PR #3391, opened 2023-08-17
apply validate yaml

d-g-town:
- wip
- validate and apply API endpoints, call in cli

### DIFF
--- a/api/server/handlers/porter_app/apply.go
+++ b/api/server/handlers/porter_app/apply.go
@@ -25,7 +25,7 @@ type ApplyPorterAppHandler struct {
 	handlers.PorterHandlerReadWriter
 }
 
-// NewApplyPorterAppHandler returns a new ApplyPorterAppHandler
+// NewApplyPorterAppHandler handles POST requests to the endpoint /apps/apply
 func NewApplyPorterAppHandler(
 	config *config.Config,
 	decoderValidator shared.RequestDecoderValidator,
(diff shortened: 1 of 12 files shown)


api/server/router/porter_app.go
```diff
@@ -566,10 +566,97 @@ func getPorterAppRoutes(
...
+			Scopes: []types.PermissionScope{
+				types.UserScope,
+				types.ProjectScope,
+				types.ClusterScope,
+			},
+		},
+	)
+
+	defaultDeploymentTargetHandler := porter_app.NewDefaultDeploymentTargetHandler(
+		config,
```
ianedwards, line 650: I'm curious why the default deployment target handler is scoped to an application rather than just under a cluster.

Won't there be one default depl target per cluster and an app could theoretically be deployed to any of them?

api/server/handlers/porter_app/validate.go
```diff
@@ -0,0 +1,154 @@
...
+		PorterHandlerReadWriter: handlers.NewDefaultPorterHandler(config, decoderValidator, writer),
+	}
+}
+
+// ValidatePorterAppRequest is the request object for the /apps/validate endpoint
+type ValidatePorterAppRequest struct {
+	Base64AppProto     string `json:"b64_app_proto"`
+	DeploymentTargetId string `json:"deployment_target_id"`
+	CommitSHA          string `json:"commit_sha"`
+}
```
ianedwards, line 44: Does it potentially make sense to get the latest commit sha based on the porter_app with the same name as is found in the proto?

Worried about a client having to get the latest commit sha. We could also handle that in another endpoint if that's easier.

api/server/handlers/porter_app/apply.go
```diff
@@ -0,0 +1,146 @@
...
+		err := telemetry.Error(ctx, span, nil, "error decoding request")
+		c.HandleAPIError(w, r, apierrors.NewErrPassThroughToClient(err, http.StatusBadRequest))
+		return
+	}
+
+	if request.Base64AppProto == "" {
+		err := telemetry.Error(ctx, span, nil, "b64 yaml is empty")
+		c.HandleAPIError(w, r, apierrors.NewErrPassThroughToClient(err, http.StatusBadRequest))
+		return
+	}
```
ianedwards, line 81: Just confirming that the logic where will eventually be that either Base64AppProto is not empty or revision ID is not empty

api/server/handlers/porter_app/default_deployment_target.go
```diff
@@ -0,0 +1,82 @@
...
+
+// DefaultDeploymentTargetResponse is the response object for the /apps/default-deployment-target endpoint
+type DefaultDeploymentTargetResponse struct {
+	DeploymentTargetID string `json:"deployment_target_id"`
+}
+
+const (
+	// DeploymentTargetSelector_Default is the selector for the default deployment target in a cluster
+	DeploymentTargetSelector_Default = "default"
+	// DeploymentTargetSelectorType_Default is the selector type for the default deployment target in a cluster
```
ianedwards, line 45: Not sure if this was discussed before, but if I wanted to delete all the apps in an entire environment, can I not do that if we're using `default` as the default selector

internal/repository/gorm/deployment_target.go
```diff
@@ -0,0 +1,36 @@
...
+func NewDeploymentTargetRepository(db *gorm.DB) repository.DeploymentTargetRepository {
+	return &DeploymentTargetRepository{db}
+}
+
+// DeploymentTargetBySelectorAndSelectorType finds a deployment target for a projectID and clusterID by its selector and selector type
+func (repo *DeploymentTargetRepository) DeploymentTargetBySelectorAndSelectorType(projectID uint, clusterID uint, selector, selectorType string) (*models.DeploymentTarget, error) {
+	deploymentTarget := &models.DeploymentTarget{}
+
+	if err := repo.db.Where("project_id = ? AND cluster_id = ? AND selector = ? AND selector_type = ?", projectID, clusterID, selector, selectorType).Limit(1).Find(&deploymentTarget).Error; err != nil {
+		return nil, err
```
ianedwards, line 28: Curious if there's some benefit to using `Limt(1).Find` vs `.First`?

I know with the first option, it just returns a 0 value if not found, and the second one returns an error. I'm just wondering if we should prefer one to the other.

api/server/handlers/porter_app/apply.go
d-g-town: yes exactly, that will be the logic when I start implementing build

api/server/handlers/porter_app/default_deployment_target.go
d-g-town: after in person discussion, we will not rely on the namespace to delete all apps

api/server/router/porter_app.go
d-g-town: I'll change this, it wasn't a good reason other than keeping it all under the same `/apps` scope

api/server/handlers/porter_app/validate.go
d-g-town: I think this will only have to be sent by the CLI being run in github CI, so we should always be able to get the commit sha at that point

jose-fully-ported (approved):
Looks good!

PR #3392, opened 2023-08-17
updated build settings form on app creation

ianedwards:
## Pull request type

- [ ] Bugfix
- [ ] Feature

## Pull request checklist

### DIFF
--- a/dashboard/src/lib/porter-apps/services.ts
+++ b/dashboard/src/lib/porter-apps/services.ts
@@ -9,6 +9,7 @@ import {
   deserializeHealthCheck,
   serializeAutoscaling,
   serializeHealth,
+  domainsValidator,
 } from "./values";
 import { Service, ServiceType } from "@porter-dev/api-contracts";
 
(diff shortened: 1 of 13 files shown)


sdess09 (approved):

dashboard/src/lib/porter-apps/services.ts
```diff
@@ -75,17 +79,14 @@ export const serviceValidator = z.object({
...
       type: z.literal("web"),
-      autoscaling: autoscalingValidator.optional(),
-      domains: z.array(
-        z.object({
-          name: serviceStringValidator,
-        })
-      ),
-      healthCheck: healthcheckValidator.optional(),
+      autoscaling: autoscalingValidator,
+      ingressEnabled: z.boolean().default(false).optional(),
```
d-g-town, line 83: does default(false) mean optional() never matters?

d-g-town (approved):

dashboard/src/lib/porter-apps/services.ts
ianedwards: It might not be fully necessary, but it just makes sure that when we submit the form that we're working with a boolean rather than undefined

PR #3395, opened 2023-08-18
enable advanced settings at cluster provisioning for non-simplified view users; add kms and guard duty to advanced settings

d-g-town:


sdess09 (approved):

dashboard/src/components/ProvisionerSettings.tsx
```diff
@@ -831,14 +831,76 @@ const ProvisionerSettings: React.FC<Props> = (props) => {
                       <Spacer y={1} />
                     </>
                   )}
+                  <FlexCenter>
+                    <>
```
ianedwards, line 835: Could remove this `<> ` tag here.

The reason this is used in some places is that you can only return 1 top-level JSX element from a function

ianedwards (approved):

PR #3402, opened 2023-08-21
load in service overrides from porter yaml

ianedwards:
## Pull request type

- [ ] Bugfix
- [ ] Feature

## Pull request checklist

### DIFF
--- a/dashboard/src/lib/porter-apps/index.ts
+++ b/dashboard/src/lib/porter-apps/index.ts
@@ -1,10 +1,8 @@
-import {
-  BUILDPACK_TO_NAME,
-  Buildpack,
-  buildpackSchema,
-} from "main/home/app-dashboard/types/buildpack";
+import { buildpackSchema } from "main/home/app-dashboard/types/buildpack";
 import { z } from "zod";
 import {
+  ClientService,
+  defaultSerialized,
   deserializeService,
   serializedServiceFromProto,
   serviceValidator,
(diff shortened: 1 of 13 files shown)


sdess09 (approved):

dashboard/src/lib/porter-apps/index.ts
```diff
@@ -64,57 +62,43 @@ export const porterAppFormValidator = z.object({
...
-): ClientPorterApp {
-  const services = Object.entries(proto.services).map(([name, service]) =>
-    deserializeService(serializedServiceFromProto({ name, service }))
-  );
-
-  const { name, env, build, predeploy, image } = proto;
+export function defaultServicesWithOverrides({
+  overrides,
+}: {
+  overrides?: PorterApp;
```
ferozemohideen, line 68: rather than null checking in this method, would it be cleaner to pass a default overrides object with empty services/pre-deploy as a function parameter?

dashboard/src/lib/porter-apps/services.ts
```diff
@@ -127,6 +79,63 @@ export type SerializedService = {
...
+  name: string;
+  type: "web" | "worker" | "job";
+}): SerializedService {
+  const baseService = {
+    name,
+    run: "",
+    instances: 1,
+    port: 3000,
+    cpuCores: 0.1,
+    ramMegabytes: 256,
```
ferozemohideen, line 95: are we switching from mebibytes?

dashboard/src/lib/porter-apps/services.ts
```diff
@@ -127,6 +79,63 @@ export type SerializedService = {
...
 };
 
+export function defaultSerialized({
+  name,
+  type,
+}: {
+  name: string;
+  type: "web" | "worker" | "job";
+}): SerializedService {
+  const baseService = {
```
ferozemohideen, line 89: these defaults should be UPPER_CASE imo

dashboard/src/lib/porter-apps/services.ts
```diff
@@ -127,6 +79,63 @@ export type SerializedService = {
...
+        type: "worker" as const,
+        autoscaling: defaultAutoscaling,
+      },
+    }))
+    .with("job", () => ({
+      ...baseService,
+      config: {
+        type: "job" as const,
+        allowConcurrent: false,
+        cron: "",
```
ferozemohideen, line 133: just wanna make sure paused: true is assigned somewhere, probably when it's validated/hydrated?

dashboard/src/lib/porter-apps/services.ts
```diff
@@ -142,10 +151,14 @@ export function serializeService(service: ClientService): SerializedService {
         ramMegabytes: service.ramMegabytes.value,
         config: {
           type: "web" as const,
-          autoscaling: serializeAutoscaling({
-            autoscaling: config.autoscaling,
-          }),
-          healthCheck: serializeHealth({ health: config.healthCheck }),
+          autoscaling: config.autoscaling
+            ? serializeAutoscaling({
```
ferozemohideen, line 155: this might be cleaner if you're able to return undefined in the `serialize` functions but up to you

dashboard/src/lib/porter-apps/services.ts
ianedwards: Yes, we decided MB would be the clearest unit to work with in user facing contexts

dashboard/src/lib/porter-apps/index.ts
ianedwards: Honestly I think this method is only going to be called with overrides, so it might make most sense to just make overrides required.

ferozemohideen (approved):

dashboard/src/lib/porter-apps/services.ts
ianedwards: I think I'm gonna keep the values close to how they're defined in porter.yaml for now, but will change if it ends up becoming difficult to read

PR #3404, opened 2023-08-22
handle predeploy with new syntax

ianedwards:
## Pull request type

- [ ] Bugfix
- [ ] Feature

## Pull request checklist

### DIFF
--- a/dashboard/src/lib/porter-apps/services.ts
+++ b/dashboard/src/lib/porter-apps/services.ts
@@ -17,6 +17,8 @@ import {
 } from "./values";
 import { Service, ServiceType } from "@porter-dev/api-contracts";
 
+type ClientServiceType = "web" | "worker" | "job" | "predeploy";
+
 // serviceValidator is the validator for a ClientService
 // This is used to validate a service when creating or updating an app
 export const serviceValidator = z.object({
@@ -85,12 +87,16 @@ export type SerializedService = {
       };
 };
 
+export function isPredeployService(service: SerializedService | ClientService) {
+  return service.config.type == "predeploy";
+}
+
 export function defaultSerialized({
   name,
   type,
 }: {
   name: string;
-  type: "web" | "worker" | "job" | "predeploy";
+  type: ClientServiceType;
 }): SerializedService {
   const baseService = {
     name,
@@ -309,9 +315,7 @@ export function deserializeService(
 }
 
 // getServiceTypeEnumProto converts the type of a ClientService to the protobuf ServiceType enum
-export const serviceTypeEnumProto = (
-  type: "web" | "worker" | "job" | "predeploy"
-): ServiceType => {
+export const serviceTypeEnumProto = (type: ClientServiceType): ServiceType => {
   return match(type)
     .with("web", () => ServiceType.WEB)
     .with("worker", () => ServiceType.WORKER)
--- a/dashboard/src/lib/porter-apps/services.ts
+++ b/dashboard/src/lib/porter-apps/services.ts
@@ -45,6 +45,9 @@ export const serviceValidator = z.object({
       allowConcurrent: serviceBooleanValidator,
       cron: serviceStringValidator,
     }),
+    z.object({
+      type: z.literal("predeploy"),
+    }),
   ]),
 });
 
@@ -76,6 +79,9 @@ export type SerializedService = {
         type: "job";
         allowConcurrent: boolean;
         cron: string;
+      }
+    | {
+        type: "predeploy";
       };
 };
 
@@ -84,7 +90,7 @@ export function defaultSerialized({
   type,
 }: {
   name: string;
-  type: "web" | "worker" | "job";
+  type: "web" | "worker" | "job" | "predeploy";
 }): SerializedService {
   const baseService = {
     name,
@@ -131,6 +137,12 @@ export function defaultSerialized({
         type: "job" as const,
         allowConcurrent: false,
         cron: "",
+      },
+    }))
+    .with("predeploy", () => ({
+      ...baseService,
+      config: {
+        type: "predeploy" as const,
       },
     }))
     .exhaustive();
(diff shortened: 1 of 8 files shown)


dashboard/src/lib/porter-apps/services.ts
```diff
@@ -84,7 +90,7 @@ export function defaultSerialized({
   type,
 }: {
   name: string;
-  type: "web" | "worker" | "job";
+  type: "web" | "worker" | "job" | "predeploy";
```
ferozemohideen, line 93: is this the only place you define the union type? personally i would pull out this union type into its own `ServiceType` or something so that the type attribute is always `ServiceType`

dashboard/src/lib/porter-apps/services.ts
```diff
@@ -375,7 +421,7 @@ export function serializedServiceFromProto({
       ...service,
       name,
       config: {
-        type: "job" as const,
+        type: isPredeploy ? ("predeploy" as const) : ("job" as const),
```
ferozemohideen, line 424: oof this is a lil clunky but idk a good way to get around it

dashboard/src/main/home/app-dashboard/create-app/CreateApp.tsx
```diff
@@ -137,21 +143,21 @@ const CreateApp: React.FC<CreateAppProps> = ({}) => {
...
+
         const { services, predeploy } = defaultServicesWithOverrides({
           overrides: proto,
         });
 
         if (services.length) {
-          setValue("app.services", services);
+          const defaultServices = predeploy
+            ? [...services, predeploy]
+            : services;
```
ferozemohideen, line 154:
```suggestion
          const defaultServices = [...services, predeploy].filter(x => x != null);
```

dashboard/src/main/home/app-dashboard/create-app/CreateApp.tsx
```diff
@@ -168,14 +174,14 @@ const CreateApp: React.FC<CreateAppProps> = ({}) => {
     // set step to 2 if source is filled out
     if (source?.type && source.type === "github") {
       if (source.git_repo_name && source.git_branch) {
-        setStep((prev) => Math.max(prev, 3));
+        setStep((prev) => Math.max(prev, 5));
```
ferozemohideen, line 177: something for future work but we should convert the vertical steps to use string keys of a map rather than indices imo, makes it less ambiguous when we update the steps like this

also just curious do you prefer functions rather than values in setters? why not `setStep(Math.max(step,5))`

dashboard/src/main/home/app-dashboard/validate-apply/app-settings/EnvVariables.tsx
```diff
@@ -0,0 +1,45 @@
...
+const EnvVariables: React.FC = () => {
+  const { control } = useFormContext<PorterAppFormData>();
+
+  const recordToKVType = useCallback((env?: Record<string, string>) => {
+    return env
+      ? Object.entries(env).map(([key, value]) => {
+          return { key, value, hidden: false, locked: false, deleted: false };
+        })
+      : [];
+  }, []);
```
ferozemohideen, line 18:
```suggestion
    return Object.entries(env ?? []).map(([key, value]) => {
          return { key, value, hidden: false, locked: false, deleted: false };
        });
  }, []);
```

dashboard/src/main/home/app-dashboard/validate-apply/services-settings/ServiceList.tsx
```diff
@@ -120,6 +122,14 @@ const ServiceList: React.FC<ServiceListProps> = ({
...
         <ServicesContainer>
           {services.map((service, idx) => {
+            if (isPredeploy && service.config.type !== "predeploy") {
+              return null;
+            }
+
+            if (!isPredeploy && service.config.type === "predeploy") {
+              return null;
+            }
+
```
ferozemohideen, line 132: what is going on here lmao

anyway isn't it cleaner to use a typeguard, maybe defined in services.ts?

typeguard could be:
```
const isNotPredeploy = (service: Service): service is NotPredeployService => {
            return (isPredeploy && service.config.type !== "predeploy") 
( idk what the exact condition in here is)
```

```suggestion
          {services.filter(isNotPredeploy).map((service, idx) => {
         ... core logic
```
though introducing extra types might be annoying

ferozemohideen (approved):
approving with style nits but looks good!

dashboard/src/main/home/app-dashboard/create-app/CreateApp.tsx
ianedwards: Yeah, definitely. The steps can be confusing as it's currently implemented.

As far as using previous state goes, the updater function ensure that the update is based on the correct value, as state updates are queued.

See here: https://react.dev/reference/react/useState#updating-state-based-on-the-previous-state

dashboard/src/lib/porter-apps/services.ts
ianedwards: Should probably pull it out into a type just to be safe.

dashboard/src/main/home/app-dashboard/create-app/CreateApp.tsx
ianedwards: This is a brutal oversight in TS, but filter actually doesn't guarantee that the return type is `T`, and instead it'll still be `T | undefined`, so we'll get type errors.

dashboard/src/main/home/app-dashboard/validate-apply/services-settings/ServiceList.tsx
ianedwards: Yeah this definitely isn't ideal. One annoying thing with the `react-hook-form` library in this case is that the `useFieldArray` hooks depends on the index of the services in form data to make updates, so if I prefilter it'll end up editing the incorrect services.

I removed the null returns from the component level and just added a flag that checks for predeploy services beforehand.

PR #3405, opened 2023-08-22
add build to new apply

d-g-town:


Tiltfile
```diff
@@ -60,7 +60,6 @@ local_resource(
   deps=[
     "api",
     "build",
-    "cli",
```
d-g-town, line 63: This stops unnecessary rebuilds of app for cli changes

api/client/porter_app.go
```diff
@@ -246,3 +248,28 @@ func (c *Client) DefaultDeploymentTarget(
...
+	}
+
+	err := c.getRequest(
+		fmt.Sprintf(
+			"/projects/%d/clusters/%d/apps/current-app-revision",
+			projectID, clusterID,
+		),
+		req,
+		resp,
+	)
```
ianedwards, line 272: I'm wondering if these routes should be scoped to an app name and then we can leave open the option for getting different revisions off a common base path.

Something like as follows:
`/projects/%d/clusters/%d/apps/%s/revisions/latest` <--- get latest revision
`/projects/%d/clusters/%d/apps/%s/revisions/%d` <--- get revision by offset from latest or id

Thoughts?

internal/repository/gorm/porter_app.go
```diff
@@ -44,6 +44,18 @@ func (repo *PorterAppRepository) ReadPorterAppByName(clusterID uint, name string
 	return app, nil
 }
 
+// ReadPorterAppsByProjectIDAndName returns a list of PorterApps by project ID and name. Multiple apps can have the same name and project id
+// if they are in different clusters.
```
ianedwards, line 48: Correct me if I'm wrong, but I thought we decided that (project_id, app_name) had to be unique?

ianedwards: Then a single application could be shared across multiple deployment targets

api/server/handlers/porter_app/current_app_revision.go
```diff
@@ -0,0 +1,149 @@
...
+		return
+	}
+
+	currentAppRevisionReq := connect.NewRequest(&porterv1.CurrentAppRevisionRequest{
+		ProjectId:          int64(project.ID),
+		PorterAppId:        int64(porterApps[0].ID),
+		DeploymentTargetId: request.DeploymentTargetID,
+	})
+
+	currentAppRevisionResp, err := c.Config().ClusterControlPlaneClient.CurrentAppRevision(ctx, currentAppRevisionReq)
```
ianedwards, line 116: Noticing that the `CurrentAppRevision` rpc on CCP gets the most recent deployed revision, but on the frontend I could imagine that we'd want to get the latest revision regardless of status.

Would it be possible to pass in a status parameter to handle that?

ianedwards (approved):
Nice! Leaving some comments for consideration, lmk what you think

api/server/handlers/porter_app/current_app_revision.go
d-g-town: This makes a lot of sense.  I unfortunately already released the api-contracts, but definitely think the status should be added to the call when we want to use it for other statuses.

internal/repository/gorm/porter_app.go
d-g-town: Yes, but it is possible in the current porter_apps DB for there to be multiple project_id/app_names (because it is unique between app_name, cluster_id atm). I just check for this case and fail out if there are multiple

api/client/porter_app.go
d-g-town: I'm a fan.  Will change the latest endpoint now, and then look at doing locking on the ccp database access to let us increment by deployment target id.

PR #3414, opened 2023-08-23
Remove dependency on cli to fix build

ferozemohideen:
## Pull request type

- [ ] Bugfix
- [ ] Feature

## Pull request checklist


jose-fully-ported (approved):
Do we not build the app in CI for PRs?

ferozemohideen:
hmm we do not but I'm not sure why. Previously builds would take up to 10 mins I believe but now that we build the node code and go code in parallel I think builds are only like 3-5 mins

PR #3416, opened 2023-08-23
submit form and run validate + apply

ianedwards:
## Pull request type

- [ ] Bugfix
- [ ] Feature

## Pull request checklist


api/server/handlers/porter_app/create_app.go
```diff
@@ -107,15 +107,15 @@ func (c *CreateAppHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	}
 	telemetry.WithAttributes(span, telemetry.AttributeKV{Key: "app-name", Value: request.Name})
 
-	if request.SourceType == "" {
+	if request.Type == "" {
```
ferozemohideen, line 110: can this be solved with the decode and validate function? if we add "required" to the key in the struct? I'm not too sure how it works but I think stefan does

dashboard/src/lib/porter-apps/index.ts
```diff
@@ -116,27 +116,40 @@ export function clientAppToProto(app: ClientPorterApp): PorterApp {
...
+          build: {
+            context: app.build.context,
+            method: app.build.method,
+            buildpacks: app.build.buildpacks.map((b) => b.buildpack),
+            builder: app.build.builder,
+            dockerfile: app.build.dockerfile,
+          },
+          ...(predeploy && {
+            predeploy: serviceProto(serializeService(predeploy)),
+          }),
```
ferozemohideen, line 136: is this null safe? does having an empty/undefined predeploy key break things? 

personally I prefer a final object with determinate keys with default/empty values rather than one with nondeterminate keys 🤷 
```suggestion
            predeploy: serviceProto(serializeService(predeploy ?? {})),
```

sdess09 (commented):
LGTM

ferozemohideen (approved):

api/server/handlers/porter_app/create_app.go
ianedwards: Yeah good point, I can probably just map type in json to SourceType

dashboard/src/lib/porter-apps/index.ts
ianedwards: In this case, undefined is going to be allowed regardless since the generated types from protobuf assume all values are optional.

We also assume undefined/nil is meaningful on the CCP side of things. If it's nil, then we don't include it. With that being said, is there a default that would work in your mind?

PR #3420, opened 2023-08-23
fix: do not show throughput chart when not referencing a web service

jose-fully-ported:
## Pull request type

- [x] Bugfix

## Pull request checklist

Issue Number: POR-1474

Currently, we show the throughput chart even when not displaying an app without ingress.

## What is the new behavior?

Hide the throughput chart when there is no ingress (and reduce references to the currentChart data).


jose-fully-ported:
Closing as @ferozemohideen has a better version in the works.

PR #3422, opened 2023-08-23
move validation to hook and add commit sha

ianedwards:
## Pull request type

- [ ] Bugfix
- [x] Feature

## Pull request checklist

- Right now app validation is occurring within `CreateApp` and doesn't take into account the commit sha.
- The goal with this PR is to have a reusable function that we can call on to get a fully validated / hydrated app on the frontend, and include the commit in the request.

Issue Number: N/A

-->

## What is the new behavior?

- The behavior change here is that the tag that will be set on the validated app's image is correctly equal to the hash of the commit sha and the build settings.
- Otherwise behavior should be the same now, although implementation has slightly changed

- Using a hook for reusable validation function call


dashboard/src/main/home/app-dashboard/app-view/AppView.tsx
```diff
@@ -0,0 +1,54 @@
...
+  // "events",
+  "overview",
+  // "logs",
+  // "metrics",
+  // "debug",
+  "environment",
+  "build-settings",
+  "settings",
+  // "helm-values",
+  // "job-history",
```
jose-fully-ported: Should we just remove the commented out values? Or was this a mistake?

jose-fully-ported (approved):

PR #3424, opened 2023-08-23
Refactor metrics, add loading states, add global refresh

ferozemohideen:
## Pull request type

https://github.com/porter-dev/porter/assets/40551216/61db3140-e12f-4669-874e-fd206eeb0959

- [x] Bugfix
- [x] Feature

## Pull request checklist

### DIFF
--- a/dashboard/src/main/home/app-dashboard/expanded-app/metrics/MetricsSection.tsx
+++ b/dashboard/src/main/home/app-dashboard/expanded-app/metrics/MetricsSection.tsx
@@ -3,30 +3,22 @@ import styled from "styled-components";
 
 import api from "shared/api";
 import { Context } from "shared/Context";
-import { ChartTypeWithExtendedConfig } from "shared/types";
+import { ChartType } from "shared/types";
 
 import TabSelector from "components/TabSelector";
 import SelectRow from "components/form-components/SelectRow";
+import { getServiceNameFromControllerName, MetricNormalizer, resolutions, secondsBeforeNow } from "./utils";
+import { Metric, MetricType, NginxStatusMetric } from "./types";
+import { match } from "ts-pattern";
+import { AvailableMetrics, NormalizedMetricsData } from "main/home/cluster-dashboard/expanded-chart/metrics/types";
 import MetricsChart from "./MetricsChart";
-import { getServiceNameFromControllerName } from "./utils";
+import { useQuery } from "@tanstack/react-query";
+import Loading from "components/Loading";
+
 type PropsType = {
-  currentChart: ChartTypeWithExtendedConfig;
+  currentChart: ChartType;
   appName: string;
   serviceName?: string;
-};
-
-export const resolutions: { [range: string]: string } = {
-  "1H": "1s",
-  "6H": "15s",
-  "1D": "15s",
-  "1M": "5h",
-};
-
-export const secondsBeforeNow: { [range: string]: number } = {
-  "1H": 60 * 60,
-  "6H": 60 * 60 * 6,
-  "1D": 60 * 60 * 24,
-  "1M": 60 * 60 * 24 * 30,
 };
 
 const MetricsSection: React.FunctionComponent<PropsType> = ({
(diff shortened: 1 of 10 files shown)


dashboard/src/main/home/app-dashboard/expanded-app/metrics/MetricsSection.tsx
```diff
@@ -90,126 +56,208 @@ const MetricsSection: React.FunctionComponent<PropsType> = ({
...
+  }, [controllerOptions]);
+
+  const renderMetrics = () => {
+    if (metricsData == null) {
+      return <Loading />;
+    }
+    return metricsData.map((metric: Metric, i: number) => {
+      return (
+        <MetricsChart
+          key={i}
```
ianedwards, line 236: Just a heads up - using index as keys can cause some unexpected behavior if the elements change frequently. Not sure it'll end up causing an issue here

dashboard/src/main/home/app-dashboard/expanded-app/metrics/MetricsSection.tsx
```diff
@@ -90,126 +56,208 @@ const MetricsSection: React.FunctionComponent<PropsType> = ({
...
+      selectedController?.metadata?.name,
+      selectedRange,
+    ],
+    async () => {
+      if (currentProject?.id == null || currentCluster?.id == null) {
+        return;
       }
-      i += 1;
-    }
+      const metrics = [] as Metric[];
```
ianedwards, line 81: any reason to use `as` instead of just a regular type def here:

`const metrics: Metric[] = [];`

dashboard/src/main/home/app-dashboard/expanded-app/metrics/MetricsSection.tsx
```diff
@@ -90,126 +56,208 @@ const MetricsSection: React.FunctionComponent<PropsType> = ({
...
+            pods: [],
+          },
+          {
+            id: currentProject.id,
+            cluster_id: currentCluster.id,
+          }
+        );
+        const metricsNormalizer = new MetricNormalizer(
+          [{ results: (aggregatedMetricsResponse.data ?? []).flatMap((d: any) => d.results) }],
+          metricType,
```
ianedwards, line 126: Is this response too complex to type? Could use a zod parseAsync to define the schema of the data and avoid any types

jose-fully-ported (approved):

dashboard/src/main/home/app-dashboard/expanded-app/metrics/MetricsSection.tsx
```diff
@@ -90,126 +56,208 @@ const MetricsSection: React.FunctionComponent<PropsType> = ({
...
-          namespace: selectedController?.metadata?.namespace,
-          selectors,
-        },
-        {
-          id: currentProject.id,
-          cluster_id: currentCluster.id,
+      if (currentChart?.config?.[serviceName]?.ingress?.enabled) {
+        metricTypes.push("nginx:status")
+      }
+
```
ianedwards, line 98: One thing I'd maybe suggest high level is to move metricTypes to a useMemo or something, and then change this to a [useQueries](https://tanstack.com/query/v4/docs/react/reference/useQueries) hook just to cut down on the logic inside the callback here and make this more purely about the result of the network request. Could define a useCallback outside of the actual hook that takes in some of these values as args.

Up to you though, haven't wrapped my head fully around the actual functionality here.

ianedwards (approved):
LGTM overall!

dashboard/src/main/home/app-dashboard/expanded-app/metrics/MetricsSection.tsx
ferozemohideen: good point, I haven't dug into this network call in the backend at all but yes would be nice to define a type for this, gonna leave that as todo

dashboard/src/main/home/app-dashboard/expanded-app/metrics/MetricsSection.tsx
ferozemohideen: I'm curious about how you'd implement that, I'll follow up with you on specifics later. Thanks for the advice tho

jose-fully-ported (approved):

PR #3430, opened 2023-08-24
fix: properly query for the cpu/memory autoscaling thresholds

jose-fully-ported:
## Pull request type

- [x] Bugfix
- [ ] Feature

## Pull request checklist

Issue Number: 1474

Previously, we would show the sum of all autoscaling thresholds for each pod, resulting in a wildly incorrect graph.

## What is the new behavior?

Fixing the sum issue, we still had our metrics downscaled by half, so we need to decrease the divisor by half in order to get the correct values.

The query itself might be incorrect/simplifiable, but at least for now things work as expected.

Screenshot 2023-08-24 at 1 12 11 PM


jose-fully-ported:
Note that the increases there are me manually changing the autoscaling values for testing.

ferozemohideen (approved):

PR #3431, opened 2023-08-24
convert between int and strings for instance counts

d-g-town:
This fixes a bug in the provisioning flow.  Previously, any edits to the max/min instance counts would cause the CreateContract api call to fail because the instance count variable was passed in as a string.  This fix properly types the instance input component and translates to int.


jose-fully-ported (approved):
<3

PR #3432, opened 2023-08-24
scaffold and ff expanded app using revisions

ianedwards:
## Pull request type

- [ ] Bugfix
- [x] Feature

## Pull request checklist

- Current implementation lives in `<ExpandedApp />` and any child components
- Many of the underlying components on this page depend on helm chart values

## What is the new behavior?

- Goal is to start porting over functionality of `<ExpandedApp />`  but with data coming from app revisions rather than helm
- this change just creates a new component sitting behind a FF

Screenshot 2023-08-24 at 2 05 34 PM

### DIFF
--- a/dashboard/src/main/home/app-dashboard/app-view/AppView.tsx
+++ b/dashboard/src/main/home/app-dashboard/app-view/AppView.tsx
@@ -108,6 +108,9 @@ const AppView: React.FC<Props> = ({ match }) => {
 
       const porterApp = await porterAppValidator.parseAsync(res.data);
       return porterApp;
+    },
+    {
+      enabled: appParamsExist,
     }
   );
 
--- a/dashboard/src/main/home/app-dashboard/app-view/AppView.tsx
+++ b/dashboard/src/main/home/app-dashboard/app-view/AppView.tsx
@@ -1,0 +1,283 @@
+import React, { useContext } from "react";
+import { useMemo } from "react";
+import { RouteComponentProps, withRouter } from "react-router";
+import { useQuery } from "@tanstack/react-query";
+import { z } from "zod";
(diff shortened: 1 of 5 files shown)


dashboard/src/main/home/app-dashboard/app-view/AppView.tsx
```diff
@@ -0,0 +1,286 @@
...
+    !!currentCluster &&
+    !!currentProject &&
+    !!deploymentTarget;
+
+  const { data: appData, status: porterAppStatus } = useQuery(
+    ["getPorterApp", currentCluster?.id, currentProject?.id, params.appName],
+    async () => {
+      if (!appParamsExist) {
+        return;
+      }
```
ferozemohideen, line 97: do you need this if the enabled has the same condition?

dashboard/src/shared/api.tsx
```diff
@@ -856,6 +856,19 @@ const applyApp = baseApi<
...
+const getLatestRevision = baseApi<
+  {
+    deployment_target_id: string;
+  },
+  {
+    project_id: number;
+    cluster_id: number;
+    porter_app_name: string;
+  }
+>("GET", (pathParams) => {
```
ferozemohideen, line 868: nit 
```suggestion
>("GET", ({project_id, cluster_id, porter_app_name}) => {
```

ferozemohideen (approved):

dashboard/src/main/home/app-dashboard/app-view/AppView.tsx
ianedwards: yes unfortunately, I need to look into a better solution for ths.

PR #3441, opened 2023-08-25
add create subdomain handler and add subdomains to proto on cli before applying

d-g-town:


api/server/handlers/porter_app/create_subdomain.go
```diff
@@ -0,0 +1,153 @@
...
+	return &CreateSubdomainHandler{
+		PorterHandlerReadWriter: handlers.NewDefaultPorterHandler(config, decoderValidator, writer),
+		KubernetesAgentGetter:   authz.NewOutOfClusterAgentGetter(config),
+	}
+}
+
+// CreateSubdomainRequest is the request object for the /apps/{porter_app_name}/subdomain endpoint
+type CreateSubdomainRequest struct {
+	ServiceName string `schema:"service_name"`
+}
```
ianedwards, line 40: Up to you but I think we also have the `requestutils.GetURLParamString(r, types.URLParamPorterAppName)` helper function that can be used in place of this

api/server/handlers/porter_app/create_subdomain.go
```diff
@@ -0,0 +1,153 @@
...
+	}
+
+	if request.ServiceName == "" {
+		err := telemetry.Error(ctx, span, nil, "service name cannot be empty")
+		c.HandleAPIError(w, r, apierrors.NewErrPassThroughToClient(err, http.StatusBadRequest))
+		return
+	}
+	telemetry.WithAttributes(span, telemetry.AttributeKV{Key: "service-name", Value: request.ServiceName})
+
+	agent, err := c.GetAgent(r, cluster, "")
```
ianedwards, line 77: Is `agent` a standard name we normally use instead of like `k8sAgent`

cli/cmd/v2/apply.go
```diff
@@ -107,6 +112,58 @@ func Apply(ctx context.Context, cliConf config.CLIConfig, client api.Client, por
...
+					return editedB64AppProto, errors.New("response subdomain is empty")
+				}
+
+				webConfig.Domains = []*porterv1.Domain{
+					{Name: domain.Subdomain},
+				}
+
+				service.Config = &porterv1.Service_WebConfig{WebConfig: webConfig}
+			}
+		}
```
ianedwards, line 154: Just want to call out that we will presumably need a way to remove subdomains when we get to deletions and remove domains if private is set to false.

ianedwards (approved):

api/server/handlers/porter_app/create_subdomain.go
d-g-town: yeah, I wasn't sure whether I wanted to add the service name as an additional path parameter, decided against it for now unless we start adding more api calls at the service level

api/server/handlers/porter_app/create_subdomain.go
d-g-town: i'll change this to k8sAgent

cli/cmd/v2/apply.go
d-g-town: I think we don't need to worry about deletions because we should treat the list of domains as the full list. Good point about the private being false, maybe we do that in validate?

PR #3442, opened 2023-08-25
create porter app before running apply in cli

d-g-town:
We must create a porter app db entry before running apply on the CLI.

This also supports a new type of porter app that is built locally (and so is not connected with a github repo).

### DIFF
--- a/cli/cmd/v2/apply.go
+++ b/cli/cmd/v2/apply.go
@@ -7,6 +7,7 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
+	"strconv"
 
 	"github.com/fatih/color"
 	"github.com/porter-dev/api-contracts/generated/go/helpers"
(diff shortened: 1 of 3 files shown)


cli/cmd/v2/apply.go
```diff
@@ -107,6 +118,50 @@ func Apply(ctx context.Context, cliConf config.CLIConfig, client api.Client, por
...
+
+	if app.Build != nil {
+		if os.Getenv("GITHUB_REPOSITORY_ID") == "" {
+			input.Local = true
+			return input, nil
+		}
+		gitRepoId, err := strconv.Atoi(os.Getenv("GITHUB_REPOSITORY_ID"))
+		if err != nil {
+			return input, fmt.Errorf("unable to parse GITHUB_REPOSITORY_ID to int: %w", err)
+		}
```
ianedwards, line 158: Where does GITHUB_REPOSITORY_ID get set?

ianedwards (approved):
lgtm!

cli/cmd/v2/apply.go
d-g-town: It will be in the environment of a workflow: https://docs.github.com/en/actions/learn-github-actions/variables

PR #3443, opened 2023-08-25
first pass at new revisions list

ianedwards:
## Pull request type

- [ ] Bugfix
- [ ] Feature

## Pull request checklist

- We currently have a component called `PorterAppRevisionSection` which is based on a legacy component that reads in releases from helm
- This component also depends on a webhook for listening for new revisions from helm

## What is the new behavior?

- This change reuses the styles from the old component while using data from the app_revisions table.
- No data is read in from helm and the webhook is removed
- To handle refreshes, the parent component `AppDataContainer` polls for the latest revision and will update the revision table anytime a number revision no comes in

Screenshot 2023-08-25 at 3 36 27 PM

### DIFF
--- a/dashboard/src/main/home/app-dashboard/app-view/AppView.tsx
+++ b/dashboard/src/main/home/app-dashboard/app-view/AppView.tsx
@@ -1,26 +1,30 @@
-import React, { useContext } from "react";
-import { useMemo } from "react";
+import React, { useContext, useMemo } from "react";
 import { RouteComponentProps, withRouter } from "react-router";
 import { useQuery } from "@tanstack/react-query";
 import { z } from "zod";
-import Loading from "components/Loading";
-import api from "shared/api";
-import { Context } from "shared/Context";
-import { useDefaultDeploymentTarget } from "lib/hooks/useDeploymentTarget";
 import { PorterApp } from "@porter-dev/api-contracts";
 import styled from "styled-components";
+
+import { Context } from "shared/Context";
+import api from "shared/api";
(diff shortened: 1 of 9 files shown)


dashboard/src/main/home/app-dashboard/app-view/AppView.tsx
```diff
@@ -135,20 +140,17 @@ const AppView: React.FC<Props> = ({ match }) => {
...
-      const porterApp = PorterApp.fromJsonString(
-        atob(rawAppData.b64_app_proto)
-      );
-
-      return porterApp;
+      return revisionData.app_revision;
     },
     {
       enabled: appParamsExist,
+      refetchInterval: 5000,
```
ferozemohideen, line 153: one nice thing about the webhook currently is when a user clicks 'Update App', the new revision comes in immediately and the history is updated. Polling with too long an interval might make porter seem less 'snappy' but of course we can play around with the interval. Just some food for thought

ferozemohideen (approved):

dashboard/src/main/home/app-dashboard/app-view/AppView.tsx
ianedwards: I could see one option here being that we use optimistic updates, such that on app update a new revision is appended to the existing list and then a full update can be done following the poll interval.

[See here](https://tanstack.com/query/v5/docs/react/guides/optimistic-updates#updating-a-list-of-todos-when-adding-a-new-todo)

PR #3444, opened 2023-08-25
get commit in cli

d-g-town:


cli/cmd/v2/apply.go
```diff
@@ -57,6 +66,11 @@ func Apply(ctx context.Context, cliConf config.CLIConfig, client api.Client, por
 	}
 	base64AppProto := validateResp.ValidatedBase64AppProto
 
+	err = errorIfMissingSHAForBuild(base64AppProto, commitSHA)
+	if err != nil {
+		return err
+	}
```
ianedwards, line 72: Curious why this error wouldn't just be sent as part of the response from validate? Wouldn't we be able to check for this in `mergeImageAndBuild`?

ianedwards (approved):

cli/cmd/v2/apply.go
d-g-town: I think validate can't actually determine when things should fail.

If I just want to run porter apply without changing the build artifact. Even if there are build settings then we shouldn't error on the fact that commitSHA wasn't passed in.

I'm actually going to change this logic so that I only fail out if the CLI is prompted to build.  In that case, if we don't have a specified SHA I'm going to fail out (for building an image with the wrong SHA attached)

PR #3446, opened 2023-08-28
app view services and overrides

ianedwards:
## Pull request type

- [ ] Bugfix
- [X] Feature

## Pull request checklist

- Current functionality exists in the ExpandedApp 

## What is the new behavior?

- This PR gets porter yaml v2 to parity on the "Overview" tab. The goal is to display services read in from the latest revision and mark values as read only if defined in porter.yaml
- Note: currently missing a functional StatusFooter, which will be addressed in a followup PR

Screenshot 2023-08-28 at 11 30 44 AM

### DIFF
--- a/dashboard/src/lib/porter-apps/index.ts
+++ b/dashboard/src/lib/porter-apps/index.ts
@@ -69,45 +69,65 @@ export const porterAppFormValidator = z.object({
 });
 export type PorterAppFormData = z.infer<typeof porterAppFormValidator>;
 
-// defaultServicesWithOverrides is used to generate the default services for an app from porter.yaml
+// serviceOverrides is used to generate the services overrides for an app from porter.yaml
 // this method is only called when a porter.yaml is present and has services defined
-export function defaultServicesWithOverrides({
+export function serviceOverrides({
   overrides,
+  useDefaults = true,
 }: {
   overrides: PorterApp;
+  useDefaults?: boolean;
 }): {
   services: ClientService[];
   predeploy?: ClientService;
 } {
   const services = Object.entries(overrides.services)
     .map(([name, service]) => serializedServiceFromProto({ name, service }))
-    .map((svc) =>
-      deserializeService(
-        defaultSerialized({
-          name: svc.name,
-          type: svc.config.type,
-        }),
-        svc
-      )
-    );
+    .map((svc) => {
+      if (useDefaults) {
+        return deserializeService({
+          service: defaultSerialized({ name: svc.name, type: svc.config.type }),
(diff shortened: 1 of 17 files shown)


dashboard/src/lib/porter-apps/index.ts
```diff
@@ -69,45 +69,65 @@ export const porterAppFormValidator = z.object({
...
+export function serviceOverrides({
   overrides,
+  useDefaults = true,
 }: {
   overrides: PorterApp;
+  useDefaults?: boolean;
 }): {
   services: ClientService[];
   predeploy?: ClientService;
 } {
```
ferozemohideen, line 83: is this type DetectedServices?

dashboard/src/lib/porter-apps/index.ts
```diff
@@ -219,19 +239,53 @@ const clientBuildFromProto = (proto?: Build): BuildOptions | undefined => {
...
 };
 
-export function clientAppFromProto(proto: PorterApp): ClientPorterApp {
+export function clientAppFromProto(
+  proto: PorterApp,
+  overrides: {
+    services: ClientService[];
+    predeploy?: ClientService;
+  } | null
+): ClientPorterApp {
```
ferozemohideen, line 248: DetectedServices | null ?

ferozemohideen: or if its not technically 'Detected' maybe we should introduce a type for ```{
    services: ClientService[];
    predeploy?: ClientService;
  }``` since I seem to see it in multiple places

dashboard/src/main/home/app-dashboard/app-view/LatestRevisionContext.tsx
```diff
@@ -0,0 +1,215 @@
...
+      clusterId: number;
+      projectId: number;
+      deploymentTargetId: string;
+    }
+  | undefined
+>(undefined);
+
+export const useLatestRevision = () => {
+  const context = useContext(LatestRevisionContext);
+  if (context === undefined) {
```
ferozemohideen, line 40: nit: i prefer == null here

ferozemohideen (approved):

dashboard/src/lib/porter-apps/index.ts
ianedwards: yeah good catch, I should be reusing the type

dashboard/src/lib/porter-apps/index.ts
ianedwards: addressed above

PR #3451, opened 2023-08-28
Update CLI to use feature flags, instead of project calls

stefanmcshane:
## Issue ID

N/A

## What does this PR do?

- Removed all project-gets, and start passing around feature flags where necessary
- Evaluate flags passed in by CLI as Viper was skipping them due to a lack of global initialisation. This is an intermediary solution for reading flag values at runtime
- Remove token from porter config on failed auth login with `--token` flag

### DIFF
--- a/api/client/api.go
+++ b/api/client/api.go
@@ -10,13 +10,11 @@ import (
 	"net/http"
 	"net/url"
 	"os"
-	"path/filepath"
 	"strings"
 	"time"
 
 	"github.com/gorilla/schema"
 	"github.com/porter-dev/porter/api/types"
-	"k8s.io/client-go/util/homedir"
 )
 
 // Client represents the client for the Porter API
@@ -81,33 +79,6 @@ func NewClientWithConfig(ctx context.Context, input NewClientInput) (Client, err
 
 // ErrNoAuthCredential returns an error when no auth credentials have been provided such as cookies or tokens
 var ErrNoAuthCredential = errors.New("unable to create an API session with cookie nor token")
-
-// NewClient constructs a new client based on a set of options
-func NewClient(baseURL string, cookieFileName string) *Client {
-	home := homedir.HomeDir()
-	cookieFilePath := filepath.Join(home, ".porter", cookieFileName)
-
-	client := &Client{
-		BaseURL:        baseURL,
-		CookieFilePath: cookieFilePath,
-		HTTPClient: &http.Client{
-			Timeout: time.Minute,
-		},
-	}
-
-	cookie, _ := client.getCookie()
-
-	if cookie != nil {
-		client.Cookie = cookie
-	}
-
-	// look for a cloudflare access token specifically for Porter
-	if cfToken := os.Getenv("PORTER_CF_ACCESS_TOKEN"); cfToken != "" {
-		client.cfToken = cfToken
-	}
-
-	return client
-}
 
 func (c *Client) getRequest(relPath string, data interface{}, response interface{}) error {
 	vals := make(map[string][]string)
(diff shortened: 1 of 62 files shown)


api/client/api.go
```diff
@@ -82,33 +80,6 @@ func NewClientWithConfig(ctx context.Context, input NewClientInput) (Client, err
 // ErrNoAuthCredential returns an error when no auth credentials have been provided such as cookies or tokens
 var ErrNoAuthCredential = errors.New("unable to create an API session with cookie nor token")
 
-// NewClient constructs a new client based on a set of options
```
stefanmcshane, line 85: Housekeeping - unused function

cli/cmd/config/config.go
```diff
@@ -56,6 +62,11 @@ func initAndLoadConfig() (CLIConfig, error) {
 	viper.SetConfigType("yaml")
 	viper.AddConfigPath(porterDir)
 
+	err = createAndLoadPorterYaml(porterDir)
```
stefanmcshane, line 65: Evaluate porter.yaml before flags and environment variables

cli/cmd/commands/all.go
```diff
@@ -50,3 +50,69 @@ func RegisterCommands() (*cobra.Command, error) {
 	rootCmd.AddCommand(registerCommand_Version(cliConf))
 	return rootCmd, nil
 }
+
+// overrideConfigWithFlags grabs the runtime value of registered flags, and overrides the values in CLIConfig.
+// It was done this was to reduce the size of a refactor, as the codebase conflates initialisation of the commands, with the runtime values.
```
ferozemohideen, line 55: nit: comment missing a word?

cli/cmd/commands/all.go
```diff
@@ -50,3 +50,69 @@ func RegisterCommands() (*cobra.Command, error) {
...
+		{
+			uintName:         "helm_repo",
+			uintConfigTarget: &config.HelmRepo,
+		},
+	}
+	for _, fl := range flagsToOverride {
+		if fl.stringName != "" {
+			st, _ := cmd.Flags().GetString(fl.stringName)
+			if st != "" {
+				*fl.stringConfigTarget = st
```
ferozemohideen, line 107: are these pointer operations panic safe?

ferozemohideen (approved):

cli/cmd/commands/auth.go
```diff
@@ -53,8 +60,9 @@ func registerCommand_Auth(cliConf config.CLIConfig) *cobra.Command {
 		Use:   "logout",
 		Short: "Logs a user out of a given Porter server",
 		Run: func(cmd *cobra.Command, args []string) {
-			err := checkLoginAndRunWithConfig(cmd.Context(), cliConf, args, logout)
+			err := checkLoginAndRunWithConfig(cmd, cliConf, args, logout)
 			if err != nil {
+				_ = cliConf.SetToken("")
```
ianedwards, line 65: Would it make sense to clear project / cluster as well? Or do we set those on login

cli/cmd/config/config.go
```diff
@@ -56,6 +59,11 @@ func initAndLoadConfig() (CLIConfig, error) {
...
```
ianedwards, line 62: Is this referring to the file where configuration is stored?

ianedwards (approved):

cli/cmd/commands/all.go
stefanmcshane: yeah - they are only used in this one function as an easier way to read the command line values in for now. There are much much better ways to manage this, but that seemed outside the scope of the PR for now as its big enough

cli/cmd/config/config.go
stefanmcshane: It is - welcome to viper. It stores an internal global state and I plan to remove it. The project that cobra and viper were created for (hugo) no longer use viper, and use a stripped down version of cobra

cli/cmd/commands/auth.go
stefanmcshane: They get set on login

PR #3452, opened 2023-08-28
Project Rename

sdess09:
## What does this PR do?

image
image


jusrhee:
Could we make the default text gray (not white) and decrease the vertical spacing so it's the same as the space under "Delete project"

jusrhee:
could we add one more line of spacing beneath the "change name" button so it's clearly two distinct blocks

jusrhee (approved):

PR #3462, opened 2023-08-30
correctly support porter yaml private web field

ianedwards:
## What does this PR do?

- Added a private field to porter.yaml for web services
- Need to make sure this accounted for when parsing yaml, and creating/updating apps on the client

### DIFF
--- a/dashboard/src/main/home/app-dashboard/app-view/AppDataContainer.tsx
+++ b/dashboard/src/main/home/app-dashboard/app-view/AppDataContainer.tsx
@@ -20,6 +20,11 @@ import { useQueryClient } from "@tanstack/react-query";
 import Settings from "./tabs/Settings";
 import BuildSettings from "./tabs/BuildSettings";
 import Environment from "./tabs/Environment";
+import AnimateHeight from "react-animate-height";
+import Banner from "components/porter/Banner";
+import Button from "components/porter/Button";
+import Icon from "components/porter/Icon";
+import save from "assets/save-01.svg";
 
 // commented out tabs are not yet implemented
 // will be included as support is available based on data from app revisions rather than helm releases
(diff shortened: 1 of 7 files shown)


dashboard/src/main/home/app-dashboard/app-view/AppDataContainer.tsx
```diff
@@ -172,6 +203,27 @@ const AppDataContainer: React.FC<AppDataContainerProps> = ({ tabParam }) => {
           sourceType={latestSource.type}
         />
         <Spacer y={1} />
+        <AnimateHeight height={isDirty && !onlyExpandedChanged ? 67 : 0}>
```
ferozemohideen, line 206: why do you need to hardcode height here? didn't you get the resize observer to work?

dashboard/src/main/home/app-dashboard/app-view/AppDataContainer.tsx
```diff
@@ -101,9 +106,35 @@ const AppDataContainer: React.FC<AppDataContainerProps> = ({ tabParam }) => {
...
+        if (typeof value === "object") {
+          dirty.push(...getAllDirtyFields(value));
+        }
+      }
+    });
+
+    return dirty;
+  };
+
+  const onlyExpandedChanged = useMemo(() => {
```
ferozemohideen, line 130: can you add a comment to these functions? kinda hard to understand what they're used for

ferozemohideen (approved):

dashboard/src/main/home/app-dashboard/app-view/AppDataContainer.tsx
ianedwards: copy paste oversight. will update

ianedwards: I'm actually going to keep the ternary for now but use auto instead of a hardcoded value. Realizing that the content doesn't change so it's only going to resize once depending on if the condition passes, and we don't have to watch for it after that.

PR #3465, opened 2023-08-30
[CUS-618] Use webkit scrollbar to override MacOS hide scrollbar setting

ferozemohideen:
## CUS-618
N/A

## What does this PR do?

Add non-default scrollbar such that the mac setting doesn't hide it

image


sdess09 (approved):

sdess09:
Would be nice to include a screenshot for this next time :)

PR #3476, opened 2023-08-31
handle reverts and deletions on revisions

ianedwards:
## What does this PR do

- Passes along deletion keys of services and env variables to CCP so that they are removed in the final app proto that is applied

### DIFF
--- a/dashboard/src/lib/hooks/useAppValidation.ts
+++ b/dashboard/src/lib/hooks/useAppValidation.ts
@@ -23,15 +23,7 @@ export const useAppValidation = ({
     current: Record<string, string>,
     previous: Record<string, string>
   ) => {
-    const removed: string[] = [];
-
-    Object.entries(previous).forEach(([key, _]) => {
-      if (!current[key]) {
-        removed.push(key);
-      }
-    });
-
-    return removed;
+    return Object.keys(previous).filter((key) => !current[key]);
   };
 
   const getBranchHead = async ({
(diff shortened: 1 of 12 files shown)


dashboard/src/main/home/app-dashboard/validate-apply/services-settings/ServiceList.tsx
```diff
@@ -122,13 +130,25 @@ const ServiceList: React.FC<ServiceListProps> = ({
   };
 
   const onSubmit = handleSubmit(async (data) => {
+    const previouslyDeleted = deletedServices.findIndex(
+      (s) => s.name === data.name
+    );
+    if (previouslyDeleted !== -1) {
+      removeDeletion(previouslyDeleted);
+    }
```
ferozemohideen, line 138: I'm assuming this is so we don't send a deletion for a service that the user has just added without saving? can you include a comment about that?

dashboard/src/lib/hooks/useAppValidation.ts
```diff
@@ -19,6 +19,21 @@ export const useAppValidation = ({
...
+    const removed: string[] = [];
+
+    Object.entries(previous).forEach(([key, _]) => {
+      if (!current[key]) {
+        removed.push(key);
+      }
+    });
+
+    return removed;
+  };
```
ferozemohideen, line 35: nit
```suggestion
 const removedEnvKeys = (
  current: Record<string, string>,
  previous: Record<string, string>
) => {
  return Object.keys(previous).filter(key => !(key in current));
};
```

ferozemohideen (approved):

PR #3504, opened 2023-09-05
[POR-1503] Refactor project retrieval

jose-fully-ported:
## What does this PR do?

This PR refactors the `GET /projects` endpoint to return the minimal data - `[]Project{ID: int, Name: string}` - as a response (and the corresponding changes for consumption). The change is necessary as feature flags are going to be retrieved from another API (LaunchDarkly) and retrieving them for every project on every payload (instead of just the current project) is expensive for users with multiple projects.

I believe with this change, we'll also need to make a release of the CLI to handle the response as the `Project` type might not deserialize properly with missing keys (I don't remember the deserialization behavior of golang, but will update this once I figure that out).

I also used this opportunity to start adding typing _inside_ of the api calls - via `res.data` casting - but ideally we're generating response types globally in the api itself vs how we're doing it here on a per-call basis.


dashboard/src/main/home/Home.tsx
```diff
@@ -125,39 +124,39 @@ const Home: React.FC<Props> = (props) => {
...
+        redirectToNewProject();
+      } else if (projectList.length > 0 && !currentProject) {
+        setProjects(projectList);
+
+        let foundProjectListEntry: ProjectListType | undefined;
+
+        if (id) {
+          foundProjectListEntry = (projectList).find(
+            (item: ProjectListType) => item.id == id
+          );
```
ferozemohideen, line 142: does this work?
```suggestion
        const foundProjectListEntry = projectList.find(
            (item: ProjectListType) => item.id === (id ?? localStorage.getItem("currentProject"))
          );
```

ferozemohideen (approved):

dashboard/src/main/home/Home.tsx
jose-fully-ported: Almost. The localstorage data is stored as a string, not integer. I'll try a modified version of what you posted though.

api/types/project.go
```diff
@@ -1,5 +1,28 @@
 package types
 
+// ProjectList type for entries in the api response on GET /projects
+type ProjectList struct {
+	ID   uint   `json:"id"`
+	Name string `json:"name"`
+
+	// note: all of these fields should be considered deprecated
```
stefanmcshane, line 8: Adding `Deprecated: ...` to each field will flag it in the linter if it gets used when it shouldnt be

internal/models/project.go
```diff
@@ -100,3 +100,36 @@ func (p *Project) ToProjectType() *types.Project {
 		FullAddOns:             p.FullAddOns,
 	}
 }
+
+// ToProjectListType returns a "minified" version of a Project
+// suitable for api responses to GET /projects
+// TODO: update this in the future to use default values for all
+// the feature flags instead of trying to retrieve them from the database
+func (p *Project) ToProjectListType() *types.ProjectList {
+	roles := make([]*types.Role, 0)
```
stefanmcshane, line 109: If you are declaring capacity, declare the full capacity, otherwise consider using `var roles []*types.Role`.

api/types/project.go
```diff
@@ -1,5 +1,28 @@
...
+	// in an api response
+	Roles                  []*Role `json:"roles"`
```
stefanmcshane, line 10: Can we remove the pointer from this? 
Can add `omitempty` if the goal is to not have it appear when empty, or just allow it to always be non-nil, and just empty (avoid them panics)

stefanmcshane (approved):
Some comments. LGTM otherwise

api/types/project.go
jose-fully-ported: I don't want to add that here as they will be used in the next PR (with a default assignment so that we don't break api responses) and also aren't being removed. Trying to avoid the linter yelling at me in those cases.

PR #3505, opened 2023-09-05
[POR-1503] pull project feature flags from launch darkly

jose-fully-ported:
## What does this PR do?

Pulls project feature flags from launch darkly. If the client could not be instantiated, each feature flag will be set to the default value.

Note that if launch darkly has an outage, we'll also use the default values. They have cache backends we could run (eg. redis or dynamodb) but that might be overkill for a first pass.

## Errata

- [x] This also requires changes in our private workstation repository to get the environment variables up
- [x] We need to ensure LD SDK keys are provisioned in all environments
- [x] All environments need a final pass of the feature flags migrated


internal/features/launch_darkly.go
```diff
@@ -0,0 +1,54 @@
...
+	client *ld.LDClient
+}
+
+// BoolVariation returns the value of a boolean feature flag for a given evaluation context.
+//
+// Returns defaultVal if there is an error, if the flag doesn't exist, or the feature is turned off and
+// has no off variation.
+//
+// For more information, see the Reference Guide: https://docs.launchdarkly.com/sdk/features/evaluating#go
+func (c *Client) BoolVariation(field string, context ldcontext.Context, defaultValue bool) (bool, error) {
```
stefanmcshane, line 24: Do we need to use a pointer for the client if the underlying ldclient is always shared? I usually default to not using pointers unless there is a good reason to do so

internal/models/project.go
```diff
@@ -74,31 +75,38 @@ type Project struct {
 }
 
 // ToProjectType generates an external types.Project to be shared over REST
-func (p *Project) ToProjectType() *types.Project {
+func (p *Project) ToProjectType(launchDarklyClient *features.Client) *types.Project {
```
stefanmcshane, line 78: Does the receiver and return value have to be a pointer?

internal/features/launch_darkly.go
jose-fully-ported: I think we want to use a pointer as the ldclient has an attached event processor (amongst other things) that it cleans up on destruct - my concern with using it by value is that we'll be duplicating requests or incorrectly destructing the object.

As an aside, I'm thinking about this as a database connection, where we only have a single connection across the entire application (vs spawning as needed).

internal/models/project.go
jose-fully-ported: See my above comment re: received.

Not sure the return value has to be a pointer, but I decided to not change that as it was already a pointer and didn't want this change to be invasive re: cleanup.

internal/features/launch_darkly.go
stefanmcshane: If we are passing the client by copy, the underlying ldclient will still maintain its single instance since we will be passing the address at all times though. That would make sense if `Client` was the one pooling the connections, as opposed to the underlying `ldclient` which is the singular connection

internal/models/project.go
stefanmcshane: I would say that pointer cleanup is fair game when we inherit a new part of the code as it makes for much safer code (which can be optimised by the compiler if performance were the concern)

internal/features/launch_darkly.go
jose-fully-ported: Ah I see what you mean. That makes sense, will update.

api/server/handlers/project/get_test.go
```diff
@@ -33,8 +34,10 @@ func TestGetProjectSuccessful(t *testing.T) {
 
 	handler.ServeHTTP(rr, req)
 
-	expProject := proj.ToProjectType()
-	gotProject := &types.Project{}
+	expProject := proj.ToProjectType(features.NewClient(nil))
+	gotProject := types.Project{}
 
-	apitest.AssertResponseExpected(t, rr, expProject, gotProject)
+	// note: these need to be passed as pointer values
```
stefanmcshane, line 40: Add this as a comment on the `AssertResponseExpected`?

internal/features/launch_darkly.go
```diff
@@ -0,0 +1,56 @@
...
+
+// BoolVariation returns the value of a boolean feature flag for a given evaluation context.
+//
+// Returns defaultVal if there is an error, if the flag doesn't exist, or the feature is turned off and
+// has no off variation.
+//
+// For more information, see the Reference Guide: https://docs.launchdarkly.com/sdk/features/evaluating#go
+func (c Client) BoolVariation(field string, context ldcontext.Context, defaultValue bool) (bool, error) {
+	if c.client == nil {
+		println("client missing")
```
stefanmcshane, line 26: remove this

internal/features/launch_darkly.go
```diff
@@ -0,0 +1,56 @@
...
+
+// NewClient returns a Client wrapping a launchdarkly client instance
+func NewClient(client *ld.LDClient) *Client {
+	return &Client{
+		client: client,
+	}
+}
+
+// GetClient retrieves a Client for interacting with LaunchDarkly
+func GetClient(envConf *envloader.EnvConf) (*Client, error) {
```
stefanmcshane, line 41: Where can we use GetClient instead of NewClient? Is this just for global state?

stefanmcshane (approved):

internal/features/launch_darkly.go
jose-fully-ported: `NewClient()` is used in tests... we could probably just construct an empty client instead. I'll do that.

stefanmcshane: Idiomatically, if you are creating a Client, the constructor-esque function would be NewClient

PR #3507, opened 2023-09-06
[POR-1654] support revision numbers on logs (applyv2)

d-g-town:
This PR adds revision numbers and the revision filter to apply v2 logs.

### DIFF
--- a/dashboard/src/main/home/app-dashboard/validate-apply/logs/Logs.tsx
+++ b/dashboard/src/main/home/app-dashboard/validate-apply/logs/Logs.tsx
@@ -100,12 +100,6 @@ const Logs: React.FC<Props> = ({
     }
 
     const createVersionOptions = (number: number) => {
-        console.log(Array.from({ length: number }, (_, index) => {
-            const version = index + 1;
-            const label = version === number ? `Version ${version} (latest)` : `Version ${version}`;
-            const value = version.toString();
-            return GenericFilterOption.of(label, value);
-        }).reverse().slice(0, 3))
         return Array.from({ length: number }, (_, index) => {
             const version = index + 1;
             const label = version === number ? `Version ${version} (latest)` : `Version ${version}`;
--- a/dashboard/src/main/home/app-dashboard/validate-apply/logs/Logs.tsx
+++ b/dashboard/src/main/home/app-dashboard/validate-apply/logs/Logs.tsx
@@ -26,6 +26,8 @@ import Button from "components/porter/Button";
 import { Service } from "../../new-app-flow/serviceTypes";
 import LogFilterContainer from "../../expanded-app/logs/LogFilterContainer";
 import StyledLogs from "../../expanded-app/logs/StyledLogs";
+import {z} from "zod";
+import {appRevisionValidator} from "../../../../../lib/revisions/types";
 
 type Props = {
     projectId: number;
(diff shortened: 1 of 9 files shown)


dashboard/src/main/home/app-dashboard/validate-apply/logs/Logs.tsx
```diff
@@ -26,13 +26,16 @@ import Button from "components/porter/Button";
 import { Service } from "../../new-app-flow/serviceTypes";
 import LogFilterContainer from "../../expanded-app/logs/LogFilterContainer";
 import StyledLogs from "../../expanded-app/logs/StyledLogs";
+import {z} from "zod";
+import {AppRevision, appRevisionValidator} from "../../../../../lib/revisions/types";
```
ferozemohideen, line 30:
```suggestion
import {AppRevision, appRevisionValidator} from "lib/revisions/types";
```

dashboard/src/main/home/app-dashboard/validate-apply/logs/Logs.tsx
```diff
@@ -41,13 +44,16 @@ const Logs: React.FC<Props> = ({
...
     deploymentTargetId,
+    latestRevision,
 }) => {
     const scrollToBottomRef = useRef<HTMLDivElement | undefined>(undefined);
     const [scrollToBottomEnabled, setScrollToBottomEnabled] = useState(true);
     const [enteredSearchText, setEnteredSearchText] = useState("");
     const [searchText, setSearchText] = useState("");
     const [selectedDate, setSelectedDate] = useState<Date | undefined>(undefined);
     const [notification, setNotification] = useState<string>();
+    const [revisionMap, setRevisionMap] = useState<Map<string, number>>(new Map());
```
ferozemohideen, line 55: nit: Record type is preferred if the key is just a string type

dashboard/src/main/home/app-dashboard/validate-apply/logs/Logs.tsx
```diff
@@ -108,13 +123,26 @@ const Logs: React.FC<Props> = ({
...
+            }
+        },
         {
             name: "output_stream",
             displayName: "Output Stream",
             default: GenericLogFilter.getDefaultOption("output_stream"),
-            options: serviceNames.map(s => {
-                return GenericFilterOption.of(s, s)
-            }) ?? [],
+            options: [
```
ferozemohideen, line 142: nice catch

dashboard/src/main/home/app-dashboard/validate-apply/logs/Logs.tsx
```diff
@@ -132,6 +160,37 @@ const Logs: React.FC<Props> = ({
...
+        return revisions;
+    };
+
+    useEffect(() => {
+        getRevisions().then((revisions) => {
+            setRevisionMap(new Map(revisions.app_revisions.map((revision) => [revision.id, revision.revision_number])))
+            setLatestRevisionNumber(revisions.app_revisions.map((revision) => revision.revision_number).reduce((a, b) => Math.max(a, b), 0))
+        }).catch((err) => console.log(err));
+
+    }, [projectId, clusterId, appName, deploymentTargetId, latestRevision]);
```
ferozemohideen, line 191: See [RevisionsList](https://github.com/porter-dev/porter/blob/master/dashboard/src/main/home/app-dashboard/app-view/RevisionsList.tsx#L55) for the same logic wrapped in a useQuery. One of the benefits of using the useQuery is we can more easily handle loading and error states as well. 

I would recommend putting this logic into a hook similar to the useLatestRevision() hook and using that hook here as well as RevisionList

ferozemohideen (approved):
approving for now but definitely worth refactoring that revisions list logic into a hook since we might use it elsewhere

PR #3514, opened 2023-09-07
[CUS-573] Moving revision culling

stefanmcshane:
## What does this PR do?

- Move revision culler to prod cluster on `move-to-production` instead of prerelease
- Accidentally removed a span which needs to go into prod anyway due to the number of traces it generates. Should sample later

Also set the job to run `0 4 * * */6` (4am UTC every sunday)

### DIFF
--- a/.github/workflows/production.yml
+++ b/.github/workflows/production.yml
@@ -1,5 +1,6 @@
 "on":
   push:
+    branches: ["sms/cull-patch"]
     tags:
       - production
 name: Deploy Porter to Production
@@ -21,7 +22,7 @@ jobs:
         uses: actions/setup-go@v4
         with:
           cache: false
-          go-version: '1.20.5'
+          go-version: "1.20.5"
           go-version-file: go.mod
       - name: Download Go Modules
         run: go mod download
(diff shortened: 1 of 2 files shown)


internal/kubernetes/config.go
```diff
@@ -400,7 +400,7 @@ func (conf *OutOfClusterConfig) GetClientConfigFromCluster(ctx context.Context)
 
 func (conf *OutOfClusterConfig) CreateRawConfigFromCluster(ctx context.Context) (*api.Config, error) {
 	ctx, span := telemetry.NewSpan(ctx, "ooc-create-raw-config-from-cluster")
-	defer span.End()
+	// defer span.End() // This span is one of most frequent spans. We need to sample this. For now, this span will not send
```
stefanmcshane, line 403: This shouldnt be part of this PR, but needs to be in prod nonetheless

.github/workflows/production.yml
```diff
@@ -21,7 +21,6 @@ jobs:
         uses: actions/setup-go@v4
         with:
           cache: false
-          go-version: '1.20.5'
```
stefanmcshane, line 24: This is read from `go-version-file` by default. This line has no effect

jose-fully-ported (approved):

PR #3527, opened 2023-09-08
remove steps from taskfile that were erroring out for cli-prerelease

ferozemohideen:
## POR-
N/A


stefanmcshane:
does this actually fix the issue where local tags may get clobbered in comparison to remote tags?

PR #3535, opened 2023-09-11
[towards POR-1668] correctly parse build and pre-deploy events in activity feed

ferozemohideen:


dashboard/src/main/home/app-dashboard/app-view/tabs/activity-feed/events/types.ts
```diff
@@ -56,6 +59,9 @@ export const porterAppEventValidator = z.object({
     if (data.type === PorterAppEventType.BUILD) {
         return porterAppBuildEventMetadataValidator.safeParse(data.metadata).success;
     }
+    if (data.type === PorterAppEventType.PRE_DEPLOY) {
+        return porterAppPreDeployEventMetadataValidator.safeParse(data.metadata).success;
+    }
```
ianedwards, line 64: Not blocking but some suggestions to consider for this validator:
1. Avoid typescript enums where possible in favor of string literals ([see here](https://www.executeprogram.com/blog/typescript-features-to-avoid))
2. In addition to the benefits mentioned in the article, you could also use zod's discriminated union to match on type and define the schemas for each type independently of one another, removing the need for the refine call

ianedwards (approved):

PR #3537, opened 2023-09-11
remove revision from deploy event metadata

ferozemohideen:


dashboard/src/main/home/app-dashboard/app-view/tabs/activity-feed/ActivityFeed.tsx
```diff
@@ -57,6 +57,7 @@ const ActivityFeed: React.FC<Props> = ({ appName, deploymentTargetId, currentClu
             setHasError(false)
         } catch (err) {
             setHasError(true);
+            console.log(err);
```
ferozemohideen, line 60: leaving this around for now in case we run into other zod validation errors; will remove later

PR #3539, opened 2023-09-11
[POR-1503] simplify retrieval of launchdarkly flags

jose-fully-ported:
## What does this PR do?

Rather than require adding a new function for each flag, simply add it to the map and call the helper function.


internal/models/project.go
```diff
@@ -1,13 +1,33 @@
...
 	"gorm.io/gorm"
 
+	"github.com/launchdarkly/go-sdk-common/v3/ldcontext"
 	"github.com/porter-dev/porter/api/types"
 	"github.com/porter-dev/porter/internal/features"
 	ints "github.com/porter-dev/porter/internal/models/integrations"
 )
 
+// ProjectFeatureFlags keeps track of all project-related feature flags
+var ProjectFeatureFlags = map[string]bool{
```
ianedwards, line 15: small suggestion might be to change this to const

ianedwards (approved):

internal/models/project.go
jose-fully-ported: You can't do that because maps cannot be used as constants.

PR #3542, opened 2023-09-11
[POR-1503] respect ld feature flags across codebase

jose-fully-ported:
## What does this PR do?

This ensures we always fetch feature flags across the codebase using the LaunchDarkly client instead of the deprecated database fields.

### DIFF
--- a/internal/models/project.go
+++ b/internal/models/project.go
@@ -11,21 +11,35 @@ import (
 	ints "github.com/porter-dev/porter/internal/models/integrations"
 )
 
+const APITokensEnabled = "api_tokens_enabled"
+const AzureEnabled = "azure_enabled"
+const CapiProvisionerEnabled = "capi_provisioner_enabled"
+const EnableReprovision = "enable_reprovision"
+const FullAddOns = "full_add_ons"
+const HelmValuesEnabled = "helm_values_enabled"
+const ManagedInfraEnabled = "managed_infra_enabled"
+const MultiCluster = "multi_cluster"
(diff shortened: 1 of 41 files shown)


internal/models/project.go
```diff
@@ -11,21 +11,60 @@ import (
...
+// AzureEnabled enables Azure Provisioning
+const AzureEnabled = "azure_enabled"
+
+// CapiProvisionerEnabled enables the CAPI Provisioning flow
+const CapiProvisionerEnabled = "capi_provisioner_enabled"
+
+// EnableReprovision enables the provisioning button after initial creation of the cluster
+const EnableReprovision = "enable_reprovision"
+
+// FullAddOns shows all addons, not just curated
```
ianedwards, line 26: Not necessary and not sure if there's a trade off here but these consts could be an alias type, which is used as the type on the argument to `GetFeatureFlag`.

Something similar to as follows:

```
type FeatureFlagLabel string

const (
  // AzureEnabled enables Azure Provisioning
  AzureEnabled FeatureFlagLabel = "azure_enabled"

  // CapiProvisionerEnabled enables the CAPI Provisioning flow
  CapiProvisionerEnabled FeatureFlagLabel = "capi_provisioner_enabled"

  // EnableReprovision enables the provisioning button after initial creation of the cluster
  EnableReprovision FeatureFlagLabel = "enable_reprovision"

  ...
)
```

ianedwards (approved):

internal/models/project.go
jose-fully-ported: Good call, updated.

PR #3551, opened 2023-09-12
move creation of porter domain to apply endpoint

ianedwards:

### DIFF
--- a/api/server/handlers/porter_app/apply.go
+++ b/api/server/handlers/porter_app/apply.go
@@ -1,7 +1,10 @@
 package porter_app
 
 import (
+	"context"
 	"encoding/base64"
+	"errors"
+	"fmt"
 	"net/http"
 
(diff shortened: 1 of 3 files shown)


api/server/handlers/porter_app/apply.go
```diff
@@ -164,3 +193,34 @@ func (c *ApplyPorterAppHandler) ServeHTTP(w http.ResponseWriter, r *http.Request
 
 	c.WriteResult(w, r, response)
 }
+
+func addPorterSubdomainsIfNecessary(ctx context.Context, app *porterv1.PorterApp, createSubdomainInput porter_app.CreatePorterSubdomainInput) (*porterv1.PorterApp, error) {
```
ferozemohideen, line 197: nit: add comment here about when adding porter subdomains is necessary

ferozemohideen (approved):

PR #3554, opened 2023-09-12
POR-1703: remove direct access to the powerdns client

jose-fully-ported:
## What does this PR do?

As part of POR-1703, we're adding support for Cloudflare for use in testing/local development. This change  will allow us to implement other DNS backends in the case that PowerDNS is not available.


internal/integrations/dns/dns.go
```diff
@@ -0,0 +1,18 @@
...
+type Client struct {
+	Client DNSClient
+}
+
+type DNSClient interface {
+	CreateARecord(value, hostname string) error
+	CreateCNAMERecord(value, hostname string) error
+}
+
+func (c Client) CreateARecord(value, hostname string) error {
```
stefanmcshane, line 12: These functions dont need set here, only the implementations need to set them

jose-fully-ported: Without these, you're accessing the client directly, which would the opposite of the purpose of this PR.

stefanmcshane: In that case, this client would abstract away the need for an interface? In future, this would be the function which wraps both PowerDNS, and Cloudflare or any subsequent DNS?

With a `dns.NewClient` which takes in supported DNS providers

stefanmcshane (approved):

PR #3556, opened 2023-09-12
read latest app revision with source

ianedwards:
## What does this PR do?

- This PR serves as a way to list the latest revision details for apps in a given deployment target
- The source (porter_app) is also included in the returned result in order to provide better context about the app


api/server/handlers/porter_app/latest_app_revisions.go
```diff
@@ -0,0 +1,128 @@
...
+	}
+
+	if len(deploymentTargets) == 0 {
+		res := &LatestAppRevisionsResponse{
+			AppRevisions: []LatestRevisionWithSource{},
+		}
+
+		c.WriteResult(w, r, res)
+		return
+	}
```
d-g-town, line 69: check if > 1, in which case error out (to remind us to implement multiple deployment target handling)

d-g-town (approved):

PR #3557, opened 2023-09-12
Intelligent sliders

sdess09:
## What does this PR do?

image
image
image


jusrhee (approved):

ferozemohideen (commented):
lgtm for now, we'll have to refactor this when we move it to the new porter.yaml components because we won't have direct access to helm values

PR #3559, opened 2023-09-12
POR-1703: add support for cloudflare dns

jose-fully-ported:
## What does this PR do?

This change allows developers to specify a `CLOUDFLARE_API_TOKEN`. If set and powerdns is not configured, Porter will costruct a Cloudflare DNS client and use that for creating dns records.

In the case that neither are available, the current behavior of disabling egress domains will still occur.

Please note the cloudflare max dns records limit: https://community.cloudflare.com/t/limit-on-dns-records-per-domain/415781


api/server/shared/config/loader/loader.go
```diff
@@ -305,7 +304,14 @@ func (e *EnvConfigLoader) LoadConfig() (res *config.Config, err error) {
 	res.Logger.Info().Msg("Created analytics client")
 
 	if sc.PowerDNSAPIKey != "" && sc.PowerDNSAPIServerURL != "" {
-		res.DNSClient = &dns.Client{Client: powerdns.NewClient(sc.PowerDNSAPIServerURL, sc.PowerDNSAPIKey, sc.AppRootDomain)}
+		res.DNSClient = &dns.Client{Client: dns.NewPowerDNSClient(sc.PowerDNSAPIServerURL, sc.PowerDNSAPIKey, sc.AppRootDomain)}
+	} else if sc.CloudflareAPIToken != "" {
```
stefanmcshane, line 308: Do we need the else here? To stop anything changing for existing customers, we should instantiate cloudflare first, then overwrite the DNS client with PowerDNS

internal/integrations/dns/cloudflare.go
```diff
@@ -0,0 +1,88 @@
...
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/cloudflare/cloudflare-go"
+)
+
+// CloudflareRType strongly types cloudflare dns entry types
+type CloudflareRType string
```
stefanmcshane, line 11: What is RType? If it's record type, rename this to `Record` as the import for this is currently `cloudflare.CloudflareRType` as opposed to `cloudflare.Record`. The linter usually complains if you have the package as the first word

internal/integrations/dns/cloudflare.go
```diff
@@ -0,0 +1,88 @@
...
+
+const (
+	// CloudflareA declares an A record type for cloudflare
+	CloudflareA CloudflareRType = "A"
```
stefanmcshane, line 15: Same as above - verbosity is easier to understand over time. `cloudflare.ARecord` vs `cloudflare.CloudflareA`

internal/integrations/dns/cloudflare.go
```diff
@@ -0,0 +1,88 @@
...
+
+// CloudflareClient is a struct wrapper around the cloudflare client
+type CloudflareClient struct {
+	zoneID string
+
+	client *cloudflare.API
+}
+
+// NewCloudflareClient creates a new cloudflare API client
+func NewCloudflareClient(apiToken string, runDomain string) (*CloudflareClient, error) {
```
stefanmcshane, line 32: The underlying client being a pointer is enough

internal/integrations/dns/cloudflare.go
```diff
@@ -0,0 +1,88 @@
...
+		return nil, err
+	}
+
+	return &CloudflareClient{client: client, zoneID: zoneID}, nil
+}
+
+// CreateCNAMERecord creates a new CNAME record for the nameserver
+//
+// The method ignores record.RootDomain in favor of the zoneID derived from c.runDomain
+func (c *CloudflareClient) CreateCNAMERecord(record Record) error {
```
stefanmcshane, line 49: does this need to be a pointer? If it needs to be later, we can change it then

internal/integrations/dns/dns.go
```diff
@@ -1,19 +1,43 @@
...
+	A RecordType = iota
+
+	// CNAME represents a DNS CNAME record
+	CNAME
+)
+
+// WrappedClient is an interface describing a wrapper
+// around a particular dns implementation
+type WrappedClient interface {
+	CreateARecord(record Record) error
```
stefanmcshane, line 17: <3

internal/integrations/dns/dns.go
```diff
@@ -1,19 +1,43 @@
 package dns
 
-import "github.com/porter-dev/porter/internal/integrations/powerdns"
+// RecordType strongly types dns record types
+type RecordType int
+
+const (
+	// A represents a DNS A record
+	A RecordType = iota
```
stefanmcshane, line 8: Consider naming these as `RecordA` and `RecordCNAME`, or `RecordType_A` and `RecordType_CNAME`. Intellisense will autopopulate once you type `dns.Record`. An example of where this happens and is useful would be the net package in the std lib.

`A` alone will be confusing when used elsewhere imo.

internal/integrations/dns/powerdns.go
```diff
@@ -1,48 +1,51 @@
...
 }
 
-type RR struct {
-	Name       string   `json:"name"`
-	Type       string   `json:"type"`
-	ChangeType string   `json:"changetype"`
-	TTL        uint     `json:"ttl"`
-	Records    []Record `json:"records"`
+// PowerDNSRR represents a dns resource record collection for PowerDNS
+type PowerDNSRR struct {
```
stefanmcshane, line 38: Import here will now be `powerdns.PowerDNSRR`, consider using `powerdns.ResourceRecord`. If it wasnt for comments, Id be confused what an RR is

internal/integrations/dns/powerdns.go
```diff
@@ -1,48 +1,51 @@
...
+	Type       string           `json:"type"`
+	ChangeType string           `json:"changetype"`
+	TTL        uint             `json:"ttl"`
+	Records    []PowerDNSRecord `json:"records"`
 }
 
-type Record struct {
+// PowerDNSRecord represents an individual record for a given
+// PowerDNS resource record
+type PowerDNSRecord struct {
```
stefanmcshane, line 48: `powerdns.PowerDNS`

internal/integrations/dns/powerdns.go
```diff
@@ -51,17 +54,17 @@ type Record struct {
 }
 
 // CreateCNAMERecord creates a new CNAME record for the nameserver
-func (c *Client) CreateCNAMERecord(value, hostname string) error {
-	valueC := canonicalize(value)
-	hostnameC := canonicalize(hostname)
+func (c *PowerDNSClient) CreateCNAMERecord(record Record) error {
```
stefanmcshane, line 57: pointer receiver needed?

internal/integrations/dns/cloudflare.go
stefanmcshane: > Avoid repetition. Since client code uses the package name as a prefix when referring to the package contents, the names for those contents need not repeat the package name. The HTTP server provided by the http package is called Server, not HTTPServer. Client code refers to this type as http.Server, so there is no ambiguity.

https://go.dev/blog/package-names

stefanmcshane (commented):
Mainly nitpicks which come from idiomatic go to make things more understandable. I dont believe in using "this isnt idiomatic go" as an excuse - the highlighted things are onces that I agree with using, and are prevalent across the Go ecosystem

internal/integrations/dns/cloudflare.go
jose-fully-ported: The package here is actually dns, not cloudflare, so the import for this is actually dns.CloudflareRType. That said, it should probably be dns.CloudflareRecordType (a bit more verbose, but more accurately reflects the naming).

internal/integrations/dns/cloudflare.go
jose-fully-ported: Same answer as above :)

jose-fully-ported:
I split out the package so that we could follow better naming conventions.

stefanmcshane (approved):

PR #3563, opened 2023-09-13
POR-1708 route update full to v2 apply

ianedwards:
## What does this PR do?

- Handles `porter update` command as if it were a `porter apply` for validate_apply_v2 projects

### DIFF
--- a/cli/cmd/v2/deploy.go
+++ b/cli/cmd/v2/deploy.go
@@ -3,11 +3,21 @@ package v2
 import (
 	"context"
 	"fmt"
+
+	api "github.com/porter-dev/porter/api/client"
+	"github.com/porter-dev/porter/cli/cmd/config"
 )
 
 // UpdateFull implements the functionality of the `porter build` command for validate apply v2 projects
-func UpdateFull(ctx context.Context) error {
-	fmt.Println("This command is not supported for your project. Contact support@porter.run for more information.")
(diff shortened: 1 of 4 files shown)


cli/cmd/v2/deploy.go
```diff
@@ -3,11 +3,21 @@ package v2
...
 	"fmt"
+
+	api "github.com/porter-dev/porter/api/client"
+	"github.com/porter-dev/porter/cli/cmd/config"
 )
 
 // UpdateFull implements the functionality of the `porter build` command for validate apply v2 projects
-func UpdateFull(ctx context.Context) error {
-	fmt.Println("This command is not supported for your project. Contact support@porter.run for more information.")
+func UpdateFull(ctx context.Context, cliConf config.CLIConfig, client api.Client, appName string) error {
```
d-g-town, line 12: I know there are a bunch of flags that can be passed into the old update command that change build settings and build contexts and what not.  Would we also need to pass them in here?

d-g-town (approved):

cli/cmd/v2/deploy.go
ianedwards: Discussed plan to add telemetry ([ticket](https://linear.app/porter/issue/POR-1711/track-update-usage-relevant-to-apply-v2-flags-subcommands)) but assuming this should unblock most use cases

PR #3571, opened 2023-09-14
include force build flag on first apply to trigger rebuild

ianedwards:
## What does this PR do?

- Includes a flag indicating that an apply call should force a build regardless of the calculated image tag

### DIFF
--- a/cli/cmd/v2/apply.go
+++ b/cli/cmd/v2/apply.go
@@ -24,7 +24,9 @@ import (
 
 // Apply implements the functionality of the `porter apply` command for validate apply v2 projects
 func Apply(ctx context.Context, cliConf config.CLIConfig, client api.Client, porterYamlPath string, appName string) error {
+	const fromCLI = true
 	var yamlB64 string
+
 	if len(porterYamlPath) != 0 {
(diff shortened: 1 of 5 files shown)


cli/cmd/v2/apply.go
```diff
@@ -149,7 +151,7 @@ func Apply(ctx context.Context, cliConf config.CLIConfig, client api.Client, por
 
 		_ = updateExistingEvent(ctx, client, appName, cliConf.Project, cliConf.Cluster, targetResp.DeploymentTargetID, eventID, types.PorterAppEventStatus_Success, buildMetadata)
 
-		applyResp, err = client.ApplyPorterApp(ctx, cliConf.Project, cliConf.Cluster, "", "", applyResp.AppRevisionId)
+		applyResp, err = client.ApplyPorterApp(ctx, cliConf.Project, cliConf.Cluster, "", "", applyResp.AppRevisionId, !forceBuild)
```
ferozemohideen, line 154: nit: this command is starting to take a lot of arguments, maybe consider passing a struct instead?

ferozemohideen (approved):

PR #3576, opened 2023-09-15
POR-1728 Build env vars

ianedwards:
## What does this PR do?

- Endpoint for retrieving all public env vars associated with a given app revision
- Feed those vars into build opts when running build from apply

### DIFF
--- a/internal/porter_app/environment.go
+++ b/internal/porter_app/environment.go
@@ -1,0 +1,75 @@
+package porter_app
+
+import (
+	"context"
+
+	porterv1 "github.com/porter-dev/api-contracts/generated/go/porter/v1"
+	"github.com/porter-dev/porter/internal/kubernetes"
+	"github.com/porter-dev/porter/internal/kubernetes/environment_groups"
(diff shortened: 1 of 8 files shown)


internal/porter_app/environment.go
```diff
@@ -0,0 +1,75 @@
...
+}
+
+// AppEnvironmentFromProto returns all envfironment groups referenced in an app proto with their variables
+func AppEnvironmentFromProto(ctx context.Context, inp AppEnvironmentFromProtoInput, varOpts ...EnvVariableOption) ([]environment_groups.EnvironmentGroup, error) {
+	ctx, span := telemetry.NewSpan(ctx, "porter-app-env-from-proto")
+	defer span.End()
+
+	var envGroups []environment_groups.EnvironmentGroup
+
+	if inp.DeploymentTarget == nil {
```
d-g-town, line 34: it would be worth checking the the deploymentTarget cluster matches the cluster id scoped in the API call

d-g-town (approved):

PR #3577, opened 2023-09-15
create or update app env group on cli run

d-g-town:
This PR adds the ability to update the environment group for an app when running the CLI.

Follow up PR will ensure that a new environment group version is only created if any of the variables change (to reduce unnecessary versions). EDIT: This was also added to this PR.

### DIFF
--- a/api/server/handlers/porter_app/update_app_environment_group.go
+++ b/api/server/handlers/porter_app/update_app_environment_group.go
@@ -168,7 +168,6 @@ func (c *UpdateAppEnvironmentHandler) ServeHTTP(w http.ResponseWriter, r *http.R
 		telemetry.WithAttributes(span, telemetry.AttributeKV{Key: "same-env-group", Value: sameEnvGroup})
 
 		if sameEnvGroup {
-
 			res := &UpdateAppEnvironmentResponse{
 				EnvGroupName:    latestEnvironmentGroup.Name,
 				EnvGroupVersion: latestEnvironmentGroup.Version,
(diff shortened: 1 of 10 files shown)


api/server/handlers/porter_app/update_app_environment_group.go
```diff
@@ -0,0 +1,219 @@
...
+		err := telemetry.Error(ctx, span, nil, "invalid request")
+		c.HandleAPIError(w, r, apierrors.NewErrPassThroughToClient(err, http.StatusBadRequest))
+		return
+	}
+
+	if request.DeploymentTargetID == "" {
+		err := telemetry.Error(ctx, span, nil, "must provide deployment target id")
+		c.HandleAPIError(w, r, apierrors.NewErrPassThroughToClient(err, http.StatusBadRequest))
+		return
+	}
```
ianedwards, line 91: Any reason this endpoint in particular needs to take in a deployment target ID instead of just getting the first one from `List` (for now)?

api/server/handlers/porter_app/update_app_environment_group.go
```diff
@@ -0,0 +1,219 @@
...
+
+	res := &UpdateAppEnvironmentGroupResponse{
+		EnvGroupName:    split[0],
+		EnvGroupVersion: version,
+	}
+
+	c.WriteResult(w, r, res)
+}
+
+func AppEnvGroupName(ctx context.Context, appName string, deploymentTargetId string, clusterID uint, porterAppRepository repository.PorterAppRepository) (string, error) {
```
ianedwards, line 189: Could this go in internal potentially? See it being useful elsewhere

ianedwards (approved):

PR #3579, opened 2023-09-15
Fix bug where custom helm values for services were being wiped when deployed from the frontend

ferozemohideen:
How things were done previously:
- releaseValues and releaseDependencies are two objects passed to parse.go. They determine what 'base' to start with before we start overriding with the config in porter.yaml.
- The front-end passes nil releaseValues and releaseDependencies, since it is expected to provide a full porter.yaml, with the current state of the world.
- The CLI passes releaseValues and releaseDependencies from the latest helm release, because if the porter.yaml that it sends to parse is missing services, they will be supplemented with releaseValues and releaseDependencies

The problem:
- If a user adds custom helm values to an existing service of the release, next time the front-end hits deploy, only the front-end-specified values for that service will be propagated to the next release
- This wipes away the custom helm values

The solution:
- Both front-end and CLI will pass releaseValues and releaseDependencies to parse
- Front-end will just 'cut out' the releaseValues and releaseDependencies for services that are not defined in the front-end (effectively acting as deletions)
- Since front-end is using non-nil releaseValues as its base now, it will simply override the releaseValues with the values specified in the front-end, keeping values that already existed which the front-end has no control over (the custom helm values)

Things I tested:
- Adding custom helm values, updating an env variable, and ensuring that the custom helm values still exist in the new revision
- Deleting services from front-end still works
- Building the image and deploying from the CLI still works (does not wipe out existing services not defined in porter.yaml)

### DIFF
--- a/api/server/handlers/porter_app/create.go
+++ b/api/server/handlers/porter_app/create.go
@@ -106,33 +106,8 @@ func (c *CreatePorterAppHandler) ServeHTTP(w http.ResponseWriter, r *http.Reques
 
 	var releaseValues map[string]interface{}
 	var releaseDependencies []*chart.Dependency
-	// if shouldCreate || request.OverrideRelease {
-	// 	releaseValues = nil
-	// 	releaseDependencies = nil
-
-	// 	// this is required because when the front-end sends an update request with overrideRelease=true, it is unable to
-	// 	// get the image info from the release. unless it is explicitly provided in the request, we avoid overwriting it
-	// 	// by attempting to get the image info from the release or the provided helm values
-	// 	if helmRelease != nil && (imageInfo.Repository == "" || imageInfo.Tag == "") {
-	// 		if request.FullHelmValues != "" {
-	// 			imageInfo, err = attemptToGetImageInfoFromFullHelmValues(request.FullHelmValues)
-	// 			if err != nil {
-	// 				err = telemetry.Error(ctx, span, err, "error getting image info from full helm values")
-	// 				telemetry.WithAttributes(span, telemetry.AttributeKV{Key: "porter-yaml-base64", Value: porterYamlBase64})
-	// 				c.HandleAPIError(w, r, apierrors.NewErrPassThroughToClient(err, http.StatusBadRequest))
-	// 				return
-	// 			}
-	// 		} else {
-	// 			imageInfo = attemptToGetImageInfoFromRelease(helmRelease.Config)
-	// 		}
-	// 	}
-	// } else {
-	// 	releaseValues = helmRelease.Config
-	// 	releaseDependencies = helmRelease.Chart.Metadata.Dependencies
-	// }
-	// this is required because when the front-end sends an update request with overrideRelease=true, it is unable to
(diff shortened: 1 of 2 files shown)


api/server/handlers/porter_app/create.go
```diff
@@ -106,30 +106,29 @@ func (c *CreatePorterAppHandler) ServeHTTP(w http.ResponseWriter, r *http.Reques
...
+		if request.FullHelmValues != "" {
+			imageInfo, err = attemptToGetImageInfoFromFullHelmValues(request.FullHelmValues)
+			if err != nil {
+				err = telemetry.Error(ctx, span, err, "error getting image info from full helm values")
+				telemetry.WithAttributes(span, telemetry.AttributeKV{Key: "porter-yaml-base64", Value: porterYamlBase64})
+				c.HandleAPIError(w, r, apierrors.NewErrPassThroughToClient(err, http.StatusBadRequest))
+				return
 			}
+		} else {
+			imageInfo = attemptToGetImageInfoFromRelease(helmRelease.Config)
```
ferozemohideen, line 121: this code block didn't change, I just brought it out of its parent conditional since it is unrelated

api/server/handlers/porter_app/create.go
```diff
@@ -106,30 +106,29 @@ func (c *CreatePorterAppHandler) ServeHTTP(w http.ResponseWriter, r *http.Reques
...
 		}
+	}
+	if shouldCreate {
+		releaseValues = nil
+		releaseDependencies = nil
 	} else {
 		releaseValues = helmRelease.Config
 		releaseDependencies = helmRelease.Chart.Metadata.Dependencies
 	}
```
ferozemohideen, line 130: here is the meat of the change; previously, the front-end would also set releaseValues and releaseDependencies to nil, but no longer.

api/server/handlers/porter_app/parse.go
```diff
@@ -290,10 +293,12 @@ func buildUmbrellaChartValues(
...
-	// add back in the existing services that were not overwritten
-	for k, v := range existingValues {
-		if values[k] == nil {
-			values[k] = v
+	if !removeDeletedValues {
+		// add back in the existing services that were not overwritten
+		for k, v := range existingValues {
+			if values[k] == nil {
+				values[k] = v
+			}
```
ferozemohideen, line 301: this condition will be true for CLI (and API), meaning that changes will only be additive

api/server/handlers/porter_app/parse.go
```diff
@@ -540,22 +545,24 @@ func buildUmbrellaChart(application *Application, config *config.Config, project
...
+						return nil, fmt.Errorf("unable to determine type of existing dependency")
+					}
+					version, err := getLatestTemplateVersion(dep.Name, config, projectID)
+					if err != nil {
+						return nil, err
+					}
+					dep.Version = version
 				}
-				dep.Version = version
+				deps = append(deps, dep)
```
ferozemohideen, line 564: this mirrors the logic that adds existing values back in buildUmbrellaChartValues

d-g-town (approved):

PR #3580, opened 2023-09-15
POR-1732 Get app env

ianedwards:
## What does this PR do?

- Motivation: Right now the only way to get environment groups is to list them out as they're defined in the base namespace
- This endpoint serves to get the env variables as they would have been defined when the latest revision was deployed, based on the name and version of each env group as defined in the app contract.
- The other benefit of this endpoint is that it allows us to not have to list out all the env groups on the FE, get the attached apps, and check if the current app is included in the attached app list.

### DIFF
--- a/internal/porter_app/environment.go
+++ b/internal/porter_app/environment.go
@@ -12,6 +12,7 @@ import (
 
 type envVariarableOptions struct {
 	includeSecrets bool
+	envGroups      []string
 }
 
 // EnvVariableOption is a function that modifies AppEnvironmentFromProto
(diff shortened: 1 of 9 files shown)


internal/porter_app/environment.go
```diff
@@ -26,6 +27,12 @@ func WithSecrets() EnvVariableOption {
 	}
 }
 
+func WithEnvGroupFilter(envGroups []string) EnvVariableOption {
```
stefanmcshane, line 30: comments on exported functions

internal/porter_app/revisions.go
```diff
@@ -26,6 +29,50 @@ type Revision struct {
 	UpdatedAt time.Time `json:"updated_at"`
 }
 
+type GetAppRevisionInput struct {
```
stefanmcshane, line 32: Comments

stefanmcshane (approved):
some comments

PR #3582, opened 2023-09-15
Porter v2 env

sdess09:
## What does this PR do?

Creating a Porter Env Group on create


dashboard/src/lib/porter-apps/index.ts
```diff
@@ -326,7 +340,8 @@ export function clientAppFromProto(
     },
     services,
     predeploy,
-    env: proto.env,
+    env: [],
+    envGroups: proto.envGroups.map((eg) => ({ name: eg.name, version: eg.version })),
```
ianedwards, line 344: Just noting that we need to address this in a follow up PR: env shoiuld be equal to the variables from the default env group for the app.

dashboard/src/lib/hooks/useAppValidation.ts
```diff
@@ -70,12 +70,21 @@ export const useAppValidation = ({
...
 
+      if (data.app.env) {
+        data.app.env?.forEach(item => {
+          if (item && !item.deleted) { // Ensure item exists and is not deleted
+
+            variables[item.key] = item.value;
+
+          }
+        });
+      }
```
ianedwards, line 83: Not sure if you ran into a case where app.env was null but I think its of type array, so we might be able to get away without the check. Would also suggest the following:

```suggestion
      const variables = data.app.env
        .filter((e) => !e.deleted)
        .reduce((acc: Record<string, string>, curr) => {
          acc[curr.key] = curr.value;
          return acc;
        }, {});
```

dashboard/src/lib/hooks/useAppValidation.ts
```diff
@@ -70,12 +70,21 @@ export const useAppValidation = ({
...
       const envVariableDeletions = removedEnvKeys(
-        data.app.env,
+        variables,
         prevRevision?.env || {}
       );
```
ianedwards, line 87: Not necessarily going to impact functionality in this PR but env is deprecated so won't include the variables that changed. Can address this later when handling updates

dashboard/src/lib/hooks/useAppValidation.ts
```diff
@@ -125,7 +134,7 @@ export const useAppValidation = ({
         atob(validAppData.validate_b64_app_proto)
       );
 
-      return validatedAppProto;
+      return { validatedAppProto: validatedAppProto, env: data.app.env };
```
ianedwards, line 137: Does everywhere calling this function account for the change in return value?

dashboard/src/main/home/app-dashboard/create-app/CreateApp.tsx
```diff
@@ -199,10 +210,57 @@ const CreateApp: React.FC<CreateAppProps> = ({ history }) => {
...
+          env.reduce((_, item) => {
+            if (item && !item.deleted) {
+              if (item.hidden) {
+                secrets[item.key] = item.value;
+              } else {
+                variables[item.key] = item.value;
+              }
+            }
+          }, null);
+        }
```
ianedwards, line 226:
```suggestion
         const variables = env
          .filter((e) => !e.hidden && !e.deleted)
          .reduce((acc: Record<string, string>, item) => {
            acc[item.key] = item.value;
            return acc;
          }, {});
        const secrets = env
          .filter((e) => !e.deleted)
          .reduce((acc: Record<string, string>, item) => {
            if (item.hidden) {
              acc[item.key] = item.value;
            }
            return acc;
          }, {});
```

ianedwards (approved):
few suggestions but otherwise lgtm

dashboard/src/lib/hooks/useAppValidation.ts
sdess09: yes should be fine I checked everywhere it was called

PR #3598, opened 2023-09-19
POR-1756: add telemetry to pr creation endpoint

jose-fully-ported:
## What does this PR do?

I'm submitting calls to the api to create a PR for a repo and getting errors. At the moment, this endpoint doesn't show any metrics in honeycomb, making it difficult to debug. This PR adds some telemetry so we can see where exactly things are failing.

### DIFF
--- a/api/server/handlers/porter_app/create_secret_and_open_pr.go
+++ b/api/server/handlers/porter_app/create_secret_and_open_pr.go
@@ -17,6 +17,7 @@ import (
 	"github.com/porter-dev/porter/internal/auth/token"
 	"github.com/porter-dev/porter/internal/integrations/ci/actions"
 	"github.com/porter-dev/porter/internal/models"
+	"github.com/porter-dev/porter/internal/telemetry"
 )
 
 type OpenStackPRHandler struct {
@@ -34,6 +35,9 @@ func NewOpenStackPRHandler(
 }
 
 func (c *OpenStackPRHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
+	ctx, span := telemetry.NewSpan(r.Context(), "serve-open-stack-pr")
+	defer span.End()
+
 	user, _ := r.Context().Value(types.UserScope).(*models.User)
 	project, _ := r.Context().Value(types.ProjectScope).(*models.Project)
 	cluster, _ := r.Context().Value(types.ClusterScope).(*models.Cluster)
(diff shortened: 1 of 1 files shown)


api/server/handlers/porter_app/create_secret_and_open_pr.go
```diff
@@ -34,6 +35,9 @@ func NewOpenStackPRHandler(
 }
 
 func (c *OpenStackPRHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
+	ctx, span := telemetry.NewSpan(r.Context(), "serve-open-stack-pr")
+	defer span.End()
+
 	user, _ := r.Context().Value(types.UserScope).(*models.User)
 	project, _ := r.Context().Value(types.ProjectScope).(*models.Project)
 	cluster, _ := r.Context().Value(types.ClusterScope).(*models.Cluster)
```
ferozemohideen, line 43: use ctx here rather than r.Context()

api/server/handlers/porter_app/create_secret_and_open_pr.go
```diff
@@ -59,12 +66,16 @@ func (c *OpenStackPRHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 		// generate porter jwt token
 		jwt, err := token.GetTokenForAPI(user.ID, project.ID)
 		if err != nil {
-			c.HandleAPIError(w, r, apierrors.NewErrInternal(fmt.Errorf("error getting token for API: %w", err)))
+			err = fmt.Errorf("error getting token for API: %w", err)
+			err := telemetry.Error(ctx, span, err, err.Error())
+			c.HandleAPIError(w, r, apierrors.NewErrInternal(err))
```
ferozemohideen, line 71: i believe telemetry.Error also formats the error like you did, so you can just do
```suggestion
			err = telemetry.Error(ctx, span, err, "error getting token for API")
			c.HandleAPIError(w, r, apierrors.NewErrPassThroughToClient(err, http.StatusInternalServerError))
```

api/server/handlers/porter_app/create_secret_and_open_pr.go
```diff
@@ -78,7 +89,9 @@ func (c *OpenStackPRHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 			request.GithubRepoName,
 		)
 		if err != nil {
-			c.HandleAPIError(w, r, apierrors.NewErrInternal(fmt.Errorf("error generating secret: %w", err)))
+			err = fmt.Errorf("error generating secret: %w", err)
+			err := telemetry.Error(ctx, span, err, err.Error())
+			c.HandleAPIError(w, r, apierrors.NewErrInternal(err))
```
ferozemohideen, line 94: same thing here, also prefer passing all error messages to client and obfuscating if necessary on client-side

ferozemohideen (changes requested):

ferozemohideen (approved):

api/server/handlers/porter_app/create_secret_and_open_pr.go
jose-fully-ported: It doesn't, that why I'm doing it this way :(

I think it should, but that change is more invasive, so I'll save it for later.

PR #3604, opened 2023-09-20
Support 3rd party tools which require kubeconfig

stefanmcshane:
## What does this PR do?
Support 3rd party tools which require a kubeconfig such as k9s.

- Added `porter kubectl --print-kubeconfig` flag which prints a generated kubeconfig to the console, with 15 minutes expiry.
- Added `cobra.Command` to all run functions, so we can start moving away from globally defined flags, allowing flag parsing within each individually run function


stefanmcshane:
Secret was a temporary certificate that was used during testing on a local cluster, and no longer exists. False positive

d-g-town (approved):

PR #3609, opened 2023-09-20
POR-1781 associated attached env versions with revision

ianedwards:
## What does this PR do?

- Previously, the latest version of env groups associated with a revision were being moved to the namespace specified by a deployment target, but the revision was not accounting for the version incrementing
- This PR returns all env groups that were updated so that the new versions can be accounted for on an upcoming revision


sdess09 (approved):

d-g-town:
Do we also have to update the CLI logic to add these new versions to the proto?

PR #3611, opened 2023-09-20
fix: switch the default heroku builder to heroku-20

jose-fully-ported:
## What does this PR do?

The :22 builder requires injecting a Procfile in order for detection to work as all the buildpack groups depend on a Procfile to detect properly, so for now we'll just revert this.


d-g-town (approved):

edmorley:
@jose-fully-ported Hi! Could you explain a bit more about the issue you encountered with the newer builder image? We would like to know about any compatibility differences so we can either resolve them, or document them in the future :-)

PR #3614, opened 2023-09-20
[POR-1673] Implement Job History in New View

ferozemohideen:
## What does this PR do?

https://github.com/porter-dev/porter/assets/40551216/8bead4d2-d267-4e98-9569-166f4556bd9f

### DIFF
--- a/api/server/handlers/porter_app/job_status.go
+++ b/api/server/handlers/porter_app/job_status.go
@@ -10,6 +10,7 @@ import (
 	"github.com/porter-dev/porter/api/server/shared"
 	"github.com/porter-dev/porter/api/server/shared/apierrors"
 	"github.com/porter-dev/porter/api/server/shared/config"
+	"github.com/porter-dev/porter/api/server/shared/requestutils"
 	"github.com/porter-dev/porter/api/types"
 	"github.com/porter-dev/porter/internal/kubernetes"
 	"github.com/porter-dev/porter/internal/models"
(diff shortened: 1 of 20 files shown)


dashboard/src/lib/hooks/useJobs.ts
```diff
@@ -0,0 +1,93 @@
...
+        },
+        {
+            enabled: revisionIdToNumber != null,
+            refetchInterval: 5000,
+            refetchOnWindowFocus: false,
+        },
+    );
+
+    useEffect(() => {
+        if (data != null) {
```
ianedwards, line 85: I think this is going to be `undefined` by default

dashboard/src/main/home/app-dashboard/validate-apply/jobs/utils.ts
```diff
@@ -0,0 +1,46 @@
+import { JobRun } from "lib/hooks/useJobs";
+import { timeFrom } from "shared/string_utils";
+
+export const ranFor = (start: string, end?: string | number) => {
+    const duration = timeFrom(start, end);
```
ianedwards, line 5: Might make sense to do date processing like this on actual timestamps in the future instead of on strings

api/server/handlers/porter_app/job_status.go
```diff
@@ -0,0 +1,116 @@
...
+	telemetry.WithAttributes(span, telemetry.AttributeKV{Key: "namespace", Value: namespace})
+
+	agent, err := c.GetAgent(r, cluster, "")
+	if err != nil {
+		err = telemetry.Error(ctx, span, err, "unable to get agent")
+		c.HandleAPIError(w, r, apierrors.NewErrPassThroughToClient(err, http.StatusInternalServerError))
+		return
+	}
+
+	labels := []kubernetes.Label{{
```
d-g-town, line 98: I would throw the `porter.run/app-name` label in here as well, in case someone names their jobs the same in two different apps (and so pass the appName from the front end into this API request)

dashboard/src/main/home/app-dashboard/validate-apply/jobs/utils.ts
```diff
@@ -0,0 +1,46 @@
...
+    const endTimeStamp = jobRun.status.completionTime ? new Date(jobRun.status.completionTime).getTime() : Date.now()
+
+    const timeDifferenceMilliseconds = endTimeStamp - startTimeStamp;
+
+    const seconds = Math.floor(timeDifferenceMilliseconds / 1000);
+    const weeks = Math.floor(seconds / 604800);
+    const remainingDays = Math.floor((seconds % 604800) / 86400);
+    const remainingHours = Math.floor((seconds % 86400) / 3600);
+    const remainingMinutes = Math.floor((seconds % 3600) / 60);
+    const remainingSeconds = seconds % 60;
```
ianedwards, line 27: possible to do some of these conversions with built in date functions / a library instead? Or just make these consts

ianedwards (approved):

dashboard/src/main/home/app-dashboard/validate-apply/jobs/utils.ts
ferozemohideen: Yeah we should swap this out with a library that does this in the future, I just copied this from existing

dashboard/src/main/home/app-dashboard/validate-apply/jobs/utils.ts
ferozemohideen: good point, swapped it out to use a library we already import

PR #3615, opened 2023-09-20
let the porter app creation api call fail if build and image are missing, not cli

d-g-town:
Previously, the CLI would fail if a porter.yaml was specified and no build/image was included.  This was because it always tries to create a Porter App (though this is a no-op if an app was already created).  However, we should support cases were the app was created with image or build settings in the UI, and the CLI is running with a porter.yaml that excludes these.

This PR lets the Porter App creation logic decide whether to error out on missing image/build info. If the app already exists, it will successfully return it.

Note: if we ever want to support changing the build/image info, this logic will have to change.

### DIFF
--- a/cli/cmd/v2/apply.go
+++ b/cli/cmd/v2/apply.go
@@ -56,21 +56,24 @@ func Apply(ctx context.Context, cliConf config.CLIConfig, client api.Client, por
 		}
 		b64AppProto = parseResp.B64AppProto
 
-		// we only need to create the app if a porter yaml is provided (otherwise it must already exist)
-		createPorterAppDBEntryInp, err := createPorterAppDbEntryInputFromProtoAndEnv(parseResp.B64AppProto)
-		if err != nil {
-			return fmt.Errorf("error creating porter app db entry input from proto: %w", err)
-		}
-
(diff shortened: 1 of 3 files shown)


cli/cmd/v2/apply.go
```diff
@@ -72,21 +72,24 @@ func Apply(ctx context.Context, cliConf config.CLIConfig, client api.Client, por
...
+			return fmt.Errorf("unable to form porter app creation input from yaml: %w", err)
 		}
 
-		// override app name if provided
-		appName, err = appNameFromB64AppProto(parseResp.B64AppProto)
+		err = client.CreatePorterAppDBEntry(ctx, cliConf.Project, cliConf.Cluster, createPorterAppDBEntryInp)
 		if err != nil {
-			return fmt.Errorf("error getting app name from b64 app proto: %w", err)
+			if err.Error() == porter_app.ErrMissingSourceType.Error() {
+				return fmt.Errorf("cannot find existing Porter app with name %s and no build or image settings were specified in porter.yaml", appName)
```
ianedwards, line 90: I just noticed we're importing `api/server/handlers/porter_app` here. Is that okay?

ianedwards (approved):

cli/cmd/v2/apply.go
d-g-town: Yeah, in this case there aren't any circular dependencies.

PR #3621, opened 2023-09-21
POR-1788 fix linked apps

ianedwards:
## What does this PR do?

- Previously, linked apps were determined by finding the unique set of namespaces containing deployments that have the env group attached by label
- for validate/apply v2, we set an app name label that we can use instead, since many apps can now be deployed in the same namespace


sdess09 (commented):
LGTM but this can be a little tricky so do a full QA 

-Delete a linked app and ensure that it is removed from the env group linked section
-Remove an env group and ensure it is removed from the env group linked section 
-Deletion should be blocked for the env group if it is linked to an application

sdess09 (approved):

PR #3628, opened 2023-09-21
[POR-1798] Porter yaml v2 porter app run command

ferozemohideen:


cli/cmd/commands/app.go
```diff
@@ -43,6 +43,11 @@ var (
 	appMemoryMi      int
 )
 
+const (
+	cnb_life_cycle_launcher = "/cnb/lifecycle/launcher"
```
ianedwards, line 47: tiny nit: not sure we ever do snake case in the Go code

cli/cmd/commands/app.go
```diff
@@ -416,17 +441,32 @@ type appPodSimple struct {
...
+	if resp == nil {
+		return nil, false, errors.New("get pods response is nil")
+	}
 	pods := *resp
 
+	if len(pods) == 0 {
+		return nil, false, errors.New("no running pods found for this application")
+	}
+
+	var containerHasLauncherStartCommand bool
```
ianedwards, line 462: could move containerHasLauncherStartCommand to the top and return zero value on error

ianedwards (approved):
nice

cli/cmd/commands/app.go
ferozemohideen: renamed

PR #3631, opened 2023-09-22
POR-1797 lower latency by moving env off of revision

ianedwards:
## What does this PR do?

- Using `AttachEnvToRevision` was causing high latency by requiring that the env variables associated with each revision be fetched from the cluster, even when just listing revisions
- This PR removes the env variables from a given revision and fetches them in separately, loading them into the environment tab

### DIFF
--- a/dashboard/src/lib/hooks/usePorterYaml.ts
+++ b/dashboard/src/lib/hooks/usePorterYaml.ts
@@ -128,7 +128,6 @@ export const usePorterYaml = ({
           overrides: proto,
           useDefaults,
         });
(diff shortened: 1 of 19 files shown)


dashboard/src/lib/hooks/usePorterYaml.ts
```diff
@@ -125,6 +128,7 @@ export const usePorterYaml = ({
           overrides: proto,
           useDefaults,
         });
+        console.log("build from serviceOverrides", build);
```
ferozemohideen, line 131: remove

ferozemohideen (approved):

PR #3634, opened 2023-09-22
POR-1592 support branch specific deploys from cli

ianedwards:
## What does this PR do?

- Adds CLI support for preview environments from a given repo
- Creates a new deployment target based on the name of the checked out branch with preview set to true; adding this bool as a new field to deployment targets because I imagine there will be many instances when we'll want to query for all preview environments in a given cluster
- Removes scenarios where we fall back on the first deployment target as it will now be possible to have multiple
- Creates a domain specific to the (app name, branch name) pair


cli/cmd/v2/apply.go
```diff
@@ -24,27 +24,65 @@ import (
...
 	var b64AppProto string
 
-	targetResp, err := client.DefaultDeploymentTarget(ctx, cliConf.Project, cliConf.Cluster)
-	if err != nil {
-		return fmt.Errorf("error calling default deployment target endpoint: %w", err)
+	cliConf := inp.CLIConfig
+	client := inp.Client
+
+	var deploymentTargetID string
+	if !inp.PreviewApply {
```
d-g-town: nit: change the order of the if statement to `if inp.PreviewApply` because it's easier to understand positive boolean values rather than negative ones.

cli/cmd/v2/apply.go
```diff
@@ -24,27 +24,65 @@ import (
...
+			branchName = os.Getenv("GITHUB_REF_NAME")
+		} else if branch, err := git.CurrentBranch(); err == nil {
+			branchName = branch
+		}
+
+		if branchName == "" {
+			return errors.New("Branch name is empty. Please run apply in a git repository with access to the git CLI.")
+		}
+
+		targetResp, err := client.CreateDeploymentTarget(ctx, cliConf.Project, cliConf.Cluster, branchName, true)
```
d-g-town: will this return the existing one if it has already been created?

cli/cmd/v2/apply.go
d-g-town: I think we probably also could get rid of this else condition by first setting the deployment target id to default, and then changing it if it is a preview apply: 
```

targetResp, err := client.DefaultDeploymentTarget(ctx, cliConf.Project, cliConf.Cluster)
if err != nil {
	return fmt.Errorf("error calling default deployment target endpoint: %w", err)
}
deploymentTargetID = targetResp.DeploymentTargetID

if inp.PreviewApply {
    var branchName string
    if os.Getenv("GITHUB_REF_NAME") != "" {
	    branchName = os.Getenv("GITHUB_REF_NAME")
    } else if branch, err := git.CurrentBranch(); err == nil {
	    branchName = branch
    }
    
    if branchName == "" {
	    return errors.New("Branch name is empty. Please run apply in a git repository with access to the git CLI.")
    }
    
    targetResp, err := client.CreateDeploymentTarget(ctx, cliConf.Project, cliConf.Cluster, branchName, true)
if err != nil {
	return fmt.Errorf("error calling create deployment target endpoint: %w", err)
}
deploymentTargetID = targetResp.DeploymentTargetID
}
```

cli/cmd/v2/apply.go
```diff
@@ -24,27 +24,43 @@ import (
...
+type ApplyInput struct {
+	// CLIConfig is the CLI configuration
+	CLIConfig config.CLIConfig
+	// Client is the Porter API client
+	Client api.Client
+	// PorterYamlPath is the path to the porter.yaml file
+	PorterYamlPath string
+	// AppName is the name of the app
+	AppName string
+	// PreviewApply is true if this is a preview apply
```
d-g-town, line 37: I would flesh out this comment.  Something like `If PreviewApply is true, Apply with create a new deployment target matching current git branch and apply to that target"

d-g-town (approved):

cli/cmd/v2/apply.go
ianedwards: good call, this is the behavior I was intending but forgot to add it to the create endpoint. updated now

PR #3638, opened 2023-09-25
Networking hotfix

sdess09:


ferozemohideen (approved):

sdess09:
> So it seems like we're injecting `3000` into a bunch of textboxes, apart from the one for the port?

Its just a placeholder- the text a user will see if the text is blank (usually apart from the port value there are default values so users wont see the place holder text anyways) but I updated all of them to be more relevant

PR #3670, opened 2023-09-27
POR-1440 use previews config as overrides in validate apply flow

ianedwards:
## What does this PR do?

- Run validate using parsed preview override values from porter.yaml when running CLI
- Incorporate env variable overrides
- Pass final app contract using overrides to apply when creating new revision

### DIFF
--- a/cli/cmd/v2/apply.go
+++ b/cli/cmd/v2/apply.go
@@ -66,6 +66,9 @@ func Apply(ctx context.Context, inp ApplyInput) error {
 		}
 	}
 
+	// overrides incorporated into the app contract baed on the deployment target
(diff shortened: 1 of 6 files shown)


cli/cmd/v2/apply.go
```diff
@@ -66,6 +66,9 @@ func Apply(ctx context.Context, inp ApplyInput) error {
 		}
 	}
 
+	// overrides incorporated into the app contract baed on the deployment target
+	var b64AppOverides string
```
d-g-town, line 70: nit:  Over*r*ides

d-g-town (approved):

PR #3689, opened 2023-09-29
feat: add support for buildpack urns

jose-fully-ported:
## What does this PR do?

URNs pull directly from the CNB registry, and should be fine to reference as is. We previously only supported git/https urls for referencing buildpacks, which aren't always available for various reasons.


dashboard/src/main/home/app-dashboard/build-settings/buildpacks/AddCustomBuildpackComponent.tsx
```diff
@@ -3,7 +3,12 @@ import React, { useState } from "react";
 import styled, { keyframes } from "styled-components";
 import { Buildpack } from "../../types/buildpack";
 
-function isValidURL(url: string): boolean {
+function isValidBuildpack(url: string): boolean {
```
ferozemohideen, line 6: nit: i would name this isValidBuildpackURL

ferozemohideen (approved):

PR #3702, opened 2023-09-29
POR-1440 setup app templates for preview envs

ianedwards:
## What does this PR do?

- Creates a unique app template per app that serves as a base revision to hydrate initial preview env deploys
- This is necessary so that users can configure an app e2e without actually having to apply it to a deployment target
- In the future, we might have many different app templates for many environments. For now, this implementation just keeps it simple and assumes that templates and apps are 1:1


api/server/handlers/app_template/create.go
```diff
@@ -0,0 +1,114 @@
...
+		err := telemetry.Error(ctx, span, nil, "error decoding request")
+		c.HandleAPIError(w, r, apierrors.NewErrPassThroughToClient(err, http.StatusBadRequest))
+		return
+	}
+	if request.B64AppProto == "" {
+		err := telemetry.Error(ctx, span, nil, "b64 app proto is empty")
+		c.HandleAPIError(w, r, apierrors.NewErrPassThroughToClient(err, http.StatusBadRequest))
+		return
+	}
+	if request.PorterAppID == 0 {
```
d-g-town, line 67: add the app id from the request to the telemetry

api/server/router/porter_app.go
```diff
@@ -1241,5 +1241,63 @@ func getPorterAppRoutes(
...
 	})
 
+	// GET /api/projects/{project_id}/clusters/{cluster_id}/apps/{porter_app_name}/app-templates -> porter_app.NewGetAppTemplateHandler
+	getAppTemplateEndpoint := factory.NewAPIEndpoint(
+		&types.APIRequestMetadata{
+			Verb:   types.APIVerbGet,
+			Method: types.HTTPVerbGet,
+			Path: &types.Path{
+				Parent:       basePath,
+				RelativePath: fmt.Sprintf("%s/{%s}/app-templates", relPathV2, types.URLParamPorterAppName),
```
d-g-town, line 1251: nit: maybe change `app-templates` to `templates`. Also, if we only ever expect to have a single template per app, maybe make it `template`

ianedwards: templates make sense and seems cleaner. I could see a scenario though where there are multiple, if we support multiple environments

PR #3713, opened 2023-10-03
fix parsing of custom buildpack urls

ferozemohideen:
Previous regex parsing runs into a crashloop for some URLs; this PR simplifies the implementation to remove the regex entirely


jose-fully-ported (approved):
❤️

PR #3715, opened 2023-10-03
POR-1435 create app templates and commit preview env workflow

ianedwards:
## What does this PR do?

- Commit workflow file for preview envs to porter-preview-{app_name}.yml
- Save app template which will be used on initial deploy to a new deployment target
- Filter out any default app env groups when creating the app template (they aren't needed)

### DIFF
--- a/api/server/handlers/porter_app/create_secret_and_open_pr.go
+++ b/api/server/handlers/porter_app/create_secret_and_open_pr.go
@@ -54,6 +54,17 @@ func (c *OpenStackPRHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	if request.Branch == "" {
+		err := telemetry.Error(ctx, span, nil, "branch cannot be empty")
+		c.HandleAPIError(w, r, apierrors.NewErrPassThroughToClient(err, http.StatusBadRequest))
+		return
+	}
+	if request.PreviewsWorkflowFilename != "" && request.DeleteWorkflowFilename != "" {
+		err := telemetry.Error(ctx, span, nil, "both preview and delete workflow filenames cannot be set")
+		c.HandleAPIError(w, r, apierrors.NewErrPassThroughToClient(err, http.StatusBadRequest))
+		return
+	}
+
 	client, err := getGithubClient(c.Config(), request.GithubAppInstallationID)
 	if err != nil {
 		err := telemetry.Error(ctx, span, err, "error creating github client")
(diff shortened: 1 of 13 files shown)


api/server/handlers/porter_app/create_secret_and_open_pr.go
```diff
@@ -100,23 +111,26 @@ func (c *OpenStackPRHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	var prRequestBody string
 	if request.DeleteWorkflowFilename == "" {
 		prRequestBody = "Hello 👋 from Porter! Please merge this PR to finish setting up your application."
-	} else {
+	} else if request.PreviewsWorkflowFilename == "" {
 		prRequestBody = "Please merge this PR to delete the workflow file associated with your application."
+	} else {
+		prRequestBody = "Hello 👋 from Porter! Please merge this PR to enable preview environments for your application."
```
ferozemohideen, line 117: we should probably pass this in directly from the request haha but not sure if you want to make that lift in this PR

dashboard/src/main/home/app-dashboard/new-app-flow/utils.ts
```diff
@@ -40,4 +51,41 @@ jobs:
...
+  stackName: string,
+  porterYamlPath: string = "porter.yaml"
+) => {
+  return `on:
+  pull_request:
+    types:
+    - opened
+    - synchronize
+    branches-ignore:
+    - porter-*
```
ferozemohideen, line 68: does this mean it will deploy when any pull request is opened (excluding porter branches)?

internal/integrations/ci/actions/stack.go
```diff
@@ -30,13 +31,21 @@ type GetStackApplyActionYAMLOpts struct {
...
+		return pr, fmt.Errorf("input options cannot be nil")
+	}
+
 	if opts.DeleteWorkflowFilename != "" {
 		prBranchName = "porter-stack-delete"
+	} else if opts.PreviewWorkflowFilename != "" {
+		prBranchName = "porter-stack-preview"
 	} else {
 		prBranchName = "porter-stack"
 	}
```
ferozemohideen, line 51: same here the branch name should probably come from the opts haha

internal/integrations/ci/actions/stack.go
```diff
@@ -120,6 +87,60 @@ func OpenGithubPR(opts *GithubPROpts) (*github.PullRequest, error) {
...
+		Preview:        preview,
+	})
+	if err != nil {
+		return err
+	}
+
+	filename := fmt.Sprintf("porter_stack_%s.yml", strings.ToLower(opts.StackName))
+	if preview {
+		filename = fmt.Sprintf("porter_preview_%s.yml", strings.ToLower(opts.StackName))
+	}
```
ferozemohideen, line 129: any way we can make filename part of opts? that way we don't need separate fields for preview vs non-preview, we can just have a filename opt and a create/delete bool opt

ferozemohideen: I guess we'll need a preview bool opt as well for the extra logic with that file

ferozemohideen (approved):

dashboard/src/main/home/app-dashboard/new-app-flow/utils.ts
ianedwards: Yeah, the workflow will always run, but we could disable auto deploy for certain apps if needed or have users enable it only for specific branch patterns:

```
 pull_request:
    types:
    - opened
    - synchronize
    branches:
    - preview/*
    branches-ignore:
    - porter-*
    ...
```

internal/integrations/ci/actions/stack.go
ianedwards: Did a quick refactor to handle some of these cases and determine the purpose of the PR upfront

internal/integrations/ci/actions/stack.go
ianedwards: handled

PR #3718, opened 2023-10-03
Add support for shimmed CNB buildpacks to porter backend

jose-fully-ported:
## What does this PR do?

The backend currently complains with the following when using a shimmed buildpack url:

```
error: error building app: error building image with pack: please provide either a github.com URL or a ZIP file URL
```

This PR adds support for those and implements some tests for buildpack name parsing.


cli/cmd/pack/pack.go
```diff
@@ -157,3 +90,93 @@ func (a *Agent) Build(ctx context.Context, opts *docker.BuildOpts, buildConfig *
...
 }
+
+func getBuildpackName(ctx context.Context, bp string) (string, error) {
+	if bp == "" {
+		return "", errors.New("please specify a buildpack name")
+	}
+
+	u, err := url.Parse(bp)
+	if err != nil {
+		return bp, nil
```
ferozemohideen, line 101: do we want to return "", err here?

ferozemohideen (approved):

cli/cmd/pack/pack.go
jose-fully-ported: Nah, the original code actually ignores errors if its not a url and uses the string as is.

PR #3725, opened 2023-10-04
POR-1851 delete preview deployment targets on pr close

ianedwards:
## What does this PR do?

- Adds webhook to Github repo to watch for any pull request and push events that occur
- On PR close, calls CCP to clean up any app revisions and delete the namespace corresponding to that branch

### DIFF
--- a/api/server/handlers/webhook/app_v2_github.go
+++ b/api/server/handlers/webhook/app_v2_github.go
@@ -1,0 +1,154 @@
+package webhook
+
+import (
+	"net/http"
+
+	"connectrpc.com/connect"
+	"github.com/google/go-github/v41/github"
+	"github.com/google/uuid"
+	porterv1 "github.com/porter-dev/api-contracts/generated/go/porter/v1"
+	"github.com/porter-dev/porter/api/server/authz"
+	"github.com/porter-dev/porter/api/server/handlers"
+	"github.com/porter-dev/porter/api/server/shared"
+	"github.com/porter-dev/porter/api/server/shared/apierrors"
+	"github.com/porter-dev/porter/api/server/shared/config"
+	"github.com/porter-dev/porter/api/server/shared/requestutils"
+	"github.com/porter-dev/porter/api/types"
+	"github.com/porter-dev/porter/internal/models"
+	"github.com/porter-dev/porter/internal/telemetry"
+)
+
+type GithubWebhookHandler struct {
+	handlers.PorterHandlerReadWriter
+	authz.KubernetesAgentGetter
+}
+
+func NewGithubWebhookHandler(
+	config *config.Config,
+	decoderValidator shared.RequestDecoderValidator,
+	writer shared.ResultWriter,
+) *GithubWebhookHandler {
+	return &GithubWebhookHandler{
+		PorterHandlerReadWriter: handlers.NewDefaultPorterHandler(config, decoderValidator, writer),
+		KubernetesAgentGetter:   authz.NewOutOfClusterAgentGetter(config),
+	}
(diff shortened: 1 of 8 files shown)


api/server/handlers/webhook/app_v2_github.go
```diff
@@ -0,0 +1,165 @@
...
+
+	projectID, reqErr := requestutils.GetURLParamUint(r, types.URLParamProjectID)
+	if reqErr != nil {
+		err := telemetry.Error(ctx, span, nil, "error parsing project id")
+		c.HandleAPIError(w, r, apierrors.NewErrPassThroughToClient(err, http.StatusBadRequest))
+		return
+	}
+
+	telemetry.WithAttributes(span, telemetry.AttributeKV{Key: "project-id", Value: projectID})
+	telemetry.WithAttributes(span, telemetry.AttributeKV{Key: "cluster-id", Value: clusterID})
```
ferozemohideen, line 84: don't need these

api/server/handlers/webhook/app_v2_github.go
```diff
@@ -0,0 +1,165 @@
...
+		c.HandleAPIError(w, r, apierrors.NewErrPassThroughToClient(err, http.StatusInternalServerError))
+		return
+	}
+
+	telemetry.WithAttributes(span, telemetry.AttributeKV{Key: "porter-app-id", Value: app.ID})
+
+	switch event := event.(type) {
+	case *github.PullRequestEvent:
+		if event.GetAction() != GithubPRStatus_Closed {
+			c.WriteResult(w, r, nil)
```
ferozemohideen, line 120: should we add any telemetry on cases when we write nothing and return?

internal/porter_app/github.go
```diff
@@ -0,0 +1,141 @@
...
+	defer span.End()
+
+	if githubAppSecret == nil {
+		return nil, telemetry.Error(ctx, span, nil, "github app secret is nil")
+	}
+	if githubAppID == "" {
+		return nil, telemetry.Error(ctx, span, nil, "github app id is empty")
+	}
+
+	appID, err := strconv.Atoi(githubAppID)
```
ferozemohideen, line 125: this method might return default 1 rather than an error, but not sure, might want to check

internal/porter_app/github.go
```diff
@@ -0,0 +1,141 @@
...
+		http.DefaultTransport,
+		int64(appID),
+		int64(repoID),
+		githubAppSecret,
+	)
+	if err != nil {
+		return nil, telemetry.Error(ctx, span, err, "could not create github app client")
+	}
+
+	return github.NewClient(&http.Client{Transport: itr}), nil
```
ferozemohideen, line 140: do we have to check if itr is nil?

ferozemohideen (approved):

api/server/handlers/webhook/app_v2_github.go
ianedwards: Since the route isn't scoped to project and cluster, I'm adding them in since I don't think it'll get added otherwise

api/server/handlers/webhook/app_v2_github.go
ianedwards: I think this could in general be noisy, since it'll be all PR events for the whole repo. I'll add some telemetry to indicate that we've processed the event though

internal/porter_app/github.go
ianedwards: I think checking for error should prevent using any defaults: https://pkg.go.dev/strconv#Atoi

PR #3731, opened 2023-10-04
Image Settings for Porter Apps V1

ferozemohideen:


dashboard/src/main/home/app-dashboard/app-view/tabs/ImageSettingsTab.tsx
```diff
@@ -36,42 +36,41 @@ const ImageSettingsTab: React.FC = () => {
 
     return (
         <>
-            {match(source)
-                .with({ type: "docker-registry" }, (source) => (
```
ianedwards, line 40: guessing we make this check elsewhere?

dashboard/src/main/home/app-dashboard/app-view/tabs/ImageSettingsTab.tsx
```diff
@@ -36,42 +36,41 @@ const ImageSettingsTab: React.FC = () => {
...
+            <Spacer y={1} />
+            <Button
+                type="submit"
+                status={buttonStatus}
+                disabled={
+                    isSubmitting
+                    || latestRevision.status === "CREATED"
+                    || latestRevision.status === "AWAITING_BUILD_ARTIFACT"
+                    || !image.repository
+                    || !image.tag
```
ianedwards, line 56: is image always defined?

ianedwards: either something I misconfigured or a bug in react-hook-form - I think `watch` return values mistakenly say they're defined when potentially they're not. Not sure if that's the case here

ianedwards (approved):

dashboard/src/main/home/app-dashboard/app-view/tabs/ImageSettingsTab.tsx
ferozemohideen: decided to add this back

dashboard/src/main/home/app-dashboard/app-view/tabs/ImageSettingsTab.tsx
ferozemohideen: added conditional

PR #3746, opened 2023-10-05
[POR-1877] [POR-1865] fix autodetection in v2, disable changing branch for existing apps 

ferozemohideen:


dashboard/src/main/home/app-dashboard/app-view/AppDataContainer.tsx
```diff
@@ -305,9 +305,9 @@ const AppDataContainer: React.FC<AppDataContainerProps> = ({ tabParam }) => {
         errorStackTrace: stack,
       });
 
-      if (err.response?.data?.message) {
+      if ((err as any).response?.data?.message) {
         setError("app", {
-          message: `App update failed: ${err.response.data.message}`,
+          message: `App update failed: ${(err as any).response.data.message}`,
```
ianedwards, line 310: is this any cast needed?

If it needs to have a type could also check `if (axios.isAxiosError(err))`

ianedwards (approved):

PR #3753, opened 2023-10-06
Report docker build logs on failure

ferozemohideen:
This is accomplished through the following steps:

1. create a file called PORTER_BUILD_LOGS.txt before the build starts
2. pass a multiWriter into the `jsonmessage.DisplayJSONMessagesStream` function, so that it writes to this file as well as os.Stderr as it does normally
3. after the build, read the log file, save it to a string, and delete the log file
4. encode the string and send it to analytics. The encoded string doesn't make it to the buildMetadata in the DB, because we pick it out before we save the metadata on the server side. Instead, it goes to segment (and telemetry and zapier)

Once this is in master, I will test through the cli-action-testing for both v1 and v2 builds, making sure they complete normally.

### DIFF
--- a/cli/cmd/docker/builder.go
+++ b/cli/cmd/docker/builder.go
@@ -59,11 +59,9 @@ func (a *Agent) BuildLocal(ctx context.Context, opts *BuildOpts) (err error) {
 		return err
 	}
 
-	var writer io.Writer
+	var writer io.Writer = os.Stderr
 	if opts.LogFile != nil {
 		writer = io.MultiWriter(os.Stderr, opts.LogFile)
-	} else {
-		writer = os.Stderr
 	}
 
 	if !opts.IsDockerfileInCtx {
--- a/cli/cmd/docker/builder.go
+++ b/cli/cmd/docker/builder.go
@@ -30,6 +30,8 @@ type BuildOpts struct {
 	UseCache          bool
 
 	Env map[string]string
+
+	LogFile *os.File
 }
 
 // BuildLocal
@@ -55,6 +57,13 @@ func (a *Agent) BuildLocal(ctx context.Context, opts *BuildOpts) (err error) {
 	})
 	if err != nil {
 		return err
+	}
+
+	var writer io.Writer
+	if opts.LogFile != nil {
+		writer = io.MultiWriter(os.Stderr, opts.LogFile)
+	} else {
+		writer = os.Stderr
 	}
 
 	if !opts.IsDockerfileInCtx {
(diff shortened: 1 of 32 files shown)


cli/cmd/docker/builder.go
```diff
@@ -57,6 +59,13 @@ func (a *Agent) BuildLocal(ctx context.Context, opts *BuildOpts) (err error) {
 		return err
 	}
 
+	var writer io.Writer
+	if opts.LogFile != nil {
+		writer = io.MultiWriter(os.Stderr, opts.LogFile)
+	} else {
+		writer = os.Stderr
+	}
+
```
d-g-town, line 68: writer = os.Stderr
if opts.LogFile != nil {
    writer = io.MultiWriter(os.Stderr, opts.LogFile)
}

cli/cmd/v2/apply.go
```diff
@@ -509,20 +510,25 @@ func updateEnvGroupsInProto(ctx context.Context, base64AppProto string, envGroup
...
+	_, err := client.UpdateRevisionStatus(ctx, cliConf.Project, cliConf.Cluster, appName, appRevisionID, models.AppRevisionStatus_BuildFailed)
+	if err != nil {
+		return err
+	}
+
 	buildMetadata := make(map[string]interface{})
 	buildMetadata["end_time"] = time.Now().UTC()
 
 	// the below is a temporary solution until we can report build errors via telemetry from the CLI
 	errorStringMap := make(map[string]string)
```
d-g-town, line 523: Why not make this into an object with fields BuildError and B64BuildLogs?

d-g-town (approved):

cli/cmd/v2/build.go
```diff
@@ -81,7 +98,16 @@ func build(ctx context.Context, client api.Client, inp buildInput) error {
...
 		)
 		if err != nil {
-			return fmt.Errorf("error resolving docker paths: %w", err)
+			output.Error = fmt.Errorf("error resolving docker paths: %w", err)
+			return output
+		}
+
+		var logFile *os.File
+		if inp.LogFilename != "" {
+			logFile, err = os.Create(inp.LogFilename)
```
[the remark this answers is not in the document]
ferozemohideen: changed to use temp file

cli/cmd/v2/apply.go
ferozemohideen: I would but then I'd have to touch the v1 code, since they both get parsed by createOrUpdateAppEvent. I'll refactor this later once the v1 migration is over

PR #3756, opened 2023-10-06
Update Porter Apps on Env Group Update V2

sdess09:
## What does this PR do?

Updates the app for a given v2 deployment

### DIFF
--- a/dashboard/src/main/home/cluster-dashboard/env-groups/ExpandedEnvGroup.tsx
+++ b/dashboard/src/main/home/cluster-dashboard/env-groups/ExpandedEnvGroup.tsx
@@ -624,7 +624,6 @@ export const ExpandedEnvGroupFC = ({
           );
           //const newEnvGroup = await pollForEnvGroup(name, currentProject, currentCluster);
           if (linkedApp) {
-            // Map through each linked application and return a Promise
             const promises = linkedApp.map(async appName => {
               if (!currentProject.validate_apply_v2) {
                 return getPorterApp({ appName: appName });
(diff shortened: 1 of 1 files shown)


dashboard/src/main/home/cluster-dashboard/env-groups/ExpandedEnvGroup.tsx
```diff
@@ -622,9 +624,56 @@ export const ExpandedEnvGroupFC = ({
           );
           //const newEnvGroup = await pollForEnvGroup(name, currentProject, currentCluster);
           if (linkedApp) {
-            for (let appName of linkedApp) {
-              await getPorterApp({ appName: appName });
-            }
+            const promises = linkedApp.map(async appName => {
+              if (!currentProject.validate_apply_v2) {
+                return getPorterApp({ appName: appName });
```
ianedwards, line 629: is this call needed or could we just return a void promise?

dashboard/src/main/home/cluster-dashboard/env-groups/ExpandedEnvGroup.tsx
```diff
@@ -622,9 +624,56 @@ export const ExpandedEnvGroupFC = ({
...
+                    {
+                      deployment_target_id: currentDeploymentTarget.id,
+                    },
+                    {
+                      project_id: currentProject.id,
+                      cluster_id: currentCluster.id,
+                      porter_app_name: appName,
+                    }
+                  )
+                  if (res) {
```
ianedwards, line 643: I think there are some exported validators that could be used to type this

dashboard/src/main/home/cluster-dashboard/env-groups/ExpandedEnvGroup.tsx
```diff
@@ -622,9 +624,56 @@ export const ExpandedEnvGroupFC = ({
...
+                      {
+                        project_id: currentProject.id,
+                        cluster_id: currentCluster.id,
+                      }
+                    );
+                  }
+                } catch (err) {
+
+                  setCurrentError(error);
+                }
```
ianedwards, line 673: since this is being passed to Promise.all, you could remove the try catch here and move that to handle the Promise.all instead

ianedwards (approved):

PR #3759, opened 2023-10-06
POR-1852 comment deploy details with URL on first preview deploy

ianedwards:
## What does this PR do?

- Bot comments on PR deploys depending on status of the revision
- Update Preview env url to use depl target ID instead of branch

### DIFF
--- a/api/server/handlers/porter_app/report_status.go
+++ b/api/server/handlers/porter_app/report_status.go
@@ -1,0 +1,258 @@
+package porter_app
+
+import (
+	"context"
+	"encoding/base64"
+	"fmt"
+	"net/http"
+	"strings"
+
+	"github.com/google/go-github/v39/github"
+	"github.com/google/uuid"
+	"github.com/porter-dev/api-contracts/generated/go/helpers"
+	porterv1 "github.com/porter-dev/api-contracts/generated/go/porter/v1"
+	"github.com/porter-dev/porter/api/server/handlers"
+	"github.com/porter-dev/porter/api/server/shared"
+	"github.com/porter-dev/porter/api/server/shared/apierrors"
+	"github.com/porter-dev/porter/api/server/shared/config"
+	"github.com/porter-dev/porter/api/server/shared/requestutils"
+	"github.com/porter-dev/porter/api/types"
+	"github.com/porter-dev/porter/internal/deployment_target"
+	"github.com/porter-dev/porter/internal/models"
+	"github.com/porter-dev/porter/internal/porter_app"
+	"github.com/porter-dev/porter/internal/telemetry"
+	"k8s.io/utils/pointer"
+)
+
+// ReportRevisionStatusHandler is the handler for the /apps/{porter_app_name}/revisions/{app_revision_id}/status endpoint
+type ReportRevisionStatusHandler struct {
+	handlers.PorterHandlerReadWriter
+}
+
+// NewReportRevisionStatusHandler handles POST requests to the endpoint /apps/{porter_app_name}/revisions/{app_revision_id}/status
+func NewReportRevisionStatusHandler(
+	config *config.Config,
+	decoderValidator shared.RequestDecoderValidator,
+	writer shared.ResultWriter,
+) *ReportRevisionStatusHandler {
+	return &ReportRevisionStatusHandler{
+		PorterHandlerReadWriter: handlers.NewDefaultPorterHandler(config, decoderValidator, writer),
+	}
+}
+
+// ReportRevisionStatusRequest is the request object for the /apps/{porter_app_name}/revisions/{app_revision_id}/status endpoint
+type ReportRevisionStatusRequest struct {
+	PRNumber  int    `json:"pr_number"`
+	CommitSHA string `json:"commit_sha"`
+}
+
+type ReportRevisionStatusResponse struct{}
+
+// ServeHTTP reports the status of a revision to Github and other integrations, depending on the status and the deployment target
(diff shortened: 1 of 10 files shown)


api/server/handlers/porter_app/report_status.go
```diff
@@ -0,0 +1,259 @@
...
+func (c *ReportRevisionStatusHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
+	ctx, span := telemetry.NewSpan(r.Context(), "serve-report-revision-status")
+	defer span.End()
+
+	project, _ := ctx.Value(types.ProjectScope).(*models.Project)
+	cluster, _ := ctx.Value(types.ClusterScope).(*models.Cluster)
+
+	if !project.GetFeatureFlag(models.ValidateApplyV2, c.Config().LaunchDarklyClient) {
+		err := telemetry.Error(ctx, span, nil, "project does not have validate apply v2 enabled")
+		c.HandleAPIError(w, r, apierrors.NewErrForbidden(err))
```
ferozemohideen, line 62: i prefer to pass through to client with status code forbidden, otherwise you won't see the message

api/server/handlers/porter_app/report_status.go
```diff
@@ -0,0 +1,259 @@
...
+		err := telemetry.Error(ctx, span, nil, "app revision id is nil")
+		c.HandleAPIError(w, r, apierrors.NewErrPassThroughToClient(err, http.StatusBadRequest))
+		return
+	}
+	telemetry.WithAttributes(span, telemetry.AttributeKV{Key: "app-revision-id", Value: appRevisionUuid.String()})
+
+	porterApp, err := c.Repo().PorterApp().ReadPorterAppByName(cluster.ID, appName)
+	if err != nil {
+		err := telemetry.Error(ctx, span, err, "error reading porter app by name")
+		c.HandleAPIError(w, r, apierrors.NewErrInternal(err))
```
ferozemohideen, line 98: any reason this is internal? a few others in this file are also internal

api/server/handlers/porter_app/report_status.go
```diff
@@ -0,0 +1,259 @@
...
+		return telemetry.Error(ctx, span, err, "error unmarshalling app proto")
+	}
+
+	body := "## Porter Preview Environments\n"
+
+	switch inp.revision.Status {
+	case models.AppRevisionStatus_BuildFailed:
+		body = fmt.Sprintf("%s❌ The latest deploy failed to build. Check the [action logs](https://github.com/%s/actions/runs/) for more information.", body, inp.porterApp.RepoName)
+	case models.AppRevisionStatus_DeployFailed:
+		body = fmt.Sprintf("%s❌ The latest SHA ([`%s`](https://github.com/%s/%s/commit/%s)) failed to deploy.", body, inp.commitSha, repoDetails[0], repoDetails[1], inp.commitSha)
```
ferozemohideen, line 227: any mitigation steps here? link to dashboard or even contact support?

api/server/handlers/porter_app/report_status.go
```diff
@@ -0,0 +1,259 @@
...
+	case models.AppRevisionStatus_Deployed:
+		body = fmt.Sprintf("%s✅ The latest SHA ([`%s`](https://github.com/%s/%s/commit/%s)) has been successfully deployed.", body, inp.commitSha, repoDetails[0], repoDetails[1], inp.commitSha)
```
ferozemohideen, line 229: might want to link to dashboard here as well but up to you

api/server/handlers/porter_app/report_status.go
```diff
@@ -0,0 +1,259 @@
...
+		return nil
+	}
+
+	for _, service := range appProto.Services {
+		webConfig := service.GetWebConfig()
+		if webConfig != nil {
+			domains := webConfig.GetDomains()
+
+			if len(domains) > 0 {
+				body = fmt.Sprintf("%s\n\n**Preview URL**: https://%s", body, domains[0].Name)
```
ferozemohideen, line 240: should we be checking if they have custom domains set up? if so, this will only show the porter domain, right? also why not display all domains?

dashboard/src/main/home/app-dashboard/new-app-flow/utils.ts
```diff
@@ -92,5 +92,6 @@ jobs:
         PORTER_PROJECT: ${projectID}
         PORTER_STACK_NAME: ${stackName}
         PORTER_TAG: \${{ steps.vars.outputs.sha_short }}
-        PORTER_TOKEN: \${{ secrets.PORTER_STACK_${projectID}_${clusterId} }}`;
+        PORTER_TOKEN: \${{ secrets.PORTER_STACK_${projectID}_${clusterId} }}
+        PORTER_PR_ID: \${{ github.event.inputs.pr_number }}`;
```
ferozemohideen, line 96: super nit: would be clearer if it was either PORTER_PR_ID and inputs.pr_id or PORTER_PR_NUMBER and inputs.pr_number

ferozemohideen (approved):

api/server/handlers/porter_app/report_status.go
ianedwards: this was unintended, switching all these now

api/server/handlers/porter_app/report_status.go
ianedwards: Thinking it makes sense in a follow up PR to disable custom domains for now. We need a way to ensure uniqueness and there isn't an automated way to do that at the moment

PR #3765, opened 2023-10-08
custom nginx ingress annotations on web svcs

ianedwards:

### DIFF
--- a/dashboard/src/lib/porter-apps/services.ts
+++ b/dashboard/src/lib/porter-apps/services.ts
@@ -19,6 +19,7 @@ import {
   serviceNumberValidator,
   serviceStringValidator,
 } from "./values";
+import _ from "lodash";
 
 export type DetectedServices = {
   services: ClientService[];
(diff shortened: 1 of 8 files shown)


dashboard/src/lib/porter-apps/services.ts
```diff
@@ -348,6 +358,17 @@ export function deserializeService({
...
+            (annotation) => {
+              const overrideAnnotation =
+                overrideWebConfig?.ingressAnnotations[annotation[0]];
+              return {
+                key: annotation[0],
+                value: overrideAnnotation ?? annotation[1],
+                readOnly: !!overrideAnnotation,
+              };
+            }
+          ),
```
ferozemohideen, line 371: if override has ingress map with keys that aren't in the service ingress map, will they be lost?

dashboard/src/lib/porter-apps/services.ts
```diff
@@ -225,6 +229,12 @@ export function serializeService(service: ClientService): SerializedService {
           domains: config.domains.map((domain) => ({
             name: domain.name.value,
           })),
+          ingressAnnotations: Object.fromEntries(
+            config.ingressAnnotations.map((annotation) => [
+              annotation.key,
+              annotation.value,
+            ])
+          ),
```
ferozemohideen, line 237: can we add quotation marks to the values here if they don't exist already? that's burned us in the past because the helm install will fail or something

ferozemohideen (approved):

dashboard/src/lib/porter-apps/services.ts
ianedwards: good call

PR #3772, opened 2023-10-09
Porter App V2 Revision Diff

ferozemohideen:
Adds a handler to generate a porter.yaml file given the app proto. I think this will miss out on env changes, but we can handle that later

image

image


dashboard/src/main/home/app-dashboard/app-view/tabs/activity-feed/events/modals/RevisionDiffModal.tsx
```diff
@@ -0,0 +1,135 @@
...
+            refetchOnWindowFocus: false,
+        }
+    );
+    useEffect(() => {
+        if (data) {
+            setBaseYamlString(data.base);
+            setChangedYamlString(data.changed);
+        }
+    }, [data]);
+
```
ianedwards, line 79: out of curiosity - did type narrowing on `status` to guarantee that data is defined not work?

internal/porter_app/v2/yaml.go
```diff
@@ -305,3 +305,130 @@ func serviceProtoFromConfig(service Service, serviceType porterv1.ServiceType) (
 
 	return serviceProto, nil
 }
+
+// AppFromProto converts a PorterApp proto object into a PorterApp struct
+func AppFromProto(appProto *porterv1.PorterApp) (PorterApp, error) {
```
ianedwards, line 310: annoying question - what are your thoughts on putting this on CCP? Reason I ask is because I believe this method is already defined there and we can avoid keeping them in sync.

ianedwards: I think it's slightly different, so if these types are more useful on the porter side then that can be reason enough not to

internal/porter_app/v2/yaml.go
```diff
@@ -305,3 +305,130 @@ func serviceProtoFromConfig(service Service, serviceType porterv1.ServiceType) (
...
+				MinInstances:           int(workerConfig.Autoscaling.MinInstances),
+				MaxInstances:           int(workerConfig.Autoscaling.MaxInstances),
+				CpuThresholdPercent:    int(workerConfig.Autoscaling.CpuThresholdPercent),
+				MemoryThresholdPercent: int(workerConfig.Autoscaling.MemoryThresholdPercent),
+			}
+		}
+		appService.Autoscaling = autoscaling
+	case porterv1.ServiceType_SERVICE_TYPE_JOB:
+		jobConfig := service.GetJobConfig()
+		appService.Type = "job"
```
ianedwards, line 427: potentially use consts on these types

ianedwards (approved):

internal/porter_app/v2/yaml.go
ferozemohideen: I thought about that, but decided to keep it in Porter since the logic to go from yaml to proto is also there. I'll see what @d-g-town has to say though

d-g-town: I think it makes sense to have separate methods that deal with going to and from the Porter YAML object (we could image this being different from the CCP object in slight ways)

d-g-town: You might also consider a test that confirms that going to and back results in the same object

dashboard/src/main/home/app-dashboard/app-view/tabs/activity-feed/events/modals/RevisionDiffModal.tsx
ferozemohideen: oh interesting, didn't know that would work - updated

PR #3774, opened 2023-10-09
POR-1902 validate name against existing porter apps

ianedwards:
## What does this PR do?

- Client and server side validation for app names
- Ensures that an app name cannot be used twice in the same project
- Also makes sure that name is valid (lowercase characters, numbers, and "-")

### DIFF
--- a/api/server/handlers/porter_app/validate.go
+++ b/api/server/handlers/porter_app/validate.go
@@ -112,10 +112,24 @@ func (c *ValidatePorterAppHandler) ServeHTTP(w http.ResponseWriter, r *http.Requ
 		}
 	}
 
+	existingApps, err := c.Repo().PorterApp().ListPorterAppsByProjectID(project.ID)
+	if err != nil {
+		err := telemetry.Error(ctx, span, err, "error listing porter apps by project id")
+		c.HandleAPIError(w, r, apierrors.NewErrPassThroughToClient(err, http.StatusInternalServerError))
+		return
+	}
+
 	if appProto.Name == "" {
 		err := telemetry.Error(ctx, span, nil, "app proto name is empty")
 		c.HandleAPIError(w, r, apierrors.NewErrPassThroughToClient(err, http.StatusBadRequest))
 		return
+	}
+	for _, existingApp := range existingApps {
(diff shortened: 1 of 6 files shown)


api/server/handlers/porter_app/validate.go
```diff
@@ -112,11 +112,25 @@ func (c *ValidatePorterAppHandler) ServeHTTP(w http.ResponseWriter, r *http.Requ
...
+		c.HandleAPIError(w, r, apierrors.NewErrPassThroughToClient(err, http.StatusInternalServerError))
+		return
+	}
+
 	if appProto.Name == "" {
 		err := telemetry.Error(ctx, span, nil, "app proto name is empty")
 		c.HandleAPIError(w, r, apierrors.NewErrPassThroughToClient(err, http.StatusBadRequest))
 		return
 	}
+	for _, existingApp := range existingApps {
```
ferozemohideen, line 127: let's log the app name to telemetry before this check

dashboard/src/main/home/app-dashboard/create-app/CreateApp.tsx
```diff
@@ -481,6 +481,8 @@ const CreateApp: React.FC<CreateAppProps> = ({ history }) => {
     }
   }, [porterApps, name.value]);
 
+  console.log("errors", errors);
```
ferozemohideen, line 484: remove

internal/repository/gorm/porter_app.go
```diff
@@ -34,6 +34,16 @@ func (repo *PorterAppRepository) ListPorterAppByClusterID(clusterID uint) ([]*mo
 	return apps, nil
 }
 
+func (repo *PorterAppRepository) ListPorterAppsByProjectID(projectID uint) ([]*models.PorterApp, error) {
```
ferozemohideen, line 37: does this need to check deployment target id?

ferozemohideen (approved):

internal/repository/gorm/porter_app.go
ianedwards: An app from the same project can be deployed to multiple deployment targets. It just has to be unique across the project

PR #3780, opened 2023-10-09
env var tab improvements

ferozemohideen:
image

image


ianedwards (approved):

dashboard/src/main/home/cluster-dashboard/env-groups/EnvGroupArrayV2.tsx
```diff
@@ -43,13 +40,26 @@ const EnvGroupArrayV2 = ({
...
+    const _values = [...values];
+    _values[i].key = e.target.value;
+    setValues(_values);
+  };
+
+  const handleValueChange = (values: KeyValueType[], i: number, e: React.ChangeEvent<HTMLInputElement>) => {
+    const _values = [...values];
+    _values[i].value = e.target.value;
+    setValues(_values);
+  };
```
ianedwards, line 61: might be worth making this into a hook form or something at some point so we don't have to update the whole thing in case it's a really big env group

dashboard/src/main/home/cluster-dashboard/env-groups/EnvGroupArrayV2.tsx
```diff
@@ -89,142 +99,160 @@ const EnvGroupArrayV2 = ({
...
-                    disabled={disabled || entry.locked}
-                    spellCheck={false}
-                    override={isKeyOverriding(entry.key)}
-                  />
-                  < Spacer x={.5} inline />
-                  {entry.hidden ? (
-                    entry.value?.includes("PORTERSECRET") ? (
+                  {entry.locked ? (
+                    <Tooltip
+                      content={"This key is part of a secret and cannot be edited"}
```
d-g-town, line 108: Maybe `To edit this secret, delete and recreate the key with your new value.`

dashboard/src/main/home/app-dashboard/validate-apply/app-settings/EnvVarRow.tsx
```diff
@@ -0,0 +1,263 @@
...
+                        width="270px"
+                        value={entry.key}
+                        disabled
+                        spellCheck={false}
+                        override={isKeyOverriding(entry.key)}
+                    />
+                </Tooltip>
+            ) : (
+                <Controller
+                    name={`app.env.${index}.key`}
```
ianedwards, line 43: Not sure if you tried this but wondering if it'd be possible to just pass the update fn from `useFieldArray` instead of using the Controller. Not sure if there's potentially some state issues that could come up

ianedwards (approved):

PR #3785, opened 2023-10-10
POR-1874 cancel pending workflows when pr is closed or merged

ianedwards:
## What does this PR do?

- If someone were to quickly close a PR or merged before a preview deploy finished, currently we'd continue to build and create a new revision
- This is also error prone as we delete the deployment target / a new deployment target will potentially be created, or applying the new revision will just error out

### DIFF
--- a/internal/repository/porter_app.go
+++ b/internal/repository/porter_app.go
@@ -1,11 +1,14 @@
 package repository
 
 import (
+	"context"
+
 	"github.com/porter-dev/porter/internal/models"
 )
 
 // PorterAppRepository represents the set of queries on the PorterApp model
 type PorterAppRepository interface {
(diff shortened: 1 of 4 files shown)


internal/repository/porter_app.go
```diff
@@ -1,11 +1,14 @@
...
 
 import (
+	"context"
+
 	"github.com/porter-dev/porter/internal/models"
 )
 
 // PorterAppRepository represents the set of queries on the PorterApp model
 type PorterAppRepository interface {
+	ReadPorterApp(ctx context.Context, id uint) (*models.PorterApp, error)
```
ferozemohideen, line 11: nit: ReadPorterAppById?

ferozemohideen (approved):

PR #3789, opened 2023-10-10
[POR-1919] Make cluster resources a context provider so we can add services with default ram/cpu depending on the nodes

ferozemohideen:

### DIFF
--- a/dashboard/src/shared/ClusterResourcesContext.tsx
+++ b/dashboard/src/shared/ClusterResourcesContext.tsx
@@ -16,7 +16,7 @@ export const ClusterResourcesContext = createContext<{
   
 export const useClusterResources = () => {
     const context = useContext(ClusterResourcesContext);
-    if (context === null) {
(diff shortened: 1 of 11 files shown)


dashboard/src/shared/ClusterResourcesContext.tsx
```diff
@@ -0,0 +1,51 @@
...
+};
+  
+const ClusterResourcesProvider = ({ children }: { children: JSX.Element }) => {
+    const { currentCluster, currentProject } = useContext(Context);
+
+    const { maxCPU, maxRAM, defaultCPU, defaultRAM } = useClusterResourceLimits({
+        projectId: currentProject?.id,
+        clusterId: currentCluster?.id,
+    });
+
```
ianedwards, line 34: any need to wait for the network call to finish to know if these values are set correctly?

ianedwards (approved):

PR #3797, opened 2023-10-11
POR-1923 setup dedicated revision rollback endpoint

ianedwards:
## What does this PR do?

- Exposes /apps/{app_name}/rollback, which rolls back to the previous successful revision or the app revision with a given ID if provided
- Changes CLI to call this endpoint

### DIFF
--- a/api/server/handlers/porter_app/rollback_revision.go
+++ b/api/server/handlers/porter_app/rollback_revision.go
@@ -1,0 +1,287 @@
+package porter_app
+
+import (
+	"context"
+	"fmt"
+	"net/http"
+
+	"connectrpc.com/connect"
+	"github.com/google/uuid"
+	porterv1 "github.com/porter-dev/api-contracts/generated/go/porter/v1"
+	"github.com/porter-dev/api-contracts/generated/go/porter/v1/porterv1connect"
+	"github.com/porter-dev/porter/api/server/authz"
+	"github.com/porter-dev/porter/api/server/handlers"
+	"github.com/porter-dev/porter/api/server/shared"
+	"github.com/porter-dev/porter/api/server/shared/apierrors"
+	"github.com/porter-dev/porter/api/server/shared/config"
+	"github.com/porter-dev/porter/api/server/shared/requestutils"
+	"github.com/porter-dev/porter/api/types"
+	"github.com/porter-dev/porter/internal/models"
+	"github.com/porter-dev/porter/internal/telemetry"
+)
+
+type RollbackAppRevisionHandler struct {
+	handlers.PorterHandlerReadWriter
+	authz.KubernetesAgentGetter
+}
+
+func NewRollbackAppRevisionHandler(
+	config *config.Config,
+	decoderValidator shared.RequestDecoderValidator,
+	writer shared.ResultWriter,
+) *RollbackAppRevisionHandler {
+	return &RollbackAppRevisionHandler{
+		PorterHandlerReadWriter: handlers.NewDefaultPorterHandler(config, decoderValidator, writer),
+		KubernetesAgentGetter:   authz.NewOutOfClusterAgentGetter(config),
+	}
+}
+
+type RollbackAppRevisionRequest struct {
+	DeploymentTargetID string `json:"deployment_target_id"`
+	AppRevisionID      string `json:"app_revision_id"`
+}
+
+type RollbackAppRevisionResponse struct {
+	AppRevisionID string `json:"app_revision_id"`
+}
+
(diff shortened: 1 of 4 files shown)


api/server/handlers/porter_app/rollback_revision.go
```diff
@@ -0,0 +1,287 @@
...
+	ctx, span := telemetry.NewSpan(r.Context(), "serve-rollback-app-revision")
+	defer span.End()
+
+	project, _ := ctx.Value(types.ProjectScope).(*models.Project)
+	cluster, _ := ctx.Value(types.ClusterScope).(*models.Cluster)
+
+	telemetry.WithAttributes(span,
+		telemetry.AttributeKV{Key: "project-id", Value: project.ID},
+		telemetry.AttributeKV{Key: "cluster-id", Value: cluster.ID},
+	)
```
ferozemohideen, line 58: do we need this?

api/server/handlers/porter_app/rollback_revision.go
```diff
@@ -0,0 +1,287 @@
...
+
+	revisions := latestAppRevisionsResp.Msg.AppRevisions
+	if len(revisions) == 0 {
+		return proto, telemetry.Error(ctx, span, nil, "no revisions found for app")
+	}
+
+	skip := 0
+	if revisions[0].RevisionNumber != 0 {
+		skip = 1
+	}
```
ferozemohideen, line 271: im confused by the skip stuff can you add comments?

cli/cmd/v2/rollback.go
```diff
@@ -30,38 +27,13 @@ func Rollback(ctx context.Context, inp RollbackInput) error {
...
-		return fmt.Errorf("error calling apply endpoint: %w", err)
-	}
-
-	if applyResp.CLIAction != porterv1.EnumCLIAction_ENUM_CLI_ACTION_NONE {
-		return fmt.Errorf("unexpected CLI action: %s", applyResp.CLIAction)
+		return fmt.Errorf("error calling rollback revision endpoint: %w", err)
 	}
 
-	color.New(color.FgGreen).Printf("Successfully rolled back to revision %d\n", rollbackTarget.RevisionNumber) // nolint:errcheck,gosec
+	color.New(color.FgGreen).Printf("Successfully applied revision %s\n", rollbackResp.AppRevisionID) // nolint:errcheck,gosec
```
ferozemohideen, line 37: does it make sense to also/instead return the revision number?

ferozemohideen (approved):

api/server/handlers/porter_app/rollback_revision.go
stefanmcshane: not anymore on porter, but yes on CCP

api/server/handlers/porter_app/rollback_revision.go
```diff
@@ -0,0 +1,292 @@
...
+			projectID:     int64(project.ID),
+			appRevisionID: request.AppRevisionID,
+			ccpClient:     c.Config().ClusterControlPlaneClient,
+		})
+		if err != nil {
+			err = telemetry.Error(ctx, span, err, "error getting revision proto")
+			c.HandleAPIError(w, r, apierrors.NewErrPassThroughToClient(err, http.StatusInternalServerError))
+			return
+		}
+	} else {
```
stefanmcshane, line 122: is this else actually needed? Why not get the last deployed proto, and if that returns as empty get the revision proto?

api/server/handlers/porter_app/rollback_revision.go
```diff
@@ -0,0 +1,292 @@
...
+	})
+}
+
+type getRevisionProtoInput struct {
+	projectID     int64
+	appRevisionID string
+	ccpClient     porterv1connect.ClusterControlPlaneServiceClient
+}
+
+func getRevisionProto(ctx context.Context, inp getRevisionProtoInput) (*porterv1.PorterApp, error) {
```
stefanmcshane: To paraphrase Rob Pike's words, which I agree with - if you need words like "get", "do" etc in your function name, then you need a better function name.

api/server/handlers/porter_app/rollback_revision.go
stefanmcshane: Agreed. Alternatively add a comment to the function on the intention of the code, and change `skip` to something meaningful

cli/cmd/v2/rollback.go
ianedwards: changed to show number instead, agreed that's clearer

api/server/handlers/porter_app/rollback_revision.go
ianedwards: updated to be more clear

PR #3803, opened 2023-10-12
add telemetry to re-run github workflow

ferozemohideen:

### DIFF
--- a/api/server/handlers/gitinstallation/rerun_workflow.go
+++ b/api/server/handlers/gitinstallation/rerun_workflow.go
@@ -11,6 +11,9 @@ import (
 	"github.com/porter-dev/porter/api/server/shared/apierrors"
 	"github.com/porter-dev/porter/api/server/shared/commonutils"
 	"github.com/porter-dev/porter/api/server/shared/config"
+	"github.com/porter-dev/porter/api/server/shared/requestutils"
+	"github.com/porter-dev/porter/api/types"
+	"github.com/porter-dev/porter/internal/telemetry"
 )
 
 type RerunWorkflowHandler struct {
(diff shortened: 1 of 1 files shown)


api/server/handlers/gitinstallation/rerun_workflow.go
```diff
@@ -59,45 +81,53 @@ func (c *RerunWorkflowHandler) ServeHTTP(w http.ResponseWriter, r *http.Request)
...
+		c.HandleAPIError(w, r, apierrors.NewErrPassThroughToClient(err, http.StatusInternalServerError))
 		return
 	}
 
 	latestWorkflowRun, err := commonutils.GetLatestWorkflowRun(client, owner, name, filename, branch)
-
 	if err != nil && errors.Is(err, commonutils.ErrNoWorkflowRuns) {
-		c.HandleAPIError(w, r, apierrors.NewErrPassThroughToClient(err, 400))
+		err = telemetry.Error(ctx, span, err, "no workflow runs found")
+		c.HandleAPIError(w, r, apierrors.NewErrPassThroughToClient(err, http.StatusBadRequest))
```
ferozemohideen, line 96: this should probably be a 404 but didn't want to change the status code in case clients were relying on it

d-g-town (approved):

api/server/handlers/gitinstallation/rerun_workflow.go
stefanmcshane: If a repo and workflow file are found, but the runs are not found, that shouldnt be an error. Not found is valid, not errorful

stefanmcshane (approved):

PR #3811, opened 2023-10-13
[POR-1928] fix job history page

ferozemohideen:
image


dashboard/src/components/OldTable.tsx
```diff
@@ -247,7 +247,7 @@ const Table: React.FC<TableProps> = ({
               {"<"}
             </PaginationAction>
             <PageCounter>
-              {currentPageIndex + 1} of {pageCount}
+              {currentPageIndex + 1} of {pageCount + 1}
```
d-g-town, line 250: I think you only want to do + 1 if it's zero, right?

d-g-town (approved):

PR #3818, opened 2023-10-16
define services as list

ianedwards:
## What does this PR do?

- Consolidates conversion between porter app proto type and business layer type
- Uses that function where needed to handle updates to the porter app from the API, and to retrieve unique list of services between service map and service array
- Updates parsing logic to expect list of services instead of map
- Utilizes updated serviceList field on porter app proto type


api/server/handlers/gitinstallation/get_porter_yaml.go
```diff
@@ -94,26 +93,25 @@ func (c *GithubGetPorterYamlHandler) ServeHTTP(w http.ResponseWriter, r *http.Re
...
 			return
 		}
 	}
 
 	// backwards compatibility so that old porter yamls are no longer valid
-	if !project.GetFeatureFlag(models.ValidateApplyV2, c.Config().LaunchDarklyClient) && parsed.Version != nil {
-		version := *parsed.Version
-		if version != "v1stack" {
+	if !project.GetFeatureFlag(models.ValidateApplyV2, c.Config().LaunchDarklyClient) {
+		if version.Version != "" && version.Version != "v2" {
```
ferozemohideen, line 114: v1 porteryaml can be "" or "v1stack" versioned

api/server/handlers/porter_app/apply.go
```diff
@@ -210,16 +225,19 @@ func (c *ApplyPorterAppHandler) ServeHTTP(w http.ResponseWriter, r *http.Request
...
+	ctx, span := telemetry.NewSpan(ctx, "add-porter-subdomains-if-necessary")
+	defer span.End()
 
-			webConfig := service.GetWebConfig()
+	services := make([]v2.Service, len(app.Services))
 
-			if !webConfig.GetPrivate() && len(webConfig.Domains) == 0 {
+	for _, service := range app.Services {
+		if service.Type == v2.ServiceType_Web {
+			if service.Private == nil || !*service.Private {
```
ferozemohideen, line 236: will a service get lost if it's private?

dashboard/src/lib/porter-apps/services.ts
```diff
@@ -142,6 +145,24 @@ export function prefixSubdomain(subdomain: string) {
...
+  );
+
+  const uniqueServices = _.uniqBy(
+    [...app.serviceList, ...servicesFromMap],
+    (service) => service.name
+  );
+
+  return uniqueServices;
+}
+
```
ferozemohideen, line 165: this might lead to some weird behavior, is there a case where both serviceList and services will be defined for an app? I feel like we should either take everything from serviceList or everything from services (preferring the former) rather than possibly mixing

internal/porter_app/v2/yaml.go
```diff
@@ -40,15 +40,20 @@ func AppProtoFromYaml(ctx context.Context, porterYamlBytes []byte) (AppWithPrevi
 		return out, telemetry.Error(ctx, span, err, "error unmarshaling porter yaml")
 	}
 
-	appProto, envVariables, err := buildAppProto(ctx, porterYaml.PorterApp)
+	fmt.Printf("porter yaml: %+v\n", porterYaml)
+
+	fmt.Printf("call loc: yaml.go #1")
```
ferozemohideen, line 45: remove

internal/porter_app/v2/yaml.go
```diff
@@ -40,15 +40,20 @@ func AppProtoFromYaml(ctx context.Context, porterYamlBytes []byte) (AppWithPrevi
...
 	if err != nil {
 		return out, telemetry.Error(ctx, span, err, "error converting porter yaml to proto")
 	}
 	out.AppProto = appProto
 	out.EnvVariables = envVariables
 
 	if porterYaml.Previews != nil {
-		previewAppProto, previewEnvVariables, err := buildAppProto(ctx, *porterYaml.Previews)
+		fmt.Printf("call loc: yaml.go #2")
+		fmt.Printf("handling previews: %+v\n", *porterYaml.Previews)
```
ferozemohideen, line 55: remove

internal/porter_app/v2/yaml.go
```diff
@@ -159,22 +181,24 @@ func buildAppProto(ctx context.Context, porterApp PorterApp) (*porterv1.PorterAp
 		}
 	}
 
+	fmt.Printf("porter app services: %+v\n%+v\n", porterApp.Services, porterApp.Services == nil)
```
ferozemohideen, line 184: remove

internal/porter_app/v2/yaml.go
```diff
@@ -87,6 +105,7 @@ type Build struct {
 	Builder    string   `yaml:"builder" validate:"required_if=Method pack"`
 	Buildpacks []string `yaml:"buildpacks"`
 	Dockerfile string   `yaml:"dockerfile" validate:"required_if=Method docker"`
+	CommitSHA  string   `yaml:"commitSha"`
```
ferozemohideen, line 108: is this for preview envs?

internal/porter_app/v2/yaml.go
```diff
@@ -339,14 +366,14 @@ func AppFromProto(appProto *porterv1.PorterApp) (PorterApp, error) {
 		}
 	}
 
-	porterApp.Services = make(map[string]Service, 0)
-	for name, service := range appProto.Services {
+	uniqueServices := uniqueServices(appProto.Services, appProto.ServiceList)
+	fmt.Printf("unique services: %+v\n", uniqueServices)
```
ferozemohideen, line 370: remove

internal/porter_app/v2/yaml.go
```diff
@@ -358,16 +385,22 @@ func AppFromProto(appProto *porterv1.PorterApp) (PorterApp, error) {
...
+	porterApp.EnvGroups = make([]EnvGroup, 0)
 	for _, envGroup := range appProto.EnvGroups {
-		porterApp.EnvGroups = append(porterApp.EnvGroups, envGroup.Name)
+		porterApp.EnvGroups = append(porterApp.EnvGroups, EnvGroup{
+			Name:    envGroup.Name,
+			Version: int(envGroup.Version),
+		})
 	}
 
+	fmt.Printf("porter app: %+v\n", porterApp)
```
ferozemohideen, line 396: remove

internal/porter_app/v2/yaml.go
```diff
@@ -444,3 +477,23 @@ func appServiceFromProto(service *porterv1.Service) (Service, error) {
...
+		mergedServices[service.Name] = service
+	}
+
+	mergedServiceList := make([]*porterv1.Service, 0, len(mergedServices))
+	for _, service := range mergedServices {
+		mergedServiceList = append(mergedServiceList, service)
+	}
+
+	return mergedServiceList
+}
```
ferozemohideen, line 499: same concern here I had with the front-end merge method

ferozemohideen (changes requested):

api/server/handlers/porter_app/apply.go
ianedwards: It shouldn't, append is after the condition

ianedwards: can maybe edit in place to make this more clear

dashboard/src/lib/porter-apps/services.ts
ianedwards: This method should only come into play on the latest revision before this change goes out, so I don't think that issue should ever come up.

With that being said. I think I can safely just take one or the other for backwards compatibility.

internal/porter_app/v2/yaml.go
ianedwards: No this just wasn't included in the build type and was being left out when converting to and from the proto object. I'm trying to avoid passing the proto around so need to make sure that business types reflect full functionality

ferozemohideen (approved):

PR #3819, opened 2023-10-17
Replace metrics filter with new version in v2

ferozemohideen:
image

- renames GenericLogFilter -> GenericFilter and LogFilterName -> FilterName since metrics also uses them now
- hides jobs from metrics since we don't have data on them

### DIFF
--- a/dashboard/src/main/home/app-dashboard/validate-apply/metrics/MetricsSection.tsx
+++ b/dashboard/src/main/home/app-dashboard/validate-apply/metrics/MetricsSection.tsx
@@ -15,7 +15,7 @@ import CheckboxRow from "components/CheckboxRow";
 import { PorterApp } from "@porter-dev/api-contracts";
 import { useLocation } from "react-router";
 import Filter from "components/porter/Filter";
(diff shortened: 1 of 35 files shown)


dashboard/src/main/home/app-dashboard/validate-apply/metrics/MetricsSection.tsx
```diff
@@ -39,34 +35,50 @@ const MetricsSection: React.FunctionComponent<PropsType> = ({
...
-    return Object.keys(services).map((name) => {
-      return {
-        label: name,
-        value: name,
-      };
-    });
+  // filter out jobs until we can display metrics on them
+  const serviceOptions: GenericFilterOption[] = useMemo(() => {
+    const nonJobServiceNames = Object.keys(services).filter((name) => services[name].config.case !== "jobConfig");
+    return nonJobServiceNames.map((name) => GenericFilterOption.of(name, name));
```
ianedwards, line 44: Curious if there are some fields that exist on the `Service` type that don't exist on the client service which are of use here?

ianedwards (approved):

PR #3830, opened 2023-10-18
[POR-1973] intercom pop up on provisioning pre-flight check and v2 update errors

ferozemohideen:
image


ianedwards (approved):

ferozemohideen:
> Once auto updates to quota checks is enabled can we hide this to not make it too busy?

Yeah we can just move that function call whenever those get implemented so it's only called on the first failure

PR #3834, opened 2023-10-18
POR-1984 support definition of multiple apps from a single porter.yaml

ianedwards:
## What does this PR do?

- Adds support for a top-level apps key that allows users to specify multiple apps in a single porter.yaml file, which will be useful for monorepos running a single deploy workflow
- These are split out into multiple app protos that will be applied and revisioned separately
- This change doesn't actually handle running validate/apply for multiple apps, but will make it feasible to do so going forward


internal/porter_app/v2/yaml.go
```diff
@@ -24,43 +24,83 @@ type AppWithPreviewOverrides struct {
...
 
 	return out, nil
 }
 
+func findPreviewApp(previews []PorterApp, name string) *PorterApp {
+	var previewOverrides *PorterApp
+
+	for _, preview := range previews {
+		if preview.Name == name {
+			previewOverrides = &preview
```
ferozemohideen, line 96: might need to copy preview here, or it will be set to the last element in the array

ferozemohideen (approved):

PR #3836, opened 2023-10-19
Fix gke provisioning

sdess09:
## What does this PR do?

Allow for reprovisioning in GCP


dashboard/src/components/GCPProvisionerSettings.tsx
```diff
@@ -281,70 +281,63 @@ const GCPProvisionerSettings: React.FC<Props> = (props) => {
...
+        })
+        .catch((err) => {
+          setErrorMessage("Error fetching clusters");
+          setErrorDetails(err)
+        });
 
-      }
-    } else {
+    } catch (err) {
+      const errMessage = err.response.data.error.replace("unknown: ", "");
```
ferozemohideen, line 327: can we add `showIntercomWithMessage({ message: "I am running into an issue provisioning a cluster." });` here as well so intercom pops up when update fails

ferozemohideen (approved):

PR #3851, opened 2023-10-20
[POR-1989] Implement file picker for Dockerfile and porter.yaml in v2

ferozemohideen:
https://github.com/porter-dev/porter/assets/40551216/8aa29dbb-37e0-4f0b-a0c8-c23776983ea9

https://github.com/porter-dev/porter/assets/40551216/b887c077-c64b-4ac5-9030-233083ef7e74


dashboard/src/components/porter/CollapsibleContainer.tsx
```diff
@@ -0,0 +1,23 @@
...
+
+import { Collapse } from "react-collapse";
+import './collapsible-container.css';
+
+type Props = {
+    isOpened: boolean;
+    children: ReactNode;
+};
+
+const CollapsibleContainer: React.FC<Props> = ({
```
ferozemohideen, line 11: I added this package/component and it works decently well, except it's kinda buggy with nested collapsibles. I think we can use it in the future for simpler cases, so I left it around

ianedwards: Go with this if it works well. I've also had good experiences with [Radix UI](https://www.radix-ui.com/primitives/docs/overview/introduction), which are headless unstyled components

dashboard/src/lib/hooks/useGithubContents.ts
```diff
@@ -0,0 +1,76 @@
...
+    );
+
+    useEffect(() => {
+        if (result.isSuccess) {
+            const folders = result.data.filter((c) => c.type === "dir").sort((a, b) => a.path.localeCompare(b.path));
+            const files = result.data.filter((c) => c.type === "file").sort((a, b) => a.path.localeCompare(b.path));
+            const updatedContents = [...folders, ...files];
+    
+            if (JSON.stringify(updatedContents) !== JSON.stringify(contents)) {
+                setContents(updatedContents);
```
ianedwards, line 67: option to use lodash isEqual

dashboard/src/main/home/app-dashboard/create-app/PorterYamlModal.tsx
```diff
@@ -46,13 +62,37 @@ const PorterYamlModal: React.FC<Props> = ({ close, setPorterYamlPath, porterYaml
...
+            {showFileSelector && 
+                <div>
+                    <FileSelector 
+                        projectId={projectId} 
+                        repoId={repoId} 
+                        repoOwner={repoOwner} 
+                        repoName={repoName} 
+                        branch={branch}
+                        onFileSelect={(path: string) => setPossiblePorterYamlPath(`./${path}`)} 
+                        isFileSelectable={(path: string) => path.endsWith("porter.yaml")}
```
ianedwards, line 91: Should we validate that the file they select is called porter.yaml? I think the use case from yesterday used porter_production.yaml, as an example

ianedwards (approved):

dashboard/src/main/home/app-dashboard/create-app/PorterYamlModal.tsx
ferozemohideen: checking that it ends in .yaml instead

PR #3865, opened 2023-10-24
Env update in apply

ianedwards:
## What does this PR do?

- Followup to moving env var updates to CCP
- Removes need to call update app environment in between validate and apply, instead delegating responsibility for env updates to apply

### DIFF
--- a/api/client/porter_app.go
+++ b/api/client/porter_app.go
@@ -215,6 +215,7 @@ func (c *Client) ValidatePorterApp(
 	return resp, err
 }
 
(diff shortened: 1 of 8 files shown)


api/client/porter_app.go
```diff
@@ -215,28 +215,39 @@ func (c *Client) ValidatePorterApp(
 	return resp, err
 }
 
+type ApplyPorterAppInput struct {
```
ferozemohideen, line 218: comments pls

ferozemohideen (approved):
nice

PR #3885, opened 2023-10-26
implement log timeout

ferozemohideen:
image

image

### DIFF
--- a/dashboard/src/main/home/app-dashboard/validate-apply/logs/Logs.tsx
+++ b/dashboard/src/main/home/app-dashboard/validate-apply/logs/Logs.tsx
@@ -27,6 +27,7 @@ import { useLatestRevision } from "../../app-view/LatestRevisionContext";
 import Filter from "components/porter/Filter";
 import { useTimer } from 'react-timer-hook';
 import { useIntercom } from "lib/hooks/useIntercom";
+import axios from "axios";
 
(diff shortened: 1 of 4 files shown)


dashboard/src/main/home/app-dashboard/validate-apply/logs/Logs.tsx
```diff
@@ -199,6 +204,20 @@ const Logs: React.FC<Props> = ({
         appID: appId,
     });
 
+    const { totalSeconds, isRunning, pause: pauseLogTimeout, restart: restartLogTimeout } = useTimer({
+        expiryTimestamp: dayjs().add(DEFAULT_LOG_TIMEOUT_SECONDS, 'seconds').toDate(),
```
ianedwards, line 208: noticing that we've used dayjs in some places and date-fns in others. do you have a preference?

ianedwards (approved):

dashboard/src/main/home/app-dashboard/validate-apply/logs/Logs.tsx
ferozemohideen: I think I've gotten used to using dayjs everywhere, until we implement some kinda wrapper for dates I think we should stick to dayjs interface tbh

PR #3890, opened 2023-10-27
consolidate image endpoints into one endpoint

ferozemohideen:

### DIFF
--- a/api/server/handlers/project/images.go
+++ b/api/server/handlers/project/images.go
@@ -15,10 +15,12 @@ import (
 	"github.com/porter-dev/porter/internal/telemetry"
 )
 
+// ImagesHandler serves the /images endpoint
 type ImagesHandler struct {
 	handlers.PorterHandlerWriter
 }
 
+// NewImagesHandler returns a new ImagesHandler
 func NewImagesHandler(
 	config *config.Config,
 	writer shared.ResultWriter,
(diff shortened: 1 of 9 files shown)


api/server/handlers/project/images.go
```diff
@@ -0,0 +1,107 @@
...
+// ImageArtifact is an instance of an image in an image repository
+type ImageArtifact struct {
+	Tag       string    `json:"tag"`
+	UpdatedAt time.Time `json:"updated_at"`
+}
+
+// Image is a representation of a docker image
+// To pull one of the image instances, you must use a string of format <image.uri>:<image.artifact.tag>
+type Image struct {
+	Uri       string          `json:"uri"`
```
ianedwards, line 42: Any instance where `Uri` is not the same as saying `Repository`?

api/server/handlers/project/images.go
```diff
@@ -0,0 +1,107 @@
...
+
+	project, _ := r.Context().Value(types.ProjectScope).(*models.Project)
+
+	if project.ID == 0 {
+		err := telemetry.Error(ctx, span, nil, "project id is 0")
+		p.HandleAPIError(w, r, apierrors.NewErrPassThroughToClient(err, http.StatusBadRequest))
+		return
+	}
+
+	var resp ImagesReponse
```
ianedwards, line 64: nit: spelling

ianedwards (approved):

api/server/handlers/project/images.go
ferozemohideen: the term repository is overloaded and used differently by different cloud providers. In most cases, repositories have a name and a URI associated with them, and before, we would move around a repository object with those fields as well as others. The goal of these changes was to simplify the logic entirely and make it cloud-agnostic, since we really ever only care about a single URI/tag pairing in order to pull an image, so i decided to obfuscate the concept of a 'repository'

PR #3892, opened 2023-10-27
Efs front end

sdess09:


dashboard/src/lib/porter-apps/index.ts
```diff
@@ -73,6 +74,10 @@ export const clientAppValidator = z.object({
         message: 'Lowercase letters, numbers, and "-" only.',
       }),
   }),
+  efsStorage: z.object({
+    enabled: z.boolean(),
+    readOnly: z.boolean().optional(),
+  }),
```
ianedwards, line 80: What do you think about enabled being the field that can be read only, rather than marking the entire option as such

dashboard/src/lib/porter-apps/index.ts
```diff
@@ -477,6 +496,10 @@ export function clientAppFromProto({
       builder: "",
     },
     helmOverrides: helmOverrides,
+    efsStorage: new EFS({
+      enabled: proto.efsStorage?.enabled ?? false,
+    })
```
ianedwards, line 501: nit: this technically isn't a type error but slightly confusing to use the generated type on the client type

internal/porter_app/v2/yaml.go
```diff
@@ -88,8 +88,9 @@ type PorterApp struct {
 	Build    *Build            `yaml:"build,omitempty"`
 	Env      map[string]string `yaml:"env,omitempty"`
 
-	Predeploy *Service   `yaml:"predeploy,omitempty"`
-	EnvGroups []EnvGroup `yaml:"envGroups,omitempty"`
+	Predeploy  *Service    `yaml:"predeploy,omitempty"`
+	EnvGroups  []EnvGroup  `yaml:"envGroups,omitempty"`
+	EfsStorage *EfsStorage `yaml:"efsStorage,omitempty"`
```
ianedwards, line 93: is efs a cloud agnostic term? We're somewhat bought in to this as the syntax if someone uses it but switches cloud providers

ianedwards (approved):

internal/porter_app/v2/yaml.go
sdess09: As per discussion will be keeping EFS

dashboard/src/lib/porter-apps/index.ts
sdess09: Wont this be needed if we add more config to efs?

ianedwards: Each field could be read only in that case, if it can be specified in porter.yaml

PR #3894, opened 2023-10-27
POR-2002 updated app template form

ianedwards:


dashboard/src/main/home/app-dashboard/validate-apply/services-settings/ServiceList.tsx
```diff
@@ -56,6 +56,7 @@ type ServiceListProps = {
     namespace: string;
     appName: string;
   };
+  noAdd?: boolean;
```
ferozemohideen, line 59: nit: would rename this to allowAddServices or something and default to true

ferozemohideen (approved):

PR #3903, opened 2023-10-31
add pr action telemetry

ianedwards:


ferozemohideen (approved):
just make sure the .GetAction() returns a string otherwise you'll have to cast it to one

PR #3921, opened 2023-11-02
[POR-1616] [POR-1469] [POR-1881] [POR-2026] Backend changes to support notifications

ferozemohideen:
## What does this PR do?

These changes are introduced to reduce user confusion when it comes to APP_EVENT type events. Instead of changing the agent code to send better app events, we now convert app events to the new Notification struct on the Porter server side.

I would recommend looking at the changes in create_and_update_events.go first, which then brings the logic to the internal/notifications package. From there, you can follow the steps of HandleNotification in notification.go


internal/repository/gorm/porter_app_event.go
```diff
@@ -136,6 +169,26 @@ func (repo *PorterAppEventRepository) ReadEvent(ctx context.Context, id uuid.UUI
...
+	}
+
+	if porterAppID == 0 {
+		return notifications, errors.New("invalid porter app ID supplied")
+	}
+
+	strAppID := strconv.Itoa(int(porterAppID))
+
+	if err := repo.db.Where("porter_app_id = ? AND type = 'NOTIFICATION' AND metadata->>'app_revision_id' = ?", strAppID, appRevisionId).Find(&notifications).Error; err != nil {
+		return notifications, err
```
ianedwards, line 186: Probably not needed now but could end up being nice just to have app_revision_id as its own column on events

internal/porter_app/notifications/app_event.go
```diff
@@ -0,0 +1,296 @@
...
+// NotificationFromPorterAppEvent converts a PorterAppEvent to a Notification
+func NotificationFromPorterAppEvent(appEvent *models.PorterAppEvent) (*Notification, error) {
+	notification := &Notification{}
+	bytes, err := json.Marshal(appEvent.Metadata)
+	if err != nil {
+		return nil, err
+	}
+	err = json.Unmarshal(bytes, notification)
+	if err != nil {
+		return nil, err
```
ianedwards, line 292: `return notification, err`

internal/repository/gorm/porter_app_event.go
```diff
@@ -136,6 +169,27 @@ func (repo *PorterAppEventRepository) ReadEvent(ctx context.Context, id uuid.UUI
...
+		return notifications, errors.New("invalid app revision ID supplied")
+	}
+
+	if porterAppID == 0 {
+		return notifications, errors.New("invalid porter app ID supplied")
+	}
+
+	strAppID := strconv.Itoa(int(porterAppID))
+
+	if err := repo.db.Where("porter_app_id = ? AND type = 'NOTIFICATION' AND metadata->>'app_revision_id' = ?", strAppID, appRevisionId).Find(&notifications).Error; err != nil {
```
stefanmcshane, line 186: On that JSONB returning, we can automatically parse that to a map if you set the type to `JSONB`
https://github.com/porter-dev/porter/blob/247e71c99c9caace26357077f5a3cf74ab8d40d9/internal/models/api_contract_revision.go#L63

This would make `metadata` effectively act as `map[string]any`

internal/kubernetes/agent.go
```diff
@@ -994,6 +994,21 @@ func (a *Agent) GetDeployment(c grapher.Object) (*appsv1.Deployment, error) {
 	return res, nil
 }
 
+// GetDeploymentsBySelector returns the deployments by label selector
+func (a *Agent) GetDeploymentsBySelector(namespace string, selector string) (*appsv1.DeploymentList, error) {
+	res, err := a.Clientset.AppsV1().Deployments(namespace).List(
+		context.TODO(),
```
stefanmcshane, line 1000: ctx, never context.TODO or context.Background without good reason

internal/porter_app/notifications/porter_error/providers.go
```diff
@@ -0,0 +1,281 @@
...
+	Detail(rawAgentDetail string) string
+	// MitigationSteps returns the mitigation steps for the given error. E.g. "Please make sure that your service handles graceful shutdown when it receives a SIGTERM signal."
+	MitigationSteps(rawAgentDetail string) string
+	// Documentation returns the documentation links that would help with troubleshooting the given error.
+	Documentation(rawAgentDetail string) []string
+}
+
+// ErrorCodeToProvider maps PorterErrorCode to their respective ErrorDetailsProvider implementations.
+var ErrorCodeToProvider = map[PorterErrorCode]ErrorDetailsProvider{
+	PorterErrorCode_NonZeroExitCode:                     &NonZeroExitCodeErrorProvider{},
```
stefanmcshane, line 20: Why do these need to be pointers?

internal/porter_app/notifications/porter_error/providers.go
```diff
@@ -0,0 +1,281 @@
+package porter_error
```
stefanmcshane, line 1: None of this is a requested change, just a suggestion for when dealing with this type of code again.

This is a good implementation but relies on never changing error messages from the agent side. This is where having a distributed contract for the errors which can be returned, which can be read into the agent, or Porter is massively helpful (like an api contract, not just in the porter-dev/api-contracts proto sense)

Since a lot of this code is copy-pasta with minor changes, that usually is a good case for code generation based on a yaml, toml, etc.

Protobuf handles this with messages, grapqhl has error extensions, open api has error definitions which would allow you to consolidate this if needed.

stefanmcshane (commented):
Few comments, but overall a great step forward

internal/porter_app/notifications/porter_error/providers.go
ferozemohideen: They don't - removed

internal/repository/gorm/porter_app_event.go
ferozemohideen: resolved offline, I'm just querying for the one attribute. Going to replace this with a normal column search in https://linear.app/porter/issue/POR-2096/add-app-revision-id-column-to-porter-app-events-table

internal/porter_app/notifications/app_event.go
```diff
@@ -0,0 +1,354 @@
...
+		return telemetry.Error(ctx, span, err, "error finding matching deploy event")
+	}
+	if matchEvent.ID == uuid.Nil {
+		return telemetry.Error(ctx, span, nil, "no matching deploy event found")
+	}
+	if matchEvent.Status != string(PorterAppEventStatus_Progressing) {
+		return nil // nothing to update here
+	}
+
+	serviceStatus, ok := matchEvent.Metadata["service_deployment_metadata"]
```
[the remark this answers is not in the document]
ferozemohideen: the service deployment metadata would be handled much easier if it was a list type, but changing that would require changing the way we publish it in ccp

Filing a ticket for making that refactor, after which we can simplify this code: https://linear.app/porter/issue/POR-2101/turn-servicedeploymentmetadata-from-a-map-into-a-list-in-ccp

PR #3928, opened 2023-11-03
Use Contract for Preflight Checks

sdess09:
## What does this PR do?

Refactor Front end to use Contract


dashboard/src/components/ProvisionerSettings.tsx
```diff
@@ -126,87 +168,63 @@ const ProvisionerSettings: React.FC<Props> = (props) => {
...
+          project_id: currentProject ? currentProject.id : 0,
         }
       );
     } catch (err) {
-      // console.log(err);
     }
   };
 
-  const getStatus = () => {
+  const getStatus = (): JSX.Element | "Provisioning is still in progress..." | undefined => {
```
ianedwards, line 208: option to also just do React.ReactNode, might be cleaner.

dashboard/src/components/ProvisionerSettings.tsx
```diff
@@ -326,48 +350,52 @@ const ProvisionerSettings: React.FC<Props> = (props) => {
...
       data["cluster"]["clusterId"] = props.clusterId;
     }
 
     try {
       setIsReadOnly(true);
-      setErrorMessage(undefined);
-
       if (!props.clusterId) {
-        markStepStarted("pre-provisioning-check-started");
+        await markStepStarted("pre-provisioning-check-started");
```
ianedwards, line 390: can do `void` for these `markStepStarted` if they're just analytics updates and you don't actually want to await them

dashboard/src/components/ProvisionerSettings.tsx
```diff
@@ -388,33 +416,16 @@ const ProvisionerSettings: React.FC<Props> = (props) => {
...
         .catch((err) => {
-          console.error(err);
         });
       // }
-      {
-        props?.closeModal &&
-          props?.closeModal()
-      };
+      props?.closeModal &&
+        props?.closeModal()
```
ianedwards, line 422:
```suggestion
      if (props?.closeModal) {
        props?.closeModal()
      }
```

ianedwards (approved):

PR #3937, opened 2023-11-07
Update to use contract

sdess09:
## What does this PR do?

Make sure to use contract over preflight values if it exists

### DIFF
--- a/api/server/handlers/project_integration/preflight_check.go
+++ b/api/server/handlers/project_integration/preflight_check.go
@@ -49,11 +49,9 @@ func (p *CreatePreflightCheckHandler) ServeHTTP(w http.ResponseWriter, r *http.R
 		ProjectId:                  int64(project.ID),
 		CloudProvider:              cloudValues.CloudProvider,
 		CloudProviderCredentialsId: cloudValues.CloudProviderCredentialsId,
+		Contract:                   cloudValues.Contract,
 	}
 
-	if cloudValues.Contract != nil {
-		input.Contract = cloudValues.Contract
-	}
 	if cloudValues.PreflightValues != nil {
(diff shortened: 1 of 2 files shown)


api/server/handlers/project_integration/preflight_check.go
```diff
@@ -51,6 +51,9 @@ func (p *CreatePreflightCheckHandler) ServeHTTP(w http.ResponseWriter, r *http.R
 		CloudProviderCredentialsId: cloudValues.CloudProviderCredentialsId,
 	}
 
+	if cloudValues.Contract != nil {
+		input.Contract = cloudValues.Contract
+	}
 	if cloudValues.PreflightValues != nil {
 		if cloudValues.CloudProvider == porterv1.EnumCloudProvider_ENUM_CLOUD_PROVIDER_GCP || cloudValues.CloudProvider == porterv1.EnumCloudProvider_ENUM_CLOUD_PROVIDER_AWS {
 			input.PreflightValues = cloudValues.PreflightValues
```
ianedwards, line 59: out of curiosity - why do these have to be conditionals? any value passed to the request is going to be optional / is a pointer, so whether or not they're set will have to be handled on CCP either way

dashboard/src/components/ProvisionerSettings.tsx
```diff
@@ -536,7 +536,7 @@ const ProvisionerSettings: React.FC<Props> = (props) => {
         preflightValues: {
           case: "eksPreflightValues",
           value: new EKSPreflightValues({
-            region: awsRegion,
+            region: clusterState.awsRegion,
```
ianedwards, line 539: could be worth deprecating the awsRegion state in the future if there's supposed to be a single source of truth

ianedwards (approved):

api/server/handlers/project_integration/preflight_check.go
sdess09: you're right I think the conditionals can are extra so Ill just remove it

PR #3938, opened 2023-11-07
add is default flag to deployment target

ianedwards:
## What does this PR do?

- Adding flag indicating whether or not a deployment target is the default for a given cluster
- each cluster can have only one default deployment target
- This is useful for identifying the default deployment target for a given cluster, since names are unique across projects

### DIFF
--- a/internal/models/deployment_target.go
+++ b/internal/models/deployment_target.go
@@ -41,6 +41,9 @@ type DeploymentTarget struct {
 
 	// Metadata is a JSONB column that stores arbitrary metadata about the deployment target
 	Metadata JSONB `json:"metadata" sql:"type:jsonb" gorm:"type:jsonb;default:'{}'"`
+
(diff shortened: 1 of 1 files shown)


internal/models/deployment_target.go
```diff
@@ -41,6 +41,9 @@ type DeploymentTarget struct {
 
 	// Metadata is a JSONB column that stores arbitrary metadata about the deployment target
 	Metadata JSONB `json:"metadata" sql:"type:jsonb" gorm:"type:jsonb;default:'{}'"`
+
+	// IsDefault indicates whether this is the default deployment target for the cluster
+	IsDefault bool `gorm:"default:true" json:"is_default"`
```
d-g-town, line 46: I think we should default to `false`, make another PR that changes all the default deployment target creations to `true`, and then run a db command to set all the existing ones as true

d-g-town (approved):

PR #3946, opened 2023-11-09
add extra helper context wrapper for getting project and cluster

ianedwards:
## What does this PR do?

- Temporary solution to avoid conditionals checking if currentProject / currentCluster / user are defined or not

### DIFF
--- a/dashboard/src/main/auth/context.tsx
+++ b/dashboard/src/main/auth/context.tsx
@@ -13,7 +13,7 @@ type PorterUser = {
 type AuthContextType = {
   currentProject: ProjectType;
   currentCluster: ClusterType;
-  user?: PorterUser;
+  user: PorterUser;
 };
 
 export const AuthStateContext = createContext<AuthContextType | null>(null);
--- a/dashboard/src/main/auth/context.tsx
+++ b/dashboard/src/main/auth/context.tsx
@@ -1,0 +1,51 @@
+import React, { createContext, useContext } from "react";
+
+import Loading from "components/Loading";
+
+import { type ClusterType, type ProjectType } from "shared/types";
+
+type PorterUser = {
+  userId: number;
+  email: string;
+  isPorterUser: boolean;
+};
+
+type AuthContextType = {
+  currentProject: ProjectType;
+  currentCluster: ClusterType;
+  user?: PorterUser;
+};
+
+export const AuthStateContext = createContext<AuthContextType | null>(null);
+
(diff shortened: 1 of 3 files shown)


dashboard/src/main/auth/context.tsx
```diff
@@ -0,0 +1,51 @@
...
+  currentProject: ProjectType;
+  currentCluster: ClusterType;
+  user?: PorterUser;
+};
+
+export const AuthStateContext = createContext<AuthContextType | null>(null);
+
+export const useAuthState = (): AuthContextType => {
+  const context = useContext(AuthStateContext);
+  if (context === null) {
```
ferozemohideen, line 23:
```suggestion
  if (context == null) {
```

dashboard/src/main/auth/context.tsx
```diff
@@ -0,0 +1,51 @@
...
+  user?: PorterUser;
+};
+
+export const AuthStateProvider: React.FC<AuthStateProviderProps> = ({
+  children,
+  currentProject,
+  currentCluster,
+  user,
+}) => {
+  if (!currentProject || !currentCluster || !user) {
```
ferozemohideen, line 42: maybe add checks for > 0?

dashboard/src/main/auth/context.tsx
```diff
@@ -0,0 +1,51 @@
...
+import { type ClusterType, type ProjectType } from "shared/types";
+
+type PorterUser = {
+  userId: number;
+  email: string;
+  isPorterUser: boolean;
+};
+
+type AuthContextType = {
+  currentProject: ProjectType;
```
ferozemohideen, line 14: should we call these project and cluster to differentiate from the old context? maybe even clientProject and clientCluster? idk

ferozemohideen (approved):

dashboard/src/main/auth/context.tsx
ianedwards: yeah fair point

dashboard/src/main/auth/context.tsx
ianedwards: I'm gonna get rid of the default values so this shouldn't apply anymore

dashboard/src/main/auth/context.tsx
ianedwards: I think we actually have a lint rule that prefers the former: https://eslint.org/docs/latest/rules/eqeqeq

PR #3953, opened 2023-11-09
cut off long project name when provisioning azure cluster

d-g-town:
If a project name is longer than 18 characters, then the naming conventions for clusters in CCP lead to an Azure resource group that exceeds the max character count. Customer can always change the vanity name of their cluster later, so I'm not worried about this trimming impacting legibility of the cluster name for users.


sdess09 (approved):

dashboard/src/components/AzureProvisionerSettings.tsx
```diff
@@ -287,7 +287,7 @@ const AzureProvisionerSettings: React.FC<Props> = (props) => {
         currentCluster?.status === "UPDATING_UNAVAILABLE")
     );
     setClusterName(
-      `${currentProject.name}-cluster-${Math.random()
+        `${currentProject?.name.substring(0,16)}-cluster-${Math.random()
```
ferozemohideen, line 290: this will throw an error if currentProject?.name is undefined

sdess09:
You can also cut down on the 6 characters we add down to 3 instead

PR #3968, opened 2023-11-13
App requirements yaml

ianedwards:
## What does this PR do?

- Parse out required apps fields from porter.yaml to send to CCP on app update (example below)

```yaml
version: v2
services:
  - name: web
    port: 3000

image:
  repository: nginx
  tag: latest

previews:
  requiredApps:
    - name: my-app
```

### DIFF
--- a/internal/porter_app/v2/yaml.go
+++ b/internal/porter_app/v2/yaml.go
@@ -100,6 +100,7 @@ type PorterYAML struct {
 	Previews  *PorterApp `yaml:"previews,omitempty"`
 }
 
+// RequiredApp specifies another porter app that this app expects to be deployed alongside it
 type RequiredApp struct {
 	Name       string `yaml:"name"`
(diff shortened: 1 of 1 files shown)


internal/porter_app/v2/yaml.go
```diff
@@ -189,10 +196,6 @@ func ProtoFromApp(ctx context.Context, porterApp PorterApp) (*porterv1.PorterApp
 		}
 	}
 
-	if porterApp.Services == nil {
```
ferozemohideen, line 192: are services not required anymore?

ferozemohideen (approved):

internal/porter_app/v2/yaml.go
ianedwards: I think we are going to treat services similarly to image/build when it comes to porter.yaml. Required that at least one exists on a hydrated app, but not necessarily required to be a valid porter.yaml

PR #3978, opened 2023-11-15
POR-2115 improve app creation in new update flow

ianedwards:
## What does this PR do?

- Properly set source type when creating an app, which has previously overridden by docker repo regardless of type
- skip validate call when running update flow
- show that app is awaiting build when the hello world image is being used

### DIFF
--- a/api/server/handlers/porter_app/update_app.go
+++ b/api/server/handlers/porter_app/update_app.go
@@ -1,6 +1,7 @@
 package porter_app
 
 import (
+	"context"
 	"encoding/base64"
 	"net/http"
 
@@ -164,10 +165,15 @@ func (c *UpdateAppHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 		appProto.Name = request.Name
 	}
 
-	sourceType, image := sourceFromAppAndGitSource(appProto, request.GitSource)
+	sourceType, image, err := sourceFromAppAndGitSource(ctx, appProto, request.GitSource)
+	if err != nil {
+		err := telemetry.Error(ctx, span, err, "error getting source from app and git source")
+		c.HandleAPIError(w, r, apierrors.NewErrPassThroughToClient(err, http.StatusBadRequest))
+		return
+	}
 
 	// create porter app if it doesn't exist for the given name
-	_, err := porter_app.CreateOrGetAppRecord(ctx, porter_app.CreateOrGetAppRecordInput{
+	_, err = porter_app.CreateOrGetAppRecord(ctx, porter_app.CreateOrGetAppRecordInput{
 		ClusterID:           cluster.ID,
 		ProjectID:           project.ID,
 		Name:                appProto.Name,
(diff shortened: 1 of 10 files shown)


api/server/handlers/porter_app/update_app.go
```diff
@@ -253,16 +259,33 @@ func (c *UpdateAppHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	c.WriteResult(w, r, response)
 }
 
-func sourceFromAppAndGitSource(appProto *porterv1.PorterApp, gitSource GitSource) (porter_app.SourceType, *porter_app.Image) {
+func sourceFromAppAndGitSource(ctx context.Context, appProto *porterv1.PorterApp, gitSource GitSource) (porter_app.SourceType, *porter_app.Image, error) {
```
ferozemohideen, line 262: any reason this has to take in a proto? why not nil check in the caller function?

ianedwards: I don't have a strong preference. / reason one way or the other, but proto objects we've always kept as pointers

dashboard/src/main/home/app-dashboard/app-view/AppDataContainer.tsx
```diff
@@ -236,6 +238,15 @@ const AppDataContainer: React.FC<AppDataContainerProps> = ({ tabParam }) => {
             variables,
             secrets,
             is_env_override: true,
+            deletions: {
+              service_names: data.deletions.serviceNames.map((s) => s.name),
+              predeploy: data.deletions.predeploy.map((s) => s.name),
+              env_group_names: data.deletions.envGroupNames.map(
+                (eg) => eg.name
+              ),
+              env_variable_names: [],
```
ferozemohideen, line 247: do we still need to pass this in?

dashboard/src/main/home/app-dashboard/app-view/tabs/activity-feed/events/cards/DeployEventCard.tsx
```diff
@@ -276,14 +276,14 @@ const DeployEventCard: React.FC<Props> = ({
                 </Link>
               </ImageTagContainer>
             </>
-          ) : (
+          ) : event.metadata.image_tag ? (
             <>
               <Spacer inline x={0.5} />
```
ferozemohideen, line 281: can you add the tag icon while you're here?

dashboard/src/main/home/app-dashboard/app-view/AppDataContainer.tsx
ferozemohideen: empty env_variable_names i mean

ferozemohideen (approved):

dashboard/src/main/home/app-dashboard/app-view/AppDataContainer.tsx
ianedwards: i can go ahead and make it optional at this point

dashboard/src/main/home/app-dashboard/app-view/tabs/activity-feed/events/cards/DeployEventCard.tsx
ianedwards: sure thing

PR #3984, opened 2023-11-16
Notifications for revision update failures

ferozemohideen:


dashboard/src/main/home/app-dashboard/app-view/tabs/activity-feed/events/cards/DeployEventCard.tsx
```diff
@@ -96,6 +99,12 @@ const DeployEventCard: React.FC<Props> = ({
     revisionIdToNumber,
   ]);
 
+  const revisionNotificationsExist = useMemo(() => {
+    return latestNotifications
+      .filter(isClientRevisionNotification)
+      .some((n) => n.appRevisionId === event.metadata.app_revision_id);
+  }, [JSON.stringify(latestNotifications)]);
+
```
ianedwards, line 107: not sure if this is reusable enough to justify this - but would it make sense to return like `appNotifications`, `serviceNotifications`, `latestRevisionNotifications` from context

ianedwards (approved):

dashboard/src/main/home/app-dashboard/app-view/tabs/activity-feed/events/cards/DeployEventCard.tsx
ferozemohideen: this is a good idea. I'll see how many scopes we end up working with and might make that refactor later

PR #3992, opened 2023-11-17
Show Porter TLS enabled checkbox so they can disable it (only for ALB clusters)

ferozemohideen:
image

### DIFF
--- a/dashboard/src/lib/hooks/useClusterResourceLimits.ts
+++ b/dashboard/src/lib/hooks/useClusterResourceLimits.ts
@@ -1,5 +1,9 @@
 import { useEffect, useState } from "react";
-import { Cluster, NodeGroupType } from "@porter-dev/api-contracts";
+import {
+  Cluster,
+  LoadBalancerType,
+  NodeGroupType,
+} from "@porter-dev/api-contracts";
 import { useQuery } from "@tanstack/react-query";
 import convert from "convert";
 import { match } from "ts-pattern";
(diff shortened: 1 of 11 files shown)


dashboard/src/lib/porter-apps/services.ts
```diff
@@ -413,12 +425,22 @@ export function deserializeService({
...
+          disableTls:
+            typeof config.disableTls === "boolean" ||
+            typeof overrideWebConfig?.disableTls === "boolean"
+              ? ServiceField.boolean(
+                  config.disableTls,
+                  overrideWebConfig?.disableTls
+                )
+              : setDefaults
+              ? ServiceField.boolean(false, undefined)
+              : undefined,
```
ferozemohideen, line 437: this is copied from the private field implementation but is clunky, works for now but cut a ticket to refactor this file: https://linear.app/porter/issue/POR-2133/refactor-servicests

dashboard/src/lib/hooks/useClusterResourceLimits.ts
```diff
@@ -286,7 +289,21 @@ export const useClusterResourceLimits = ({
         })
         .otherwise(() => false);
 
+      const loadBalancerType: ClientLoadBalancerType = match(contract)
+        .with({ kindValues: { case: "eksKind" } }, (c) => {
```
d-g-town, line 293: just want to confirm that this won't crash on non-AWS clusters

d-g-town (approved):

dashboard/src/lib/hooks/useClusterResourceLimits.ts
ferozemohideen: it shouldn't

PR #4004, opened 2023-11-20
export yaml

d-g-town:
Screenshot 2023-11-20 at 12 29 42 PM
Screenshot 2023-11-20 at 12 30 13 PM


api/server/handlers/porter_app/yaml_from_revision.go
```diff
@@ -110,3 +180,168 @@ func (c *PorterYAMLFromRevisionHandler) ServeHTTP(w http.ResponseWriter, r *http
...
+	if err != nil {
+		return env, "", telemetry.Error(ctx, span, err, "error attaching env to revision")
+	}
+
+	for key, val := range revisionWithEnv.Env.Variables {
+		env[key] = val
+	}
+	for key, val := range revisionWithEnv.Env.SecretVariables {
+		env[key] = val
+	}
```
ianedwards, line 248: Do we want to set secrets on `env`? If they were to use this porter.yaml to recreate an app, they'd end up as public values in the new app

ianedwards (approved):
nice, overall just curious as to why we'll zero out so many values?

api/server/handlers/porter_app/yaml_from_revision.go
```diff
@@ -110,3 +180,168 @@ func (c *PorterYAMLFromRevisionHandler) ServeHTTP(w http.ResponseWriter, r *http
...
+			if app.Services[i].Autoscaling != nil && !app.Services[i].Autoscaling.Enabled {
+				app.Services[i].Autoscaling = nil
+			}
+			// remove port
+			app.Services[i].Port = 0
+		case v2.ServiceType_Job:
+			// remove port
+			app.Services[i].Port = 0
+			// remove instances
+			app.Services[i].Instances = nil
```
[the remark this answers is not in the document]
d-g-town: only removing them for jobs and workers! I keep them around for web services, but I'm honestly not sure why they are there for the others

api/server/handlers/porter_app/yaml_from_revision.go
d-g-town: The secrets will actually be returned as `'********'`... I didn't want to totally get rid of them in case people were confused why they were missing.

d-g-town:
> nice, overall just curious as to why we'll zero out so many values?



> nice, overall just curious as to why we'll zero out so many values?

I tried to only zero out the ones that I thought people would be confused to see.  E.g. if private is false, then chances are they don't want to explicitly declare it anyway. I was trying to return a yaml that was as succinct as possible to get them the app they had configured.

PR #4007, opened 2023-11-21
Notification logs

ferozemohideen:
These changes:

1. use the new should_view_logs variable to determine whether logs should be rendered for a notification or not
2. logs are localized to the time period around when the incident occurred if the notification is not deploy related. otherwise live logs are shown
3. logs are collapsible
4. misc refactor which allows for UNKNOWN app revision status so that the frontend isnt borked if we forget to support a new status

image


internal/porter_app/notifications/notification.go
```diff
@@ -60,6 +60,8 @@ type PorterError struct {
 	MitigationSteps string `json:"mitigation_steps"`
 	// Documentation is a list of links to documentation that can be used to resolve the error
 	Documentation []string `json:"documentation"`
+	// ShouldViewLogs is a boolean indicating whether the user should look at container logs for more information
```
ianedwards, line 63: is this set on CCP? not seeing where in this PR

internal/porter_app/revisions.go
```diff
@@ -110,8 +110,10 @@ func EncodedRevisionFromProto(ctx context.Context, appRevision *porterv1.AppRevi
 	b64 := base64.StdEncoding.EncodeToString(encoded)
 
 	status, err := appRevisionStatusFromProto(appRevision.Status)
+	telemetry.WithAttributes(span, telemetry.AttributeKV{Key: "status", Value: string(status)})
 	if err != nil {
-		return revision, telemetry.Error(ctx, span, err, "error getting app revision status from proto")
+		// ignore error so that we can still return the revision
+		telemetry.WithAttributes(span, telemetry.AttributeKV{Key: "get-status-error", Value: err.Error()})
 	}
```
d-g-town, line 117: I would not swallow this error, and instead either do 
```
telemetry.WithAttributes(span, telemetry.AttributeKV{Key: "re-visionstatus", Value: string(appRevision.Status)})
_ = telemetry.Error(ctx, span, nil, "unknown revision type") // flagged as an error for visibilityu
```
or just don't flag it as an error and have a CCP notification on it

d-g-town: just don't want to normalize using values (like `status` here) if the error is not nil

internal/porter_app/revisions.go
```diff
@@ -202,7 +204,7 @@ func AttachEnvToRevision(ctx context.Context, inp AttachEnvToRevisionInput) (Rev
 }
 
 func appRevisionStatusFromProto(status string) (models.AppRevisionStatus, error) {
-	var appRevisionStatus models.AppRevisionStatus
+	appRevisionStatus := models.AppRevisionStatus_Unknown
```
ianedwards, line 207: shouldn't block this but might be worth it to have the porter_app type use a different enum other than the model revision status options, if it's going to deviate from what's stored in the DB. 

Couldn't find where the status is actually used after running `EncodedRevisionFromProto` so it's probably fine as is for now

ianedwards (approved):

d-g-town (approved):

internal/porter_app/notifications/notification.go
ferozemohideen: Yep it's set in CCP, this just ensures we parse it correctly on Porter

PR #4033, opened 2023-11-29
simplify create and update events

ferozemohideen:
removing logic that is now handled by CCP and hopefully making it easier to read

### DIFF
--- a/api/server/handlers/porter_app/create_and_update_events.go
+++ b/api/server/handlers/porter_app/create_and_update_events.go
@@ -39,10 +39,6 @@ func NewCreateUpdatePorterAppEventHandler(
 	}
 }
 
-const (
-	crashLoopBackoffSubstring string = "stuck in a restart loop"
-)
-
 func (p *CreateUpdatePorterAppEventHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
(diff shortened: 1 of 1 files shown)


d-g-town (approved):

api/server/handlers/porter_app/create_and_update_events.go
```diff
@@ -78,33 +74,32 @@ func (p *CreateUpdatePorterAppEventHandler) ServeHTTP(w http.ResponseWriter, r *
...
+				return
+			}
 		}
-		return
-	}
-
-	if request.ID == "" {
-		event, err := p.createNewAppEvent(ctx, *cluster, appName, request.DeploymentTargetID, request.Status, string(request.Type), request.TypeExternalSource, request.Metadata)
+	} else { // event id provided, so update an existing event matching that event
+		event, err = p.updateExistingAppEvent(ctx, *cluster, appName, *request)
```
ianedwards, line 97: suggestion but not necessary- 
1. check `if request.ID != ""` first, run update
2. check `if event.ID != "" && request.DeploymentTargetID == ""`, create new app event
3. handle notification

might get rid of some elses but I might be getting the logic wroing

ianedwards (approved):

PR #4036, opened 2023-11-30
manage deployment targets

d-g-town:


dashboard/src/lib/hooks/useDeploymentTarget.ts
```diff
@@ -54,3 +66,50 @@ export function useDefaultDeploymentTarget() {
 
   return deploymentTarget;
 }
+
+export function useListDeploymentTargets(preview: boolean): DeploymentTarget[] | undefined {
```
ferozemohideen, line 70: nit: would rename to useDeploymentTargetList

dashboard/src/lib/hooks/useDeploymentTarget.ts
```diff
@@ -54,3 +66,50 @@ export function useDefaultDeploymentTarget() {
...
+    const { currentProject, currentCluster } = useContext(Context);
+    const [
+        deploymentTargets,
+        setDeploymentTargets,
+    ] = useState<DeploymentTarget[] | undefined>(undefined);
```
ferozemohideen, line 75: why not just default to empty array? that way you don't have to return undefined

dashboard/src/main/home/app-dashboard/apps/Apps.tsx
```diff
@@ -33,9 +31,9 @@ import DeleteEnvModal from "main/home/cluster-dashboard/preview-environments/v2/
 import { useHistory } from "react-router";
 import DashboardPlaceholder from "components/porter/DashboardPlaceholder";
 
-type Props = {};
+type Props = unknown;
```
ferozemohideen, line 34: if there's no props just remove this here and just change it to `const Apps: React.FC = () => {`

dashboard/src/lib/hooks/useDeploymentTarget.ts
ferozemohideen, line 70: nit: make this take in an object `{preview: boolean}` instead of a value

dashboard/src/main/home/app-dashboard/create-app/CreateApp.tsx
```diff
@@ -197,14 +198,25 @@ const CreateApp: React.FC<CreateAppProps> = ({ history }) => {
     source: source?.type === "github" ? source : null,
     appName: "", // only want to know if porter.yaml has name set, otherwise use name from input
   });
-  const deploymentTarget = useDefaultDeploymentTarget();
+  const defaultDeploymentTarget = useDefaultDeploymentTarget();
+  const deploymentTargets = useListDeploymentTargets(false)
+  const [ deploymentTargetID, setDeploymentTargetID ] = React.useState("");
```
ferozemohideen, line 203: maybe this should be part of the form object? not sure though @ianedwards what do you think

dashboard/src/main/home/app-dashboard/create-app/CreateApp.tsx
```diff
@@ -197,14 +198,25 @@ const CreateApp: React.FC<CreateAppProps> = ({ history }) => {
...
   });
   const { currentClusterResources } = useClusterResources();
 
+  useEffect(() => {
+    if (deploymentTargetID === "") {
+      setDeploymentTargetID(defaultDeploymentTarget?.id ?? "")
+    }
+
+  }, [
+    defaultDeploymentTarget
```
ferozemohideen, line 217: this depency should be deploymentTargetID

dashboard/src/main/home/app-dashboard/create-app/CreateApp.tsx
```diff
@@ -595,6 +607,32 @@ const CreateApp: React.FC<CreateAppProps> = ({ history }) => {
...
                     />
+                    {currentProject?.managed_deployment_targets_enabled && (
+                        <>
+                          <Spacer y={1} />
+                          <Select
+                              value={deploymentTargetID}
+                              width="300px"
+                              options={deploymentTargets ? deploymentTargets.filter(
+                                  (target: DeploymentTarget) => {
+                                    return !target.is_preview
```
ferozemohideen, line 618: filter not necessary

dashboard/src/main/home/app-dashboard/create-app/CreateApp.tsx
```diff
@@ -595,6 +607,32 @@ const CreateApp: React.FC<CreateAppProps> = ({ history }) => {
...
+                                  }
+                              ).map((target: DeploymentTarget) => {
+                                return {
+                                  value: target.id,
+                                  label: target.name,
+                                };
+                              }) : []}
+                              setValue={(value) => {
```
ferozemohideen, line 626: gonna replace the select with the select+create component right?

dashboard/src/main/home/app-dashboard/create-app/CreateApp.tsx
```diff
@@ -595,6 +607,32 @@ const CreateApp: React.FC<CreateAppProps> = ({ history }) => {
...
+                                if (value !== deploymentTargetID) {
```
ferozemohideen, line 627: this check not necessary

dashboard/src/main/home/cluster-dashboard/preview-environments/v2/PreviewEnvs.tsx
```diff
@@ -1,80 +1,32 @@
...
-  id: z.string(),
-  project_id: z.number(),
-  cluster_id: z.number(),
-  selector: z.string(),
-  selector_type: z.string(),
-  created_at: z.string(),
-  updated_at: z.string(),
-});
-export type RawDeploymentTarget = z.infer<typeof rawDeploymentTargetValidator>;
+import {useListDeploymentTargets} from "../../../../../lib/hooks/useDeploymentTarget";
```
ferozemohideen, line 19: fix

dashboard/src/shared/DeploymentTargetContext.tsx
```diff
@@ -22,34 +21,68 @@ export const useDeploymentTarget = () => {
   return context;
 };
 
-const DeploymentTargetProvider = ({ children }: { children: JSX.Element }) => {
+const DeploymentTargetProvider = ({ children }: { children: JSX.Element }):  JSX.Element => {
   const { search } = useLocation();
+  const { currentCluster, currentProject } = useContext(Context)
   const queryParams = new URLSearchParams(search);
 
   const idParam = queryParams.get("target");
```
ferozemohideen, line 29: nit: ambiguous variable name

dashboard/src/lib/hooks/useDeploymentTarget.ts
ferozemohideen: if you're worried about differentiating between no deployment targets exist vs we are still fetching them, just export the isLoading from the useQuery

dashboard/src/shared/DeploymentTargetContext.tsx
```diff
@@ -22,34 +21,68 @@ export const useDeploymentTarget = () => {
...
+              deployment_target_id: idParam,
+            }
+        );
+
+        const deploymentTarget  = await z.object({deployment_target: deploymentTargetValidator}).parseAsync(res.data);
+
+        return deploymentTarget.deployment_target;
+      },
+      {
+        enabled:
```
ferozemohideen, line 60: instead of conditioning on currentProject?.multiple_deployment_targets flag in various places on the frontend, you can just add that flag in the enabled here. That way, we will always send users to URLs that end in `?target=...` but unless the flag is true, we don't do anything with it

ferozemohideen (changes requested):

api/server/handlers/porter_app/latest_app_revisions.go
```diff
@@ -121,6 +115,28 @@ func (c *LatestAppRevisionsHandler) ServeHTTP(w http.ResponseWriter, r *http.Req
...
 
+		details, err := c.Config().ClusterControlPlaneClient.DeploymentTargetDetails(ctx, connect.NewRequest(&porterv1.DeploymentTargetDetailsRequest{
+			ProjectId:          int64(project.ID),
+			DeploymentTargetId: encodedRevision.DeploymentTarget.ID,
+		}))
+		if err != nil {
+			err := telemetry.Error(ctx, span, err, "error getting deployment target details")
+			c.HandleAPIError(w, r, apierrors.NewErrPassThroughToClient(err, http.StatusInternalServerError))
+			return
+		}
```
d-g-town, line 126: I will clean this up to make it slightly more efficient

d-g-town: with a set

dashboard/src/main/home/app-dashboard/create-app/CreateApp.tsx
ianedwards: I guess it depends if we ever plan to update deploymentTargetID for an existing app. Don't want to create additional overhead to set it on the client app if it's not going to be changed

dashboard/src/lib/hooks/useDeploymentTarget.ts
```diff
@@ -4,12 +4,20 @@ import { Context } from "shared/Context";
...
+export const deploymentTargetValidator = z.object({
+    id: z.string(),
+    project_id: z.number(),
+    cluster_id: z.number(),
+    namespace: z.string(),
+    name: z.string(),
+    is_preview: z.boolean(),
+    is_default: z.boolean(),
+    created_at: z.string(),
+    updated_at: z.string(),
```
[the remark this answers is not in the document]
d-g-town: We use it for sorting preview environments.

api/server/handlers/deployment_target/get.go
```diff
@@ -77,8 +80,31 @@ func (c *GetDeploymentTargetHandler) ServeHTTP(w http.ResponseWriter, r *http.Re
...
-		DeploymentTarget: deploymentTarget,
+		DeploymentTarget: types.DeploymentTarget{
+			ID:        id,
+			ProjectID: project.ID,
+			ClusterID: cluster.ID,
+			Name:      deploymentTarget.Name,
+			Namespace: deploymentTarget.Namespace,
+			IsPreview: deploymentTarget.IsPreview,
+			IsDefault: deploymentTarget.IsDefault,
+			CreatedAt: time.Time{},
```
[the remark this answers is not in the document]
d-g-town: I think I'm leaning towards always using full structs (I'll add a comment about why we don't have the values to fill in here)

dashboard/src/main/home/app-dashboard/create-app/CreateApp.tsx
d-g-town: I'm not going to do this right now

dashboard/src/main/home/app-dashboard/create-app/CreateApp.tsx
d-g-town: I want this to be defaultDeploymentTarget, because that's the only time I would call setDeploymentTargetID outside the selector

dashboard/src/main/home/app-dashboard/create-app/CreateApp.tsx
d-g-town: yes, but in a following PR

dashboard/src/shared/DeploymentTargetContext.tsx
d-g-town: The thing is we also need this flag for preview environments, so I can't use it as an indicator that `multiple_deployment_targets` is true

dashboard/src/main/home/app-dashboard/create-app/CreateApp.tsx
d-g-town: removing

api/server/handlers/porter_app/default_deployment_target.go
```diff
@@ -37,7 +40,9 @@ type DefaultDeploymentTargetRequest struct{}
 
 // DefaultDeploymentTargetResponse is the response object for the /default-deployment-target endpoint
 type DefaultDeploymentTargetResponse struct {
-	DeploymentTargetID string `json:"deployment_target_id"`
+	// Deprecated: use inline types.DeploymentTarget fields instead
+	DeploymentTargetID     string `json:"deployment_target_id"`
+	types.DeploymentTarget `json:"deployment_target"`
```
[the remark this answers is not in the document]
d-g-town: I like the fact that it is not inline because then I can ignore the deprecated field when I am zod parsing

dashboard/src/main/home/app-dashboard/create-app/CreateApp.tsx
```diff
@@ -362,7 +376,11 @@ const CreateApp: React.FC<CreateAppProps> = ({ history }) => {
         });
 
         if (source.type === "docker-registry") {
-          history.push(`/apps/${app.name}`);
+          let targetSuffix = "";
+          if (currentProject?.managed_deployment_targets_enabled) {
+            targetSuffix = `?target=${deploymentTargetID}`;
+          }
+          history.push(`/apps/${app.name}${targetSuffix}`);
```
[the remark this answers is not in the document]
d-g-town: yes we do, great catch

dashboard/src/main/home/app-dashboard/apps/Apps.tsx
```diff
@@ -67,10 +64,16 @@ const Apps: React.FC<Props> = ({ }) => {
...
+      let deploymentTargetID: string | undefined = currentDeploymentTarget.id;
+      if (currentProject.managed_deployment_targets_enabled && !currentDeploymentTarget.is_preview) {
+        deploymentTargetID = undefined;
+      }
+
       const res = await api.getLatestAppRevisions(
         "<token>",
         {
-          deployment_target_id: currentDeploymentTarget?.id,
+          deployment_target_id: deploymentTargetID,
```
ferozemohideen, line 75: nit: can inline this with a ternary operator

ferozemohideen (approved):
lgtm but imo in general it's best to use `consts` created through ternaries wherever possible in typescript, rather than `lets`, which translates more from Go logic

PR #4048, opened 2023-12-05
filter porter domains on yaml export

ianedwards:

### DIFF
--- a/api/server/handlers/porter_app/yaml_from_revision.go
+++ b/api/server/handlers/porter_app/yaml_from_revision.go
@@ -161,7 +161,7 @@ func (c *PorterYAMLFromRevisionHandler) ServeHTTP(w http.ResponseWriter, r *http
 	app.Env = env
 	app.EnvGroups = envGroups
 
-	app = zeroOutValues(app)
+	app = zeroOutValues(app, c.Config().ServerConf.AppRootDomain)
 
 	if request.ShouldFormatForExport {
 		app = formatForExport(app)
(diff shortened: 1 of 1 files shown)


api/server/handlers/porter_app/yaml_from_revision.go
```diff
@@ -344,6 +344,14 @@ func zeroOutValues(app v2.PorterApp) v2.PorterApp {
 			if app.Services[i].Private != nil && !*app.Services[i].Private {
 				app.Services[i].Private = nil
 			}
+			// remove porter domains
+			var filteredDomains []v2.Domains
+			for j := range app.Services[i].Domains {
```
d-g-town, line 349: nit: more readable to use `for _, domain := range app.Services[i].Domains` because you aren't editing the domains in places

api/server/handlers/porter_app/yaml_from_revision.go
```diff
@@ -161,7 +161,7 @@ func (c *PorterYAMLFromRevisionHandler) ServeHTTP(w http.ResponseWriter, r *http
 	app.Env = env
 	app.EnvGroups = envGroups
 
-	app = zeroOutValues(app)
+	app = zeroOutValues(app, c.Config().ServerConf.AppRootDomain)
```
d-g-town, line 164: do we want to remove porter custom domains from the diff changes too? If so, this is great. If not, this logic should be moved to the `formatForExport` method

d-g-town (approved):

api/server/handlers/porter_app/yaml_from_revision.go
ianedwards: I see. Missed that this endpoint is used in both cases, I'll move this to `formatForExport`

PR #4061, opened 2023-12-07
support doppler

d-g-town:
- basic doppler env group minus app attach
- basic doppler attach flow
- basic doppler attach flow
- support doppler environment groups


api/server/handlers/environment_groups/are_external_providers_enabled.go
```diff
@@ -0,0 +1,62 @@
+package environment_groups
+
```
ferozemohideen, line 2: comments on this file

api/server/handlers/environment_groups/are_external_providers_enabled.go
```diff
@@ -0,0 +1,62 @@
...
+func (c *AreExternalProvidersEnabledHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
+	ctx, span := telemetry.NewSpan(r.Context(), "serve-are-external-providers-enabled")
+	defer span.End()
+
+	cluster, _ := ctx.Value(types.ClusterScope).(*models.Cluster)
+
+	resp, err := c.Config().ClusterControlPlaneClient.AreExternalEnvGroupProvidersEnabled(ctx, connect.NewRequest(&porterv1.AreExternalEnvGroupProvidersEnabledRequest{
+		ProjectId: int64(cluster.ProjectID),
+		ClusterId: int64(cluster.ID),
```
ianedwards, line 49: not sure what the authz implications would be but does it make more sense to use the project id from the project scope just to make sure that there's not a mismatch

dashboard/src/components/porter/Banner.tsx
```diff
@@ -20,6 +20,10 @@ const Banner: React.FC<Props> = ({
   suffix,
 }) => {
   const renderIcon = () => {
+    if (icon === "none") {
+      return null;
+    }
```
ianedwards, line 25: if icon is allowed to be undefined is passing in "none" necessary? Could change the condition to just check for whether or not it's passed in

```
if (!icon) {
  return null
}
```

dashboard/src/main/home/integrations/DopplerIntegrationList.tsx
```diff
@@ -0,0 +1,252 @@
...
+                onToggle={installDoppler}
+                disabled={dopplerToggled}
+            >
+                {dopplerToggled ? dopplerEnabled ? "Doppler integration enabled": "Enabling Doppler integration . . ." : "Enable Doppler integration"}
+            </ToggleRow>
+        </Banner>
+        <Spacer y={1} />
+        <Button onClick={() => { setShowServiceTokenModal(true) }}>
+        + Add Doppler env group
+      </Button>
```
ianedwards, line 168: looks like your formatter isn't working as a heads up

dashboard/src/main/home/integrations/DopplerIntegrationList.tsx
```diff
@@ -0,0 +1,252 @@
...
+            (<Placeholder>
+                To enable integration with Doppler, <Spacer inline x={.5}/><Link to={`/cluster-dashboard`} hasunderline>re-provision your cluster</Link>.
+            </Placeholder>)
+            :
+            (
+            <>
+                <Banner icon="none">
+                    <ToggleRow
+                        isToggled={dopplerToggled}
+                        onToggle={installDoppler}
```
ferozemohideen, line 137: what happens if you toggle it on and off and on and off

ferozemohideen (approved):

api/server/handlers/environment_groups/list.go
```diff
@@ -66,6 +78,16 @@ func (c *ListEnvironmentGroupsHandler) ServeHTTP(w http.ResponseWriter, r *http.
 		return
 	}
 
+	if request.Type != "" {
+		var filteredEnvGroupVersions []environmentgroups.EnvironmentGroup
```
ianedwards, line 82: potentially validate the type here?

api/server/router/cluster.go
```diff
@@ -1751,5 +1751,63 @@ func getClusterRoutes(
...
 	})
 
+	// POST /api/projects/{project_id}/clusters/{cluster_id}/environment-groups/enable-external-providers
+	enableExternalProvidersEndpoint := factory.NewAPIEndpoint(
+		&types.APIRequestMetadata{
+			Verb:   types.APIVerbUpdate,
+			Method: types.HTTPVerbPost,
+			Path: &types.Path{
+				Parent:       basePath,
+				RelativePath: relPath + "/environment-groups/enable-external-providers",
```
ianedwards, line 1761: sprintf

api/server/router/cluster.go
```diff
@@ -1751,5 +1751,63 @@ func getClusterRoutes(
...
+	})
+
+	// GET /api/projects/{project_id}/clusters/{cluster_id}/environment-groups/are-external-providers-enabled
+	areExternalProvidersEnabledEndpoint := factory.NewAPIEndpoint(
+		&types.APIRequestMetadata{
+			Verb:   types.APIVerbGet,
+			Method: types.HTTPVerbGet,
+			Path: &types.Path{
+				Parent:       basePath,
+				RelativePath: relPath + "/environment-groups/are-external-providers-enabled",
```
ianedwards, line 1790: sprintf

dashboard/src/main/home/integrations/DopplerIntegrationList.tsx
```diff
@@ -0,0 +1,252 @@
...
+            },
+            {
+              id: currentProject.id,
+              cluster_id: currentCluster.id,
+            }
+        ).then( () => {
+            setShowServiceTokenModal(false);
+            history.push("/env-groups")
+        }).catch(() => {});
+  };
```
ianedwards, line 114: could make sense to wrap this in some loading state and prevent the button from being clicked twice, in case it's potentially error prone for this to run back to back

ianedwards (approved):

dashboard/src/main/home/integrations/DopplerIntegrationList.tsx
d-g-town: toggling is disabled when dopplerToggled is true

api/server/handlers/environment_groups/list.go
d-g-town: Not going to do this now because I don't want to copy constants to porter, I think we should move this call to CCP and do type checking there

d-g-town: let me add it to telemetry though

dashboard/src/components/porter/Banner.tsx
d-g-town: I think null icon is currently defaulting to an `info` icon

PR #4072, opened 2023-12-11
create deployment target in app creation

d-g-town:


dashboard/src/components/porter/Selector.tsx
```diff
@@ -0,0 +1,354 @@
...
+
+  const handleOptionClick = (option: { value: T; label: string }): void => {
+    setActiveValue(option.value);
+    setExpanded(false);
+  };
+
+  const renderDropdownLabel = (): JSX.Element => {
+    return (
+      <>{dropdownLabel && <DropdownLabel>{dropdownLabel}</DropdownLabel>}</>
+    );
```
ianedwards, line 85: maybe just use an if to check if `dropdownLabel` is undefined and return null if so

dashboard/src/components/porter/Selector.tsx
```diff
@@ -0,0 +1,354 @@
...
+
+  const renderIcon = (): JSX.Element => {
+    let icon;
+    options.forEach(
+      (option: { value: T; label: string; icon?: JSX.Element }) => {
+        if (option.icon && option.value === activeValue) {
+          icon = option.icon;
+        }
+      }
+    );
```
ianedwards, line 141: `const icon = options.find((opt) => opt.icon && option.value === activeValue)`

dashboard/src/components/porter/Selector.tsx
```diff
@@ -0,0 +1,354 @@
...
+    return (
+      <>
+        {icon && (
+          <Icon>
+            <img src={icon} />
+          </Icon>
+        )}
```
ianedwards, line 148: same thing here as `renderDropdownLabel`

dashboard/src/lib/hooks/UseOutsideClick.ts
```diff
@@ -0,0 +1,24 @@
+import type React from "react";
+import { useEffect, useRef } from "react";
+
+export const useOutsideClick = (
```
ianedwards, line 4: nice

dashboard/src/components/CreateDeploymentTargetModal.tsx
```diff
@@ -0,0 +1,207 @@
...
+  const [deploymentTargetName, setDeploymentTargetName] = useState("");
+  const { deploymentTargetList, isDeploymentTargetListLoading } =
+    useDeploymentTargetList({ preview: false });
+  const { currentProject, currentCluster } = useContext(Context);
+
+  const isRestrictedName = (name: string): boolean =>
+    RestrictedNamespaces.includes(name);
+
+  const hasInvalidCharacters = (name: string): boolean =>
+    name !== "" && !/^([a-z0-9]|-)+$/.test(name);
```
ianedwards, line 39: for the future - this component could have been well suited to use react-hook-form, and it would have been essentially the same implementation as the modal for adding services

ianedwards: also I'd check that this length is < 63

dashboard/src/components/porter/Selector.tsx
```diff
@@ -0,0 +1,354 @@
...
+    return (
+      <>
+        {options.map(
+          (
+            option: { value: T; label: string; icon?: JSX.Element },
+            i: number
+          ) => {
+            return (
+              <Option
+                key={i}
```
ianedwards, line 59: watch out for indices as keys when possible. If the items ever changed order for some reason then React wouldn't know

ianedwards (approved):

PR #4075, opened 2023-12-11
Add support for GCP Instance Types for slider

sdess09:
## What does this PR do?

Add support for GCP nodes for sliders

### DIFF
--- a/dashboard/src/lib/hooks/useClusterResourceLimits.ts
+++ b/dashboard/src/lib/hooks/useClusterResourceLimits.ts
@@ -106,13 +106,26 @@ const clusterNodesValidator = z
       return defaultResources;
     }
     const instanceType = data.labels["beta.kubernetes.io/instance-type"];
-    const res = z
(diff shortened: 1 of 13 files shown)


dashboard/src/lib/hooks/useClusterResourceLimits.ts
```diff
@@ -106,13 +106,26 @@ const clusterNodesValidator = z
...
-    if (!res.success) {
+
+    // Updated parsing logic
+    let parsedType;
+    if (instanceType && instanceType.includes(".")) {
+      parsedType = z
+        .tuple([z.string(), z.string()])
+        .safeParse(instanceType.split("."));
+    } else if (instanceType && instanceType.includes("-")) {
+      const parts = instanceType.split("-");
```
ianedwards, line 117: Could you set `parsedType` to the validated result here, instead of the validation type, and just return the defaults of the validation fails?

ianedwards (approved):

PR #4082, opened 2023-12-12
updated entry point for configuring preview envs

ianedwards:
## What does this PR do?

- Moving form for setting up preview envs to the Preview Env tab in the dashboard
- This is the first step towards providing a full form for getting to parity with porter.yaml in the UI for previews

https://github.com/porter-dev/porter/assets/104453631/6da127a8-8525-42dd-9960-cf1e8b2b6bcd


dashboard/src/main/home/cluster-dashboard/preview-environments/v2/PreviewEnvGrid.tsx
```diff
@@ -45,10 +47,34 @@ const PreviewEnvGrid: React.FC<PreviewEnvGridProps> = ({
...
+          <Spacer y={1} />
+
+          <Text color={"helper"}>
+            Get started by enabling preview envs for your apps.
+          </Text>
+          <Spacer y={1} />
+          <Button
+            onClick={() => {
+              // setTab("config");
+            }}
```
ferozemohideen, line 64: remove?

ferozemohideen (approved):

ferozemohideen:
Code changes lgtm, just some minor design feedback from me:
- for the environments tab, are those environments 'living' aka PR hasn't been merged yet? could be cool to have an X icon on the far right that deletes the environment if so
- I'm not a huge fan of 'Configure Apps' - maybe just Configuration or Settings?
- what does clicking into each tile on the next-test get you to? does it take you to the app?
- The clickable of 'Update Previews' is a bit weird inside the parent clickable tile, but not sure a better way around that
- In the setup previews modal, are the environment variables overrides or pulled directly from the parent app? If they are pulled directly, the modal can possibly be hard to scroll through because some apps have lots of env vars. Maybe consider making the whole form a collapsing list type component like the cloudformation screen or into its own dedicated page
- Not sure what the Update Settings under the preview environments enabled does (I guess it takes you straight to the setup previews modal)? What do you think about making that a toggle with a gear icon next to it? Or actually, do we even need to take them back to the app after they configure preview environments from the preview env page? could we just get rid of that part in the app's settings tab altogether?

dashboard/src/main/home/cluster-dashboard/preview-environments/v2/PreviewEnvGrid.tsx
ianedwards: meant to pass in a prop and uncomment, will fix

ianedwards:
> Code changes lgtm, just some minor design feedback from me:
> 
> * for the environments tab, are those environments 'living' aka PR hasn't been merged yet? could be cool to have an X icon on the far right that deletes the environment if so
> * I'm not a huge fan of 'Configure Apps' - maybe just Configuration or Settings?
> * what does clicking into each tile on the next-test get you to? does it take you to the app?
> * The clickable of 'Update Previews' is a bit weird inside the parent clickable tile, but not sure a better way around that
> * In the setup previews modal, are the environment variables overrides or pulled directly from the parent app? If they are pulled directly, the modal can possibly be hard to scroll through because some apps have lots of env vars. Maybe consider making the whole form a collapsing list type component like the cloudformation screen or into its own dedicated page
> * Not sure what the Update Settings under the preview environments enabled does (I guess it takes you straight to the setup previews modal)? What do you think about making that a toggle with a gear icon next to it? Or actually, do we even need to take them back to the app after they configure preview environments from the preview env page? could we just get rid of that part in the app's settings tab altogether?

- Yes they're living. Agree with having some more settings and info for the env but planning to do that in a follow up once I can attach metadata on the deployment target
- Updated the tab text
- Used foreground color instead of clickable, so it's clearer the button is how to update things
- Agree that the modal didn't make much sense, reverting to the old screen which should give us more space to work with especially when including addons
- I removed the settings entry point from the app so only the preview envs screen can be used

https://github.com/porter-dev/porter/assets/104453631/2b2dce55-2158-44a3-841f-133e0a28b8e2

PR #4089, opened 2023-12-13
fix routing for multiple deployment targets

ferozemohideen:
tested with multiple DT flag turned on and off; seems to work fine


ianedwards (approved):
nice. lgtm, only potential thought might be to use a more descriptive name than `internalLinkBuilder` so that it's clear what info is used

PR #4090, opened 2023-12-13
reorganize preview form under tabs

ianedwards:
## What does this PR do?

- Since required apps and addons introduce quite a few more fields, the existing app template form design will become a bit cluttered over time
- Introducing tabs in this PR for Services and Environment Variables, with Required Apps and Addons being added after this is merged
- Also routing app template updates to CCP so that addons can be properly accounted for

https://github.com/porter-dev/porter/assets/104453631/af20c469-2039-49bb-9663-9e33ff08b47d


ferozemohideen (approved):
sick

PR #4097, opened 2023-12-14
support selecting deployment target for app commands

d-g-town:


cli/cmd/commands/app.go
```diff
@@ -58,6 +59,14 @@ func registerCommand_App(cliConf config.CLIConfig) *cobra.Command {
 		Short: "Runs a command for your application.",
 	}
 
+	appCmd.PersistentFlags().StringVarP(
+		&deploymentTarget,
+		"target",
+		"x",
+		"default",
+		"the deployment target for the app, default is \"default\"",
```
ianedwards, line 67: not sure this matters at the moment, but aren't names scoped to project so this default name wouldn't necessarily be true for multi cluster?

cli/cmd/v2/update_image.go
```diff
@@ -9,21 +9,16 @@ import (
...
-	targetResp, err := client.DefaultDeploymentTarget(ctx, projectId, clusterId)
-	if err != nil {
-		return "", fmt.Errorf("error calling default deployment target endpoint: %w", err)
-	}
-
-	if targetResp.DeploymentTargetID == "" {
-		return "", errors.New("deployment target id is empty")
+func UpdateImage(ctx context.Context, tag string, client api.Client, projectId, clusterId uint, appName string, deploymentTargetName string) (string, error) {
+	if deploymentTargetName == "" {
+		return "", errors.New("please provide a deployment target")
```
ianedwards, line 14: is the default passed in by the caller?

ianedwards (approved):
any reason we can't do this for apply?

cli/cmd/v2/update_image.go
d-g-town: yes, it is

cli/cmd/commands/app.go
d-g-town: that's a great point, I'll have to think about what the most intuitive way to do this for multicluster is. I'll make adjustments in another PR.

d-g-town:
No reason we can't do it for apply, but I think that might be slightly more involved, so I'll tackle it in another PR

PR #4102, opened 2023-12-15
Preliminary work for status tab

ferozemohideen:


api/server/router/porter_app.go
```diff
@@ -1183,14 +1183,14 @@ func getPorterAppRoutes(
 		Router:   r,
 	})
 
-	// GET /api/projects/{project_id}/clusters/{cluster_id}/apps/{porter_app_name}/pods -> cluster.NewPodStatusHandler
-	appPodStatusEndpoint := factory.NewAPIEndpoint(
```
d-g-town, line 1187: this will break the `app run` CLI before we upgrade, right?

ferozemohideen: ah you're right, i forgot the CLI was using this. going to re-add the file so the cli continues using it

api/server/handlers/porter_app/service_status.go
```diff
@@ -0,0 +1,175 @@
...
+	if listAppRevisionsResp == nil || listAppRevisionsResp.Msg == nil {
+		err = telemetry.Error(ctx, span, nil, "list app revisions response is nil")
+		c.HandleAPIError(w, r, apierrors.NewErrPassThroughToClient(err, http.StatusInternalServerError))
+		return
+	}
+
+	appRevisions := listAppRevisionsResp.Msg.AppRevisions
+	if appRevisions == nil {
+		appRevisions = []*porterv1.AppRevision{}
+	}
```
ianedwards, line 143: don't think this check is necessary. `range` on a nil slice should be valid

dashboard/src/lib/hooks/useAppStatus.ts
```diff
@@ -1,215 +1,201 @@
...
+    z.object({
+      revision_id: z.string(),
+      revision_number: z.number(),
+      instance_status_list: z.array(
+        z.object({
+          status: z.union([
+            z.literal("PENDING"),
+            z.literal("RUNNING"),
+            z.literal("FAILED"),
+          ]),
```
ianedwards, line 33: `z.enum(['PENDING', ''RUNNING', 'FAILED'])`

dashboard/src/lib/hooks/useAppStatus.ts
```diff
@@ -1,215 +1,201 @@
...
+  const [serviceStatusMap, setServiceStatusMap] = useState<
+    Record<string, SerializedServiceStatus>
+  >({});
+
+  const { newWebsocket, openWebsocket, closeAllWebsockets, closeWebsocket } =
+    useWebsockets();
+
+  const setupWebsocket = (serviceName: string): void => {
+    const selectors = `porter.run/service-name=${serviceName},porter.run/deployment-target-id=${deploymentTargetId}`;
+    const apiEndpoint = `/api/projects/${projectId}/clusters/${clusterId}/apps/${kind}/status?selectors=${selectors}`;
```
ianedwards, line 67: just want to confirm functionality here - we just use the websocket to get pinged about changes in the pod status, and then update the service status?

internal/porter_app/status.go
```diff
@@ -0,0 +1,245 @@
...
+	if inp.DeploymentTarget.Namespace == "" {
+		return serviceStatus, telemetry.Error(ctx, span, nil, "must provide deployment target namespace")
+	}
+
+	selectorString := fmt.Sprintf(
+		"%s=%s,%s=%s,%s=%s",
+		LabelKey_DeploymentTargetID, inp.DeploymentTarget.ID,
+		LabelKey_AppName, inp.AppName,
+		LabelKey_ServiceName, inp.ServiceName,
+	)
```
ianedwards, line 104: could be nice to have a helper for this

ianedwards (approved):

dashboard/src/lib/hooks/useAppStatus.ts
ferozemohideen: yeah exactly, that's the behavior that we had before as well as legacy i'm pretty sure

PR #4108, opened 2023-12-19
display A record when IP

d-g-town:


sdess09 (approved):

dashboard/src/utils/ip.ts
```diff
@@ -0,0 +1,17 @@
...
+
+export const isIP = (value: string): boolean => {
+    const ip = z.string().ip();
+
+    const parsed = ip.safeParse(value);
+
+    return parsed.success;
+};
+
+export const dnsRecordType = (value: string): string => {
```
ferozemohideen, line 11: nit: make this more extensible by returning a type instead of a string, then condition the text in the front-end based on that type

ferozemohideen (approved):

dashboard/src/utils/ip.ts
d-g-town: I'll do this if we start using it for more things, right now I just wanted to keep it simple for the people calling it

PR #4111, opened 2023-12-20
show list of addons for preview targets

ianedwards:
## What does this PR do?

- Show users basic card for preview postgres
- This is just a start to indicate to users that the database exists - therefore the card won't be clickable until we have a screen for updating the addon after it has been created


api/server/handlers/addons/list.go
```diff
@@ -0,0 +1,108 @@
...
+	DeploymentTargetID string `schema:"deployment_target_id"`
+}
+
+// LatestAddonsResponse represents the response from the /addons/latest endpoint
+type LatestAddonsResponse struct {
+	Base64Addons []string `json:"base64_addons"`
+}
+
+func (c *LatestAddonsHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
+	ctx, span := telemetry.NewSpan(r.Context(), "serve-list-addons")
```
ferozemohideen, line 46: span name

api/server/handlers/addons/list.go
```diff
@@ -0,0 +1,108 @@
...
+	request := &LatestAddonsRequest{}
+	if ok := c.DecodeAndValidate(w, r, request); !ok {
+		err := telemetry.Error(ctx, span, nil, "error decoding request")
+		c.HandleAPIError(w, r, apierrors.NewErrPassThroughToClient(err, http.StatusBadRequest))
+		return
+	}
+
+	telemetry.WithAttributes(span,
+		telemetry.AttributeKV{Key: "project-id", Value: project.ID},
+		telemetry.AttributeKV{Key: "cluster-id", Value: cluster.ID},
```
ferozemohideen, line 61: are these still needed? i forget

api/server/handlers/addons/list.go
```diff
@@ -0,0 +1,108 @@
...
+		telemetry.AttributeKV{Key: "deployment-target-id", Value: request.DeploymentTargetID},
+	)
+
+	var deploymentTargetIdentifier *porterv1.DeploymentTargetIdentifier
+	if request.DeploymentTargetID != "" {
+		deploymentTargetIdentifier = &porterv1.DeploymentTargetIdentifier{
+			Id: request.DeploymentTargetID,
+		}
```
ferozemohideen, line 69: can you pass this in even if request.DeploymentTargetID == ""?

api/server/handlers/addons/list.go
```diff
@@ -0,0 +1,108 @@
...
+	}
+
+	res := &LatestAddonsResponse{
+		Base64Addons: []string{},
+	}
+
+	for _, addon := range latestAddonsResp.Msg.Addons {
+		by, err := helpers.MarshalContractObject(ctx, addon)
+		if err != nil {
+			err = telemetry.Error(ctx, span, err, "error marshaling addon")
```
ferozemohideen, line 98: maybe log the addon b64 before erroring (if it's not sensitive)

api/server/router/addons.go
```diff
@@ -0,0 +1,90 @@
...
+
+	newPath := &types.Path{
+		Parent:       basePath,
+		RelativePath: relPath,
+	}
+
+	var routes []*router.Route
+
+	// GET /api/projects/{project_id}/clusters/{cluster_id}/addons/latest -> addons.LatestAddonsHandler
+	latestAddonsEndpoint := factory.NewAPIEndpoint(
```
ferozemohideen, line 61: just want to put on the radar that we might want to introduce deployment target scoped routes (/projects/<>/clusters/<>/targets/<>/addons) so we don't have to keep passing it in as a query param

ferozemohideen (approved):

api/server/handlers/addons/list.go
ferozemohideen: or that might be correct actually, not sure if there's a difference between list all and list latest

api/server/router/addons.go
ianedwards: sounds good, should be a pretty easy refactor if needed

api/server/handlers/addons/list.go
ianedwards: it'll be nil if the ID is an empty string, which will cause for the cluster default to be used

api/server/handlers/addons/list.go
ianedwards: Do we have a different pattern here?

api/server/handlers/addons/list.go
ianedwards: probably not, I'll remove

api/server/handlers/addons/list.go
ianedwards: looks like when I do this in a loop then only the last is included, so it could be a bit confusing

PR #4118, opened 2024-01-02
Compliance dashboard

sdess09:
https://github.com/porter-dev/porter/assets/37374498/da4a0a5f-4d98-4f3b-970a-49ac52df6c3a


sdess09:
> Minor nitpick: if someone has something enabled and then disables but doesn't apply the change, should that still be a pending change? Should we differentiate between pending enabled/disabled states?

Good point Ill try doing so

PR #4124, opened 2024-01-05
[POR-2186] DB Polish Part 1

ferozemohideen:
old:
image

new: 
image


jusrhee:
Can we make the cards flush on the left hand side

jusrhee (approved):

PR #4128, opened 2024-01-05
Add a  Deleting Placeholder 

sdess09:


d-g-town (approved):
Definitely think this should also be added to the infra tab (I expect that a customer would try to answer the question "What is happening to my cluster infra?" by going to that tab)

PR #4133, opened 2024-01-09
[POR-2187] Refactor DB form backend

ferozemohideen:
This PR does not change the UI of the 3 db forms we currently have much - it mostly just cleans up current components and makes it easier to add a new db type in the future. Notably, I abstracted out all the logic involved with creating the database into its own hook, as well as form submission into its own provider


ianedwards (approved):
💯

PR #4134, opened 2024-01-09
parse env var references from porter.yaml

ianedwards:
## What does this PR do?

- Parses out env vars either as array of key/vals and reference objects, or as legacy map
- Uses different structs depending on how the value is defined
- Sends result back to CCP on proto object


internal/porter_app/v2/env.go
```diff
@@ -0,0 +1,231 @@
...
+	// EnvVariableSource_Value indicates that the env variable should be set to an explicitly provided value
+	EnvVariableSource_Value EnvVariableSource = "value"
+	// EnvVariableSource_FromApp indicates that the env variable should be set to a value from another app
+	EnvVariableSource_FromApp EnvVariableSource = "app"
+)
+
+// EnvVariableDefinition is a struct containing information about how to set an env variable
+type EnvVariableDefinition struct {
+	Key     string
+	Source  EnvVariableSource
```
stefanmcshane, line 25: Add some comments to these since they can seem generic

internal/porter_app/v2/env.go
```diff
@@ -0,0 +1,231 @@
...
+	}
+
+	if !def.Value.IsSet {
+		switch raw.From.Source {
+		case EnvVariableSource_FromApp:
+			def.Source = EnvVariableSource_FromApp
+
+			value, err := fromAppValue(raw.From.Value)
+			if err != nil {
+				return err
```
stefanmcshane, line 162: wrap this - custom unmarshaller errors can be tough to trace through. If there is too much info in the error, let whomever is logging it decide what to do

stefanmcshane (approved):

PR #4139, opened 2024-01-10
Require role access for credential updates

stefanmcshane:
Ensure that aws accounts can not be externally enumerated, unless a user has access to all projects which use that AWS account

### DIFF
--- a/api/server/handlers/project_integration/create_aws.go
+++ b/api/server/handlers/project_integration/create_aws.go
@@ -102,9 +102,10 @@ func (p *CreateAWSHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 				}
 			}
 
-			for _, required := range requiredProjects {
+			for proj, required := range requiredProjects {
 				if !required {
 					err = telemetry.Error(ctx, span, err, "user does not have access to all projects that use this AWS account")
+					telemetry.WithAttributes(span, telemetry.AttributeKV{Key: "missing-project", Value: proj})
 					p.HandleAPIError(w, r, apierrors.NewErrPassThroughToClient(err, http.StatusForbidden, "user does not have access to all projects that use this AWS account"))
 					return
 				}
--- a/api/server/handlers/project_integration/create_aws.go
+++ b/api/server/handlers/project_integration/create_aws.go
@@ -2,6 +2,7 @@ package project_integration
 
 import (
 	"net/http"
+	"strings"
 
 	"connectrpc.com/connect"
 	porterv1 "github.com/porter-dev/api-contracts/generated/go/porter/v1"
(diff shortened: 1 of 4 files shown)


api/server/handlers/project_integration/create_aws.go
```diff
@@ -70,6 +71,45 @@ func (p *CreateAWSHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 				return
 			}
 
+			// if a user is changing the external ID, then we need to update the external ID for all projects that use that AWS account.
+			// This is required since the same AWS account can be used across multiple projects. In order to change the external ID for a project,
+			// the user must then have access to all projects that use that AWS account.
+			// If we ever do a higher abstraction about porter projects, then we can tie the ability to access a cloud provider account to that higher abstraction.
+			awsAccountIdPrefix := strings.TrimPrefix(request.TargetArn, "arn:aws:iam::")
+			awsAccountId := strings.TrimPrefix(awsAccountIdPrefix, ":role/porter-manager")
```
d-g-town, line 79: TrimSuffix?

api/server/handlers/project_integration/create_aws.go
```diff
@@ -70,6 +71,45 @@ func (p *CreateAWSHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
+			}
+
+			for _, project := range usersProject {
+				if _, ok := requiredProjects[int(project.ID)]; ok {
+					requiredProjects[int(project.ID)] = true
+				}
+			}
+
+			for _, required := range requiredProjects {
+				if !required {
```
d-g-town, line 106: consider adding the project id to telemetry

api/server/handlers/project_integration/create_aws.go
```diff
@@ -70,6 +71,45 @@ func (p *CreateAWSHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 				return
 			}
 
+			// if a user is changing the external ID, then we need to update the external ID for all projects that use that AWS account.
+			// This is required since the same AWS account can be used across multiple projects. In order to change the external ID for a project,
+			// the user must then have access to all projects that use that AWS account.
+			// If we ever do a higher abstraction about porter projects, then we can tie the ability to access a cloud provider account to that higher abstraction.
+			awsAccountIdPrefix := strings.TrimPrefix(request.TargetArn, "arn:aws:iam::")
+			awsAccountId := strings.TrimPrefix(awsAccountIdPrefix, ":role/porter-manager")
+			roles, err := p.Repo().AWSAssumeRoleChainer().ListByAwsAccountId(ctx, awsAccountId)
```
d-g-town, line 80: nit: `assumeRoles` instead of `roles` only because it could be confused with the `roles` tables (especially since we then query for users)

d-g-town (approved):

api/server/handlers/project_integration/create_aws.go
stefanmcshane: done by default in porter

stefanmcshane: i see what you mean, nvm. Fixed

PR #4141, opened 2024-01-11
support waiting for update-image call

d-g-town:


api/server/handlers/porter_app/get_app_revision_status.go
```diff
@@ -35,9 +35,23 @@ func NewGetAppRevisionStatusHandler(
...
+	HighLevelStatus_Successful HighLevelStatus = "successful"
+	// HighLevelStatus_Failed indicates that the revision has failed
+	HighLevelStatus_Failed HighLevelStatus = "failed"
+)
+
 // GetAppRevisionStatusResponse represents the response from the /apps/{porter_app_name}/revisions/{app_revision_id}/status endpoint
 type GetAppRevisionStatusResponse struct {
 	AppRevisionStatus porter_app.RevisionProgress `json:"app_revision_status"`
+	// HighLevelStatus is a high level status that can be used to determine whether the revisions is progressing, successful or failed
+	HighLevelStatus HighLevelStatus `json:"high_level_status"`
```
ianedwards, line 54: If a user were to hit this endpoint I'd imagine `high_level_status` would be unexpected. Any reason this can't be `status`?

ianedwards (approved):

api/server/handlers/porter_app/get_app_revision_status.go
d-g-town: no reason, that makes much more sense

cli/cmd/v2/update_image.go
```diff
@@ -4,24 +4,83 @@ import (
...
+		triggeredBackgroundColor = color.FgBlue
+	}
+
+	_, _ = color.New(triggeredBackgroundColor).Printf("Updated application %s to use tag \"%s\"\n", input.AppName, tag)
+
+	if !input.WaitForSuccessfulUpdate {
+		return nil
+	}
+
+	timeoutMinutes := 10
```
ferozemohideen, line 54: nit: make const

ferozemohideen (approved):

PR #4144, opened 2024-01-12
introduce wait flag on apply that will wait for app revision status ready, also refactor update-tag

ferozemohideen:


d-g-town (approved):
love the refactor, but why call it outside of the v2 package? Seems like we could keep it inside the v2.Apply and v2.UpdateImage calls

PR #4148, opened 2024-01-16
Add Kubectl and Helm into the CLI Dockerfile

jdenquin:
## POR-
N/A
## What does this PR do?
- Add `kubectl` and `helm` into the `cli.Dockerfile` to be able to use them with the Github Action

### DIFF
--- a/docker/cli.Dockerfile
+++ b/docker/cli.Dockerfile
@@ -30,7 +30,15 @@ RUN make build-cli
 # Deployment environment
 # ----------------------
 FROM ubuntu:latest
-RUN apt-get update && apt-get install -y ca-certificates
+
+ENV KUBE_VERSION="v1.29.0"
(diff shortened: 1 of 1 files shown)


stefanmcshane:
Declining as we prefer to keep the image small, and rely on composition of tooling, instead of bundling. If using with github actions, consider using the Azure installs for helm and kubectl.

A sample manifest which will allow all further github action steps to include the porter cli, kubectl, and helm is as follows:

```yaml
    - uses: azure/setup-kubectl@v3

    - uses: azure/setup-helm@v3

    - name: Setup porter
       uses: porter-dev/setup-porter@v0.1.0

    - name: Deploy stack
      timeout-minutes: 30
      run: porter --version
      env:
        PORTER_CLUSTER: "1234"
        PORTER_HOST: https://dashboard.getporter.dev/
        ...
```

PR #4155, opened 2024-01-17
quick fix for disabling registries in diff cloud

ianedwards:
## What does this PR do?
Screenshot 2024-01-17 at 10 11 05 AM

### DIFF
--- a/dashboard/src/main/home/modals/IntegrationsModal.tsx
+++ b/dashboard/src/main/home/modals/IntegrationsModal.tsx
@@ -54,7 +54,6 @@ export default class IntegrationsModal extends Component<PropsType, StateType> {
   renderIntegrationsCatalog = () => {
     if (this.context.currentModalData) {
       const { setCurrentIntegration } = this.context.currentModalData;
-      console.log(this.state.integrations, this.context.currentCluster);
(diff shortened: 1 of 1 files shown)


dashboard/src/main/home/modals/IntegrationsModal.tsx
```diff
@@ -18,40 +20,73 @@ export default class IntegrationsModal extends Component<PropsType, StateType> {
...
+          integration.auth_mechanism;
 
         if (!disabled) {
-          return (
-            <IntegrationOption
+          return !sameCloudProvider ? (
+            <Tooltip
               key={i}
-              disabled={disabled}
+              content="Must be on the same cloud provider as your current cluster"
```
ferozemohideen, line 69: maybe "Can only connect registries hosted in the same cloud provider your cluster is provisioned with" if that's not too wordy

ferozemohideen (approved):

dashboard/src/main/home/modals/IntegrationsModal.tsx
ianedwards: trimmed it down a bit to fit in the tooltip

PR #4156, opened 2024-01-17
set image override for manual service run if provided

ianedwards:

### DIFF
--- a/api/server/handlers/porter_app/run_app_job.go
+++ b/api/server/handlers/porter_app/run_app_job.go
@@ -5,7 +5,6 @@ import (
 
 	"github.com/porter-dev/porter/api/server/authz"
 	"github.com/porter-dev/porter/api/server/shared/requestutils"
-	"k8s.io/utils/pointer"
 
 	"connectrpc.com/connect"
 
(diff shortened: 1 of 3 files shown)


api/server/handlers/porter_app/run_app_job.go
```diff
@@ -86,11 +91,30 @@ func (c *RunAppJobHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	}
 	telemetry.WithAttributes(span, telemetry.AttributeKV{Key: "deployment-target-id", Value: request.DeploymentTargetID})
 
+	var commandOptional *string
+	if request.RunCommand != "" {
+		commandOptional = pointer.String(request.RunCommand)
```
d-g-town, line 96: nit: why not just use `&request.RunCommand`?

api/server/handlers/porter_app/run_app_job.go
```diff
@@ -86,11 +91,30 @@ func (c *RunAppJobHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
+	}
+
+	var imageOverrideOptional *porterv1.AppImage
+	if request.Image.Repository != "" || request.Image.Tag != "" {
```
d-g-town, line 100: personally don't like the idea of just changing repository and not tag

d-g-town (approved):

api/server/handlers/porter_app/run_app_job.go
ianedwards: fair

api/server/handlers/porter_app/run_app_job.go
ianedwards: don't have a preference; remember having used this helper previously

PR #4170, opened 2024-01-19
[POR-2202] allow users to link apps to their datastore from the dashboard

ferozemohideen:
https://github.com/porter-dev/porter/assets/40551216/22a2e805-6b29-4b0a-ae84-d75484577923


dashboard/src/lib/hooks/useLatestAppRevisions.ts
```diff
@@ -0,0 +1,59 @@
...
+
+      return apps.app_revisions;
+    },
+    {
+      refetchOnWindowFocus: false,
+      enabled: clusterId !== 0 && projectId !== 0,
+    }
+  );
+  return {
+    revisions: apps ?? [],
```
ianedwards, line 57: should be unnecessary to have this ternary if there's the default when de-structuring but not sure if you're seeing type errors without this

ianedwards (approved):

PR #4183, opened 2024-01-23
[POR-2218] CLI changes to support datastore connect

ferozemohideen:
how it works:
1. get datastore row from the table
2. get the cluster to pull datastore credentials from
3. parse the datastore credentials according to the datastore type
4. use the credentials (host name, port) to create a proxy pod in the cluster from 2, which connects to the datastore
5. print out the credentials
6. portforward to the proxy pod


api/client/datastore.go
```diff
@@ -0,0 +1,28 @@
...
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/porter-dev/porter/api/types"
+)
+
+// GetDatastore gets a datastore by name for a project
+func (c *Client) GetDatastore(
```
d-g-town, line 11: nit: don't call this `Get`

cli/cmd/commands/datastore.go
```diff
@@ -0,0 +1,419 @@
...
+	return forwardPorts("POST", req.URL(), config.RestConf, []string{fmt.Sprintf("%d:%d", port, datastoreCredential.Port)}, stopChannel, readyChannel)
+}
+
+type datastoreCredentialInput struct {
+	DatastoreName string
+	DatastoreType types.DatastoreType
+	Clientset     *kubernetes.Clientset
+}
+
+func datastoreCredential(inp datastoreCredentialInput) (types.DatastoreCredential, error) {
```
d-g-town, line 286: this should happen server-side

cli/cmd/commands/datastore.go
```diff
@@ -0,0 +1,419 @@
...
+			},
+		},
+	}
+
+	s = spinner.New(spinner.CharSets[9], 100*time.Millisecond)
+	s.Color("cyan") // nolint:errcheck,gosec
+	s.Suffix = " Creating secure tunnel to datastore..."
+
+	s.Start()
+	proxyPod, err := config.Clientset.CoreV1().Pods(Namespace_Default).Create(context.TODO(), pod, metav1.CreateOptions{})
```
stefanmcshane, line 235: shouldnt use context.Todo

cli/cmd/commands/datastore.go
```diff
@@ -0,0 +1,195 @@
...
+	datastoreConnectCmd := &cobra.Command{
+		Use:   "connect <DATASTORE_NAME>",
+		Short: "Forward a local port to a remote datastore.",
+		Args:  cobra.MinimumNArgs(1),
+		Run: func(cmd *cobra.Command, args []string) {
+			err := checkLoginAndRunWithConfig(cmd, cliConf, args, datastoreConnect)
+			if err != nil {
+				os.Exit(1)
+			}
+		},
```
d-g-town, line 46: consider using flags for datastore name instead of args

PR #4200, opened 2024-01-25
add telemetry to slack integration

d-g-town:


internal/models/notification.go
```diff
@@ -17,6 +17,8 @@ type NotificationConfig struct {
 
 	LastNotifiedTime time.Time
 	NotifLimit       string
+
+	Config JSONB `json:"config" sql:"type:jsonb" gorm:"type:jsonb"`
```
ferozemohideen, line 21: what is this

ferozemohideen (approved):

internal/models/notification.go
d-g-town: for later

d-g-town: removed

PR #4210, opened 2024-01-29
if deployment target is set as query param, set cluster to match

d-g-town:


dashboard/src/shared/DeploymentTargetContext.tsx
```diff
@@ -101,6 +100,29 @@ const DeploymentTargetProvider = ({
...
+          cluster_id: deploymentTarget.cluster_id,
+        }
+      )
+      .then(({ data }) => {
+        if (data) {
+          setCurrentCluster(data);
+        }
+      })
+      .catch(() => {});
+  }
```
ianedwards, line 124: shouldn't really just call a function in a component. It'd be ideal to just have a useQuery that makes this call, which will only refetch on changes to one of the values in the query key.

From that point, update your cluster data in a useEffect when  `deploymentTarget.cluster_id !== currentCluster.id`

ianedwards (approved):

PR #4226, opened 2024-02-02
enable azure gpu on frontend

d-g-town:


dashboard/src/components/AzureProvisionerSettings.tsx
```diff
@@ -471,6 +505,50 @@ const AzureProvisionerSettings: React.FC<Props> = (props) => {
...
+        </Button>
+
+        <Spacer y={0.5} />
+      </>
+    );
+  };
+
+  if (props.gpuModal) {
+    return renderGPUSettings();
+  }
```
ianedwards, line 550: can just return the jsx directly instead of wrapping it in a function

ianedwards (approved):
are we going to continue using the modal to expose this?

PR #4257, opened 2024-02-09
resolve app env comments

jusrhee:
## What does this PR do?
resolving commenst from app env update/overhaul


dashboard/src/main/home/env-dashboard/EnvGroupArray.tsx
```diff
@@ -31,12 +31,12 @@ type PropsType = {
 const EnvGroupArray = ({
   label,
   values,
-  setValues,
+  setValues=()=>{},
```
ferozemohideen, line 34: don't need to do this if setValues is always defined now, just need it if you kept it as setValues?: ... at the top

ferozemohideen (approved):

PR #4269, opened 2024-02-11
New list contracts backend

ferozemohideen:
allow for specifying a cluster id and latest flags

### DIFF
--- a/internal/repository/gorm/api_contract.go
+++ b/internal/repository/gorm/api_contract.go
@@ -116,28 +116,27 @@ func (cr APIContractRepository) Latest(ctx context.Context, projectID uint, filt
 		opt(&opts)
 	}
 
-	if opts.ClusterID == 0 {
-		queryString := `
+	queryString := `
 		SELECT DISTINCT ON (cluster_id) *
 		FROM api_contract_revisions
 		WHERE project_id = ?
 		ORDER BY cluster_id, created_at DESC
-	`
-		tx := cr.db.Raw(queryString, projectID).Scan(&confs)
-		if tx.Error != nil {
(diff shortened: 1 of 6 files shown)


internal/repository/gorm/api_contract.go
```diff
@@ -85,3 +106,39 @@ func (cr APIContractRepository) Get(ctx context.Context, revisionID uuid.UUID) (
...
+		SELECT DISTINCT ON (cluster_id) *
+		FROM api_contract_revisions
+		WHERE project_id = ? AND cluster_id = ?
+		ORDER BY cluster_id, created_at DESC
+	`
+		tx := cr.db.Raw(queryString, projectID, opts.ClusterID).Scan(&confs)
+		if tx.Error != nil {
+			return nil, tx.Error
+		}
+	}
```
d-g-town, line 141: remove else

```
queryString := `
		SELECT DISTINCT ON (cluster_id) *
		FROM api_contract_revisions
		WHERE project_id = ?
		ORDER BY cluster_id, created_at DESC
	`
args = []any{projectID}

if opts.ClusterID != 0 {
queryString := `
		SELECT DISTINCT ON (cluster_id) *
		FROM api_contract_revisions
		WHERE project_id = ?
		ORDER BY cluster_id, created_at DESC
	`
    args = append(args, opts.ClusterID)
}

tx := cr.db.Raw(queryString, args...).Scan(&confs)
if tx.Error != nil {
	return nil, tx.Error
}
```

d-g-town (approved):
at some point this should move to ccp, right? I'm assuming it was just easier to add the logic to the existing call flow here

PR #4290, opened 2024-02-16
Cluster advanced settings

ferozemohideen:
image


jusrhee:
In general, don't think it makes sense to artificially clamp the width on these really long descriptions:
image

jusrhee (approved):

PR #4299, opened 2024-02-20
update log structure used by api clients

ianedwards:
## What does this PR do?

- Giving a more defined structure to logs when they're initially loaded in, so that we can expose this API more generally and not ask consumers to parse raw JSON from Loki
- FE changes are for accommodating the object format change
- also adding in a filter for job run name
- This PR does not update the websocket output, as Loki writes directly to the output stream at the moment

### DIFF
--- a/api/server/handlers/porter_app/logs_apply_v2.go
+++ b/api/server/handlers/porter_app/logs_apply_v2.go
@@ -206,7 +206,6 @@ func (c *AppLogsHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 
 	logs, err := porter_agent.Logs(ctx, k8sAgent.Clientset, agentSvc, logRequest)
 	if err != nil {
-		fmt.Printf("error getting logs: %v\n", err)
 		_ = telemetry.Error(ctx, span, err, "unable to get logs")
 		c.HandleAPIError(w, r, apierrors.NewErrPassThroughToClient(fmt.Errorf("unable to get logs"), http.StatusInternalServerError))
 		return
--- a/api/server/handlers/porter_app/logs_apply_v2.go
+++ b/api/server/handlers/porter_app/logs_apply_v2.go
@@ -15,6 +15,7 @@ import (
 	"github.com/porter-dev/porter/internal/deployment_target"
 	porter_agent "github.com/porter-dev/porter/internal/kubernetes/porter_agent/v2"
 	"github.com/porter-dev/porter/internal/models"
+	"github.com/porter-dev/porter/internal/porter_app"
 	"github.com/porter-dev/porter/internal/telemetry"
 )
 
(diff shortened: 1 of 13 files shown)


internal/porter_app/logs.go
```diff
@@ -0,0 +1,69 @@
...
+			OutputStream: log.Metadata.OutputStream,
+		}
+
+		if log.Timestamp != nil {
+			structuredLog.Timestamp = *log.Timestamp
+		}
+		if serviceName, ok := log.Metadata.RawLabels[lokiLabel_PorterServiceName]; ok {
+			structuredLog.ServiceName = serviceName
+		}
+		if appRevisionID, ok := log.Metadata.RawLabels[lokiLabel_PorterAppRevisionID]; ok {
```
d-g-town, line 49: no need to check ok every time, right? Because the behavior will be the nil value either way

dashboard/src/main/home/app-dashboard/validate-apply/logs/utils.ts
```diff
@@ -238,32 +267,37 @@ export const useLogs = ({
...
   };
 
-  const filterLogs = (logs: PorterLog[]) => {
-    return logs.filter(log => {
-      if (log.metadata == null) {
-        return true;
-      }
-
-      if (jobRunID !== "" && log.metadata.raw_labels?.controller_uid !== jobRunID) {
+  const filterLogs = (logs: PorterLog[]): PorterLog[] => {
```
ferozemohideen, line 270: can we do this filtering on the backend? it's sometimes confusing debugging issues where the customer tells us they don't see logs because these filters silently apply without us knowing. Would be great if whatever I see on the network tab is what I expect to see in the frontend

dashboard/src/main/home/app-dashboard/validate-apply/logs/utils.ts
```diff
@@ -294,50 +327,55 @@ export const useLogs = ({
...
+        .object({
+          logs: z.array(porterLogValidator),
+          backward_continue_time: z.string().nullable(),
+          forward_continue_time: z.string().nullable(),
+        })
+        .safeParse(logsResp.data);
 
-      if (logsResp.data == null) {
+      if (!parsedRes.success) {
+        console.log("parsedRes.error", parsedRes.error);
```
ferozemohideen, line 345: remove

api/server/handlers/porter_app/logs_apply_v2.go
```diff
@@ -116,14 +137,32 @@ func (c *AppLogsHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	namespace := deploymentTarget.Namespace
 	telemetry.WithAttributes(span, telemetry.AttributeKV{Key: "namespace", Value: namespace})
 
-	if request.StartRange.IsZero() || request.EndRange.IsZero() {
-		err := telemetry.Error(ctx, span, nil, "must provide start and end range")
-		c.HandleAPIError(w, r, apierrors.NewErrPassThroughToClient(err, http.StatusBadRequest))
-		return
+	startRange := request.StartRange
+	if request.StartRange.IsZero() {
+		dayAgo := time.Now().Add(-24 * time.Hour).UTC()
```
d-g-town, line 142: heads up that we might have to reduce this default time frame (a couple months ago I was dealing with an issue where loki timed out trying to grab too many logs). I say we go for this for now though and adjust if necessary

d-g-town (approved):

ferozemohideen (approved):

internal/porter_app/logs.go
ianedwards: I was just doing this for clarity but if it's verbose I can remove it

dashboard/src/main/home/app-dashboard/validate-apply/logs/utils.ts
ianedwards: moving the ones that I know are supported to the backend, but I don't think output stream is currently supported

PR #4319, opened 2024-02-22
Add Cidr table

stefanmcshane:
- **Remove unused tests**
- **Add cidr table**


internal/models/ipam.go
```diff
@@ -0,0 +1,27 @@
+package models
+
+import (
+	"net"
+
+	"github.com/google/uuid"
+	"gorm.io/gorm"
+)
+
+// Ipam represents a an entry in the Ipam table
```
ferozemohideen, line 10: an

ferozemohideen (approved):

PR #4331, opened 2024-02-26
new preflight check endpoint/handler for AWS

ferozemohideen:

### DIFF
--- a/api/server/router/project.go
+++ b/api/server/router/project.go
@@ -1543,6 +1543,34 @@ func getProjectRoutes(
 		Router:   r,
 	})
 
+	// POST /api/projects/{project_id}/contract/preflight_check -> apiContract.NewPreflightCheckHandler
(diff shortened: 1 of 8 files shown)


api/server/router/project.go
```diff
@@ -1543,6 +1543,34 @@ func getProjectRoutes(
...
 	})
 
+	// POST /api/projects/{project_id}/contract/preflight_check -> apiContract.NewPreflightCheckHandler
+	preflightCheckEndpoint := factory.NewAPIEndpoint(
+		&types.APIRequestMetadata{
+			Verb:   types.APIVerbCreate,
+			Method: types.HTTPVerbPost,
+			Path: &types.Path{
+				Parent:       basePath,
+				RelativePath: fmt.Sprintf("%s/contract/preflight_check", relPath),
```
d-g-town, line 1553: nit: I would just call it `preflight` only because I don't like the underscore

d-g-town (approved):

PR #4339, opened 2024-02-28
support worker health command on frontend

d-g-town:


ianedwards (approved):

internal/porter_app/v2/yaml.go
```diff
@@ -215,6 +215,7 @@ type Domains struct {
 type HealthCheck struct {
 	Enabled  bool   `yaml:"enabled"`
 	HttpPath string `yaml:"httpPath"`
+	Command  string `yaml:"command"`
```
ferozemohideen, line 218: update the comment of this type

dashboard/src/lib/porter-apps/services.ts
```diff
@@ -261,6 +263,7 @@ export function defaultSerialized({
       config: {
         type: "worker" as const,
         autoscaling: defaultAutoscaling,
+        healthCheck: defaultHealthCheck,
```
ferozemohideen, line 266: shouldn't the default health check for a worker have the command rather than http path?

dashboard/src/main/home/app-dashboard/validate-apply/services-settings/tabs/WorkerTabs.tsx
```diff
@@ -59,6 +60,7 @@ const WorkerTabs: React.FC<Props> = ({
         ))
         .with("advanced", () => (
           <>
+            <Health index={index} command={true} />
```
ferozemohideen, line 63: no need to pass in command={true}

dashboard/src/main/home/app-dashboard/validate-apply/services-settings/tabs/Health.tsx
```diff
@@ -58,20 +59,34 @@ const Health: React.FC<HealthProps> = ({ index }) => {
...
-          <ControlledInput
-            type="text"
-            placeholder="ex: /healthz"
-            {...register(
-              `app.services.${index}.config.healthCheck.httpPath.value`
-            )}
-          />
-        </>
-      )}
+      {healthCheckEnabled.value &&
```
ferozemohideen, line 62: no need for the command prop if we just condition on the service config

PR #4341, opened 2024-02-28
small azure patches

d-g-town:


dashboard/src/components/porter/Link.tsx
```diff
@@ -11,6 +12,7 @@ type Props = {
   color?: string;
   hoverColor?: string;
   showTargetBlankIcon?: boolean;
+  removeInline?: boolean;
```
ferozemohideen, line 15: removeInline is an anti-pattern; should just be inline

dashboard/src/main/home/infrastructure-dashboard/ClusterFormContextProvider.tsx
```diff
@@ -130,6 +130,13 @@ const ClusterFormContextProvider: React.FC<ClusterFormContextProviderProps> = ({
     errors,
   ]);
 
+  const formatContractErrorMessage = (errorMessage: string): string => {
```
ferozemohideen, line 133: this should be in updateCluster() imo

ferozemohideen (changes requested):

ferozemohideen (approved):

PR #4343, opened 2024-02-28
remove beta features enabled, add cluster vanity name, id, remove delete gpu

ferozemohideen:

### DIFF
--- a/dashboard/src/main/home/infrastructure-dashboard/shared/NodeGroups.tsx
+++ b/dashboard/src/main/home/infrastructure-dashboard/shared/NodeGroups.tsx
@@ -29,7 +29,7 @@ const NodeGroups: React.FC<Props> = ({ availableMachineTypes }) => {
   const {
     fields: nodeGroups,
     append,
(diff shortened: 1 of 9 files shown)


d-g-town (approved):

dashboard/src/main/home/infrastructure-dashboard/shared/NodeGroups.tsx
```diff
@@ -255,24 +255,24 @@ const I = styled.i`
...
-  > span {
-    font-size: 20px;
-  }
-  margin-right: 5px;
-`;
+//   > span {
+//     font-size: 20px;
+//   }
+//   margin-right: 5px;
+// `;
```
d-g-town, line 278: Are these comments intentional? Or should it be deleted?

ferozemohideen: intentional to pass lint - plan on uncommenting this when we can actually delete the gpu node

PR #4362, opened 2024-03-02
handle unknown preflight checks on frontend

ferozemohideen:
image


dashboard/src/lib/clusters/types.ts
```diff
@@ -507,22 +507,20 @@ export const clusterContractValidator = z.object({
...
-  "vcpu",
-  "vpc",
-  "natGateway",
-  "apiEnabled",
-  "cidrAvailability",
-  "iamPermissions",
-  "authz",
-  "enforceCidrUniqueness",
-]);
-type PreflightCheckKey = z.infer<typeof preflightCheckKeyValidator>;
```
ianedwards, line 521: another option, depending on what's more convenient: https://github.com/colinhacks/zod/discussions/1723#discussioncomment-4449900

ferozemohideen: nice i like it

PR #4390, opened 2024-03-07
poll for cloud provider permission status

ferozemohideen:
only supporting aws for now, azure and gcp can continue using the old endpoint because their perm update is sync

### DIFF
--- a/api/server/handlers/project_integration/get_cloud_provider_permissions_status.go
+++ b/api/server/handlers/project_integration/get_cloud_provider_permissions_status.go
@@ -38,28 +38,9 @@ const (
 	CloudProviderAzure CloudProviderType = "Azure"
 )
 
-type Aws struct {
-	TargetArn  string `schema:"target_arn"`
-	ExternalID string `schema:"external_id"`
-}
-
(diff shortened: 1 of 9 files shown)


api/server/handlers/project_integration/get_cloud_provider_permissions_status.go
```diff
@@ -0,0 +1,202 @@
...
+type CloudProviderType string
+
+const (
+	CloudProviderAWS   CloudProviderType = "AWS"
+	CloudProviderGCP   CloudProviderType = "GCP"
+	CloudProviderAzure CloudProviderType = "Azure"
+)
+
+type Aws struct {
+	TargetArn  string `schema:"target_arn"`
```
stefanmcshane, line 42: What does the schema tag do?

stefanmcshane (approved):

api/server/handlers/project_integration/get_cloud_provider_permissions_status.go
ferozemohideen: this allows us to read in query params for get requests

PR #4392, opened 2024-03-07
only block app creations for apps that have a revision

d-g-town:


api/server/handlers/porter_app/app_instances.go
```diff
@@ -0,0 +1,109 @@
...
+	}
+
+	protoAppInstances := latestAppInstancesResp.Msg.AppInstances
+	if protoAppInstances == nil {
+		protoAppInstances = []*porterv1.AppInstance{}
+	}
+
+	var appInstances []porter_app.AppInstance
+
+	for _, instance := range protoAppInstances {
```
ianedwards, line 97: I think you can `range` on a nil slice

api/server/handlers/porter_app/update_app.go
```diff
@@ -239,6 +239,10 @@ func (c *UpdateAppHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	// err = telemetry.Error(ctx, span, err, "blah blah blah")
+	// c.HandleAPIError(w, r, apierrors.NewErrPassThroughToClient(err, http.StatusInternalServerError))
+	// return
```
ianedwards, line 244: still want this?

ianedwards (approved):

Issue #3896, opened 2023-10-30
Migrate away from deprecated `heroku/buildpacks:20` and `heroku/builder-classic:22` CNB builder images


edmorley:
Hi 👋

This project uses Heroku's `heroku/buildpacks:20` and `heroku/builder-classic:22` CNB builder images:
https://github.com/search?q=repo%3Aporter-dev%2Fporter+%28%22heroku%2Fbuilder-classic%3A22%22+OR+%22heroku%2Fbuildpacks%3A20%22%29&type=code

We have deprecated this builder image as of https://github.com/heroku/cnb-builder-images/pull/429.

Please migrate to `heroku/builder:22` or `heroku/builder:20`, to continue to receive security updates, and to avoid disruption when the build time deprecation warning is upgraded to an error in the future.

More information on the difference between the various builder images can be seen here:
https://github.com/heroku/cnb-builder-images#heroku-cnb-builder-images

I also see there are references to the even older (and EOL/insecure) `heroku/buildpacks:18` - these need to be migrated too:
https://github.com/search?q=repo%3Aporter-dev%2Fporter+%22heroku%2Fbuildpacks%3A18%22&type=code

The docs here will also need updating:
https://github.com/search?q=repo%3Aporter-dev%2Fdocs+%22heroku%2Fbuildpacks%22&type=code

edmorley:
@jose-fully-ported Gentle nudge that Porter is still using these deprecated images - builds using these images will start erroring in the future (likely in the next couple of months; there will be an env var opt out), and the underlying images will stop receiving security updates.

edmorley:
@jose-fully-ported @ferozemohideen @stefanmcshane ^

edmorley:
> and to avoid disruption when the build time deprecation warning is upgraded to an error in the future.

This will be happening within the next couple of weeks in https://github.com/heroku/cnb-builder-images. We'll be adding an env var override for the errors, but I strongly recommend you switch from `heroku/buildpacks:20` and `heroku/builder-classic:22` to `heroku/builder:*` as soon as possible.

edmorley:
> Please migrate to `heroku/builder:22` or `heroku/builder:20`, to continue to receive security updates, and to avoid disruption when the build time deprecation warning is upgraded to an error in the future.

The deprecation warnings were upgraded to errors this morning:
https://github.com/heroku/cnb-builder-images/pull/478

edmorley:
[...]

PR #4420, opened 2024-03-18
inject env var for deprecated builder

ferozemohideen:

### DIFF
--- a/cli/cmd/v2/build.go
+++ b/cli/cmd/v2/build.go
@@ -151,6 +151,13 @@ func build(ctx context.Context, client api.Client, inp buildInput) buildOutput {
 			Buildpacks: inp.BuildPacks,
 		}
 
+		if buildConfig.Builder == "heroku/buildpacks:20" {
+			if opts.Env == nil {
+				opts.Env = map[string]string{}
+			}
+			opts.Env["ALLOW_EOL_SHIMMED_BUILDER"] = "1"
+		}
+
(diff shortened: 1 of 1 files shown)


ianedwards (approved):

cli/cmd/v2/build.go
```diff
@@ -153,6 +153,13 @@ func build(ctx context.Context, client api.Client, inp buildInput) buildOutput {
 			Buildpacks: inp.BuildPacks,
 		}
 
+		if buildConfig.Builder == "heroku/buildpacks:20" {
+			if opts.Env == nil {
+				opts.Env = map[string]string{}
+			}
+			opts.Env["ALLOW_EOL_SHIMMED_BUILDER"] = "1"
```
edmorley, line 160: I would not recommend injecting this by default. Users may miss the EOL warning and not realise they need to upgrade ASAP.

As mentioned in https://github.com/heroku/cnb-builder-images/issues/474 these images were superseded way back in June 2022, and we've been showing deprecation warnings in the build logs for 6 months now (we even attempted outreach about this in #3896).

edmorley: (These images will soon receive zero security updates, and at some further point after that may even be deleted)

edmorley: @ferozemohideen @ianedwards @jose-fully-ported ^

PR #4443, opened 2024-03-25
check feature flag

jusrhee:
## What does this PR do?
ensure billing modal only blocks for projects where the feature flag is enabled


MauAraujo (approved):

jusrhee:
Not urgent but for some reason on initial create it took a while to register the payment method as the default:
image

PR #4444, opened 2024-03-25
Stripe integaration fixes

MauAraujo:
## POR-2413
This PR aims to fix some problems with the Stripe integration. It fixes:
- Users being able to delete the credit card after deploying
- Small bug where the customer was not being created when opening the modal on the apps page without opening the billing page.

With this PR, payment methods are properly set as default invoice settings on Stripe.


jusrhee:
Not urgent but for some reason on initial create it took a while to register the payment method as the default:
image

jusrhee (approved):

PR #4463, opened 2024-03-27
cancel job run from dashboard

ianedwards:
## What does this PR do?

- Manually cancel a job run from the dashboard

https://github.com/porter-dev/porter/assets/104453631/a9d139ec-5168-4b46-9c37-d0e3e6fc85b1


dashboard/src/main/home/app-dashboard/validate-apply/jobs/JobRunDetails.tsx
```diff
@@ -52,6 +67,45 @@ const JobRunDetails: React.FC<Props> = ({ jobRun }) => {
...
+
+      history.push(
+        `/apps/${appName}/job-history?service=${jobRun.service_name}`
+      );
+    } catch (err) {
+      if (axios.isAxiosError(err)) {
+        setJobRunCancelError(err.message);
+      } else {
+        setJobRunCancelError("An error occurred while canceling the job run.");
+      }
```
ferozemohideen, line 103: can use the getErrorMessage.. utility we have in useCluster (which we should move to a shared place)

ferozemohideen (approved):
nice, would be great to have this on the pre-deploy activity feed card eventually too

PR #4465, opened 2024-03-28
support target CLI

d-g-town:
Screenshot 2024-03-28 at 2 18 20 PM
Screenshot 2024-03-28 at 2 18 34 PM

### DIFF
--- a/api/client/deployment_target.go
+++ b/api/client/deployment_target.go
@@ -4,28 +4,53 @@ import (
 	"context"
 	"fmt"
 
-	"github.com/porter-dev/porter/api/server/handlers/deployment_target"
+	"github.com/porter-dev/porter/api/types"
 )
 
-// CreateDeploymentTarget creates a new deployment target for a given project and cluster with the provided name
+// DeploymentTarget retrieves a deployment target by id
+func (c *Client) DeploymentTarget(
+	ctx context.Context,
+	projectId uint,
+	targetIdentifer string,
+) (*types.ReadDeploymentTargetResponse, error) {
+	resp := &types.ReadDeploymentTargetResponse{}
+
+	err := c.getRequest(
+		fmt.Sprintf(
+			"/projects/%d/targets/%s", projectId, targetIdentifer),
+		nil,
+		resp,
+	)
+
+	return resp, err
+}
+
+// ListDeploymentTargets retrieves all deployment targets in a cluster
+func (c *Client) ListDeploymentTargets(
+	ctx context.Context,
+	projectId uint,
+) (*types.ListDeploymentTargetsResponse, error) {
+	resp := &types.ListDeploymentTargetsResponse{}
+
+	err := c.getRequest(
+		fmt.Sprintf("/projects/%d/targets", projectId),
+		nil,
+		resp,
+	)
+
+	return resp, err
+}
+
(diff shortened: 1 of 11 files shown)


stefanmcshane (approved):

cli/cmd/commands/target.go
```diff
@@ -0,0 +1,95 @@
...
+	w := new(tabwriter.Writer)
+	w.Init(os.Stdout, 3, 8, 0, '\t', tabwriter.AlignRight)
+
+	fmt.Fprintf(w, "%s\t%s\t%s\n", "ID", "NAME", "CLUSTER ID")
+
+	for _, target := range targets.DeploymentTargets {
+		fmt.Fprintf(w, "%s\t%s\t%d\n", target.ID, target.Name, target.ClusterID)
+	}
+
+	w.Flush()
```
ianedwards, line 92: only other thing that might be useful here is whether or not it's the default for a given cluster

ianedwards (approved):

api/client/deployment_target.go
```diff
@@ -4,28 +4,53 @@ import (
...
+	)
 
-	req := &deployment_target.CreateDeploymentTargetRequest{
-		Selector: selector,
-		Preview:  preview,
-	}
+	return resp, err
+}
+
+// ListDeploymentTargets retrieves all deployment targets in a cluster
```
ferozemohideen, line 28: in a project or cluster?

api/server/handlers/deployment_target/create.go
```diff
@@ -69,14 +56,24 @@ func (c *CreateDeploymentTargetHandler) ServeHTTP(w http.ResponseWriter, r *http
 		return
 	}
 
+	clusterId := request.ClusterId
+	if clusterOk {
```
ferozemohideen, line 60: I feel like if you use the cluster endpoint but also specify cluster id in the request, then request cluster id should take precedence

internal/repository/test/deployment_target.go
```diff
@@ -22,8 +22,13 @@ func (repo *DeploymentTargetRepository) DeploymentTargetBySelectorAndSelectorTyp
 	return nil, errors.New("cannot read database")
 }
 
+// ListForCluster returns all deployment targets for a project
+func (repo *DeploymentTargetRepository) ListForCluster(projectID uint, clusterID uint, preview bool) ([]*models.DeploymentTarget, error) {
```
ferozemohideen, line 26: would be nice to make cluster id and preview functional arguments

cli/cmd/commands/target.go
d-g-town: for sure: will add it in

api/server/handlers/deployment_target/create.go
d-g-town: if there is any auth associated with the cluster scope, I want to stick to that though

internal/repository/test/deployment_target.go
d-g-town: for sure, will add

d-g-town: actually the interface logic and gorm made this annoying: honestly the fix is to move this logic to CCP. Will do later

PR #4494, opened 2024-04-02
Metronome business logic

MauAraujo:
## What does this PR do?
This PR modifies the billing code to introduce the Metronome business logic alongside the existing Stripe integration. It modifies the existing Billing Manager to be more extensible and so that it encapsulates both the Stripe and Metronome clients.

It also adds the necessary logic for creating Metronome customers, adding and removing from plans, checking credit grants and creating the porter agent with the correct environment.

All the Metronome specific logic has been put behind the `metronome_enabled` feature flag, for testing it has to be turned on in the project.


api/server/shared/config/loader/loader.go
```diff
@@ -62,9 +63,11 @@ func sharedInit() {
 		panic(err)
 	}
 
-	InstanceBillingManager = &billing.StripeBillingManager{
-		StripeSecretKey:      InstanceEnvConf.ServerConf.StripeSecretKey,
-		StripePublishableKey: InstanceEnvConf.ServerConf.StripePublishableKey,
+	stripeClient := billing.NewStripeClient(InstanceEnvConf.ServerConf.StripeSecretKey, InstanceEnvConf.ServerConf.StripePublishableKey)
```
stefanmcshane, line 66: No need to rewrite, but try initialising these in a non-global location. Where are we initialising the connections to cluster control plane, and database?

api/types/billing.go
```diff
@@ -1,5 +1,9 @@
 package types
 
+import "github.com/google/uuid"
+
+// Stripe types
```
stefanmcshane, line 5: If you feel that the file is so large that it justifies comments like this, break the file into smaller files, for example `billing_stripe.go` and `billing_metronome.go`.

If the file doesnt feel large enough to break it down like this, then its likely that comments like `//Stripe types` and `// metronom types` are not needed

api/types/billing.go
```diff
@@ -10,3 +14,100 @@ type PaymentMethod = struct {
...
+}
+
+// CustomerArchiveRequest will archive the customer in Metronome.
+type CustomerArchiveRequest struct {
+	CustomerID uuid.UUID `json:"id"`
+}
+
+// BillingConfig is the configuration for the billing provider (Stripe, etc.)
+type BillingConfig struct {
+	BillingProviderType       string `json:"billing_provider_type"` // Required. Can be any of "aws_marketplace", "stripe", "netsuite", "custom", "azure_marketplace", "quickbooks_online", or "workday"
```
stefanmcshane, line 36: Prefer using field comments (`// BillingProviderType ...`)  as opposed to inline comments (at the end of the line here) where possible

api/types/billing.go
```diff
@@ -10,3 +14,100 @@ type PaymentMethod = struct {
...
+	BillingProviderCustomerID string `json:"billing_provider_customer_id"`
+	StripeCollectionMethod    string `json:"stripe_collection_method"` // Can be any of "charge_automatically" or "send_invoice"
+}
+
+// AddCustomerPlanRequest represents a request to add a customer plan with specific details.
+type AddCustomerPlanRequest struct {
+	PlanID              uuid.UUID `json:"plan_id"`                          // Required. The customer ID, plan ID, and date range for the plan to be applied.
+	StartingOn          string    `json:"starting_on"`                      // Required. RFC 3339 timestamp for when the plan becomes active for this customer. Must be at 0:00 UTC (midnight).
+	EndingBefore        string    `json:"ending_before,omitempty"`          // Optional. RFC 3339 timestamp for when the plan ends (exclusive) for this customer. Must be at 0:00 UTC (midnight).
```
stefanmcshane, line 45: If this must be UTC, consider adding UTC to the name `EndingBeforeUTC` for example. You cant guarantee that a developer will read the docs, but they _have_ to read the parameter names. More explicit parameter names are usually much easier to maintain and debug

api/types/billing.go
```diff
@@ -10,3 +14,100 @@ type PaymentMethod = struct {
...
+	NetPaymentTermsDays int       `json:"net_payment_terms_days,omitempty"` // Number of days after issuance of invoice after which the invoice is due (e.g., Net 30).
+}
+
+// AddCustomerPlanResponse is a response to the add customer plan request. Returns customer-plan relationship id.
+type AddCustomerPlanResponse struct {
```
stefanmcshane, line 50: If you need inline wrapping like this, only for a single request/response, define the wrapper inline on that request/response func

internal/billing/stripe.go
```diff
@@ -13,23 +13,31 @@ import (
...
-type StripeBillingManager struct {
-	StripeSecretKey      string
-	StripePublishableKey string
+type StripeClient struct {
+	SecretKey      string
+	PublishableKey string
+}
+
+// NewStripeClient creates a new client to call the Stripe API
+func NewStripeClient(secretKey string, publishableKey string) *StripeClient {
```
stefanmcshane, line 24: Must a pointer be used here? It looks like all values can be passed by copy. Default to not using pointers unless there is a specific reason that you _must_ use pointers. The code without pointers is usually safer, and definitely is less panic-prone

zarf/helm/.serverenv
```diff
@@ -67,8 +67,14 @@ HELM_APP_REPO_URL=http://chartmuseum:8080
 TELEMETRY_NAME=porter
 TELEMETRY_COLLECTOR_URL=otel-collector:4317
 
-# STRIPE_SECRET_KEY is required if billing is enabled
+# STRIPE_SECRET_KEY is used to create customers and payment method in Stripe. If empty all functionality will be disabled.
```
stefanmcshane, line 70: Thanks for adding this. How can I find a stripe key? Same with `Metronome api key`

api/server/shared/config/loader/loader.go
MauAraujo: Go it, I moved the initialization to where the CCP client is initialized.

api/types/billing.go
MauAraujo: Ok,  I split the types to separate files to have a more maintainable setup

api/types/billing.go
MauAraujo: Sure, I fixed and improved all comments in my latest commit

api/types/billing.go
MauAraujo: Great idea, added UTC to all fields that must be in UTC

api/types/billing.go
MauAraujo: You are right, moved the wrappings to where the responses are defined

zarf/helm/.serverenv
MauAraujo: You can get them from their respective dashboards, but if you don't have access I can send them to you.

PR #4502, opened 2024-04-04
Metronome feature flag fixes

MauAraujo:
## What does this PR do?
This PR adds some small fixes after merging the Metronome business logic PR so that the feature flag covers all logic. It also adds a small fix for an error in the credits endpoint when the feature flag is disabled.

### DIFF
--- a/api/server/handlers/billing/credits.go
+++ b/api/server/handlers/billing/credits.go
@@ -35,6 +35,7 @@ func (c *GetCreditsHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 
 	if !proj.GetFeatureFlag(models.MetronomeEnabled, c.Config().LaunchDarklyClient) {
 		c.WriteResult(w, r, "")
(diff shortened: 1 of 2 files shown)


api/server/handlers/billing/credits.go
```diff
@@ -35,6 +35,7 @@ func (c *GetCreditsHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 
 	if !proj.GetFeatureFlag(models.MetronomeEnabled, c.Config().LaunchDarklyClient) {
 		c.WriteResult(w, r, "")
+		return
```
stefanmcshane, line 38: Add some telemetry here to say that metronome FF is disabled.

stefanmcshane (approved):

api/server/handlers/billing/credits.go
MauAraujo: Added!

PR #4510, opened 2024-04-05
handle prometheus alerts

yosefmih:
handles prometheus alerts

### DIFF
--- a/api/server/handlers/webhook/prometheus_incoming.go
+++ b/api/server/handlers/webhook/prometheus_incoming.go
@@ -4,7 +4,6 @@ import (
 	"context"
 	"fmt"
 	"net/http"
-	"strconv"
 
 	"connectrpc.com/connect"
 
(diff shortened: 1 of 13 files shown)


api/server/handlers/webhook/prometheus_incoming.go
```diff
@@ -0,0 +1,126 @@
...
+	})
+	label_key_values := ""
+	for _, alert := range prometheusAlert.Alerts {
+		for k, v := range alert.Labels {
+			label_key_values += fmt.Sprintf("%s %s", k, v)
+		}
+		if alert.Labels["alertname"] == "NoopAlert" {
+			continue
+		}
+		desiredReplicas, err := strconv.ParseInt(alert.Labels["desiredReplicas"], 10, 32)
```
ferozemohideen, line 86: can we decode the Labels map into a struct?

api/server/handlers/webhook/prometheus_incoming.go
```diff
@@ -0,0 +1,126 @@
...
+	recordPrometheusAlertRequest := connect.NewRequest(&porterv1.RecordPrometheusAlertRequest{
+		ProjectId: projectId,
+		ClusterId: clusterId,
+	})
+	label_key_values := ""
+	for _, alert := range prometheusAlert.Alerts {
+		for k, v := range alert.Labels {
+			label_key_values += fmt.Sprintf("%s %s", k, v)
+		}
+		if alert.Labels["alertname"] == "NoopAlert" {
```
ferozemohideen, line 83: why does the request make it to the server if it is a no-op?

api/server/handlers/webhook/prometheus_incoming.go
```diff
@@ -0,0 +1,126 @@
...
+}
+
+func (p *PrometheusAlertWebhookHandler) handlePrometheusAlert(ctx context.Context, projectId, clusterId int64, prometheusAlert *types.PrometheusAlert) error {
+	ctx, span := telemetry.NewSpan(ctx, "porter-process-prom-alert")
+	defer span.End()
+	recordPrometheusAlertRequest := connect.NewRequest(&porterv1.RecordPrometheusAlertRequest{
+		ProjectId: projectId,
+		ClusterId: clusterId,
+	})
+	label_key_values := ""
```
ferozemohideen, line 78: prefer camelCase for this

ferozemohideen (approved):

api/server/handlers/webhook/prometheus_incoming.go
yosefmih: We use that as a heartbeat

api/server/handlers/webhook/prometheus_incoming.go
yosefmih: Hmm I will look into that, actually we are not setting the replica labels right now, removing this part of the code

yosefmih: But we basically have map[string]string so a struct would simply obfuscate that atm

PR #4522, opened 2024-04-10
refactoring main, adding separate authn/z providers

d-g-town:


ferozemohideen (approved):

dashboard/src/shared/auth/AuthnContext.tsx
```diff
@@ -0,0 +1,89 @@
...
+  const [isLoggedIn, setIsLoggedIn] = useState(false);
+  const [isEmailVerified, setIsEmailVerified] = useState(false);
+  const [isLoading, setIsLoading] = useState(true);
+  const [userId, setUserId] = useState(-1);
+  const [hasInfo, setHasInfo] = useState(false);
+
+  const authenticate = (): void => {
+    api
+      .checkAuth("", {}, {})
+      .then((res) => {
```
ianedwards, line 33: can we refactor these to return promises? just for consistency and so that we're not unexpectedly assuming an async task has finished

dashboard/src/main/Main.tsx
```diff
@@ -1,183 +1,117 @@
...
+    currentCluster,
+  } = useContext(Context);
+  const {
+    authenticate,
+    handleLogOut,
+    isLoggedIn,
+    isLoading,
+    hasInfo,
+    isEmailVerified,
+  } = useContext(AuthnContext);
```
ianedwards, line 39: one suggestion that goes along with my other comment - wrap main in AuthnContext, wait for authenticate to complete, if not authenticated then return something other than main (redirect, un-authed page, etc).

This comes with the benefit that we don't need a flag like `hasInfo` and can instead assume the user is logged in from any child component defined in the router of Main

ianedwards: we can then also have a user object provided by this context that is guaranteed to be defined

PR #4528, opened 2024-04-11
Porter Standard Plan, new billing page

MauAraujo:
## What does this PR do?
This PR introduces several changes to the backend and frontend so that the Metronome integration can be rolled out to Porter Standard users. The main changes are:

- Add logic for adding customers to Porter Standard Plan
- Fix credits endpoint to display sum of all grants
- Display total and remaining credits in billing page
- Add endpoint for requesting an embeddable dashboard to Metronome
- Add usage dashboard to billing page
- Add logic for retrieving plan details and trial information
- Add current plan information, as well as trial expiration to billing page
- Fix issue where the credits had incorrect type

The new billing page looks like this:
image

Porter Standard Users who have an active free trial will see the expiration:
image
image

The new usage dashboard from Metronome:
image

The embedded dashboard looks out of place because its inside an iframe, and there are limitations to what styling we can apply. So for now, its more of a temporary solution until we build a proper usage dashboard.


dashboard/src/lib/hooks/useStripe.tsx
```diff
@@ -154,8 +183,50 @@ export const checkIfProjectHasPayment = (): TCheckHasPaymentEnabled => {
   };
 };
 
+export const useCustomerDashboard = (dashboard: string): TGetUsageDashboard => {
```
ianedwards, line 186: maybe `useCustomerUsageDashboard`?

dashboard/src/lib/hooks/useStripe.tsx
```diff
@@ -154,8 +183,50 @@ export const checkIfProjectHasPayment = (): TCheckHasPaymentEnabled => {
...
+        "<token>",
+        {
+          dashboard,
+          color_overrides: colorOverrides,
+        },
+        {
+          project_id: currentProject?.id,
+        }
+      );
+      return res.data;
```
ianedwards, line 216: could validate the type on this response so its not any

dashboard/src/lib/hooks/useStripe.tsx
```diff
@@ -202,7 +273,33 @@ export const usePorterCredits = (): TGetCredits => {
...
+        return;
+      }
+      const res = await api.getCustomerPlan(
+        "<token>",
+        {},
+        {
+          project_id: currentProject?.id,
+        }
+      );
+      return res.data;
```
ianedwards, line 297: same here

dashboard/src/main/home/project-settings/BillingPage.tsx
```diff
@@ -40,10 +44,44 @@ function BillingPage(): JSX.Element {
...
+  const monthDiff = (d1: Date, d2: Date) => {
+    var months;
+    months = (d2.getFullYear() - d1.getFullYear()) * 12;
+    months -= d1.getMonth();
+    months += d2.getMonth();
+    return months <= 0 ? 0 : months;
+  };
+
+  const daysDiff = (d1: Date, d2: Date) => {
+    const _MS_PER_DAY = 1000 * 60 * 60 * 24;
```
ianedwards, line 62: there are libraries like `date-fns` that will do a lot of this for you for future reference

ianedwards (approved):

api/server/handlers/billing/plan.go
```diff
@@ -0,0 +1,160 @@
...
+	config *config.Config,
+	writer shared.ResultWriter,
+) *ListPlansHandler {
+	return &ListPlansHandler{
+		PorterHandlerWriter: handlers.NewDefaultPorterHandler(config, nil, writer),
+	}
+}
+
+func (c *ListPlansHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
+	ctx, span := telemetry.NewSpan(r.Context(), "list-plans-endpoint")
```
d-g-town, line 31: We have been naming every top-level ServeHTTP span with `serve-{rest of span name}`

d-g-town: so in this case, `serve-list-plans`

api/server/handlers/billing/plan.go
```diff
@@ -0,0 +1,160 @@
...
+		err := telemetry.Error(ctx, span, err, "error listing plans")
+		c.HandleAPIError(w, r, apierrors.NewErrInternal(err))
+		return
+	}
+
+	telemetry.WithAttributes(span,
+		telemetry.AttributeKV{Key: "metronome-enabled", Value: true},
+		telemetry.AttributeKV{Key: "project-id", Value: proj.ID},
+		telemetry.AttributeKV{Key: "usage-id", Value: proj.UsageID},
+	)
```
d-g-town, line 56: add these attributes as soon as you have the values. Otherwise, if there is an error in ListCustomerPlan, we won't have the `usage-id` field in the trace for debugging

Also, we have middleware that adds the `project-id` to the span whenever `types.ProjectScope` is in the context, so no need to add it again

api/server/handlers/billing/plan.go
```diff
@@ -0,0 +1,160 @@
...
+	config *config.Config,
+	writer shared.ResultWriter,
+) *ListCreditsHandler {
+	return &ListCreditsHandler{
+		PorterHandlerWriter: handlers.NewDefaultPorterHandler(config, nil, writer),
+	}
+}
+
+func (c *ListCreditsHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
+	ctx, span := telemetry.NewSpan(r.Context(), "list-credits-endpoint")
```
d-g-town, line 77: `serve-list-credits`

api/types/billing_metronome.go
```diff
@@ -60,6 +60,41 @@ type ListCreditGrantsRequest struct {
...
+
+// Plan is a pricing plan to which a user is currently subscribed
+type Plan struct {
+	ID                  uuid.UUID `json:"id"`
+	PlanID              uuid.UUID `json:"plan_id"`
+	PlanName            string    `json:"plan_name"`
+	PlanDescription     string    `json:"plan_description"`
+	StartingOn          string    `json:"starting_on"`
+	EndingBefore        string    `json:"ending_before"`
+	NetPaymentTermsDays int       `json:"net_payment_terms_days"`
```
d-g-town, line 89: nit: `Term` instead of `Terms`?

api/types/billing_metronome.go
```diff
@@ -68,26 +103,40 @@ type CreditType struct {
...
+	Name        string      `json:"name"`
+	GrantAmount GrantAmount `json:"grant_amount"`
+	Balance     Balance     `json:"balance"`
+	Reason      string      `json:"reason"`
+	EffectiveAt string      `json:"effective_at"`
+	ExpiresAt   string      `json:"expires_at"`
+}
+
+// DashboardOptions are optional dashboard specific options
+type DashboardOptions struct {
```
d-g-town, line 133: nit: `DashboardOption`, not plural

api/types/billing_metronome.go
```diff
@@ -68,26 +103,40 @@ type CreditType struct {
...
+	Key   string `json:"key"`
+	Value string `json:"value"`
+}
+
+// ColorOverrides is an optional list of colors to override
+type ColorOverrides struct {
```
d-g-town, line 139: likewise, `Override` not plural

dashboard/src/lib/billing/types.tsx
```diff
@@ -13,4 +12,25 @@ export const PaymentMethodValidator = z.object({
   is_default: z.boolean(),
 });
 
+type Trial = z.infer<typeof Trial>;
+
+const Trial = z.object({
+  ending_before: z.string(),
```
d-g-town, line 18: should this be a timestamp?

dashboard/src/lib/billing/types.tsx
```diff
@@ -13,4 +12,25 @@ export const PaymentMethodValidator = z.object({
...
+});
+
+export type Plan = z.infer<typeof Plan>;
+export const Plan = z.object({
+  id: z.string(),
+  plan_name: z.string(),
+  plan_description: z.string(),
+  starting_on: z.string(),
```
d-g-town, line 26: should this be a timestamp?

dashboard/src/lib/billing/types.tsx
MauAraujo: I tried setting as a timestamp but unfortunately zod doesn't have support for rfc3339 timestamps, and it fails if set to a datetime. Also doesn't look like their validation is very robust, and since this is mostly for displaying on the frontend I think it makes sense to keep as string

api/server/handlers/billing/plan.go
MauAraujo: fixed!

api/types/billing_metronome.go
MauAraujo: fixed!

api/types/billing_metronome.go
MauAraujo: fixed!

dashboard/src/lib/billing/types.tsx
MauAraujo: same as above

api/server/handlers/billing/plan.go
MauAraujo: Oh got it, thanks for the heads up. Fixed in latest commit.

api/server/handlers/billing/plan.go
MauAraujo: I see, fixed

dashboard/src/lib/hooks/useStripe.tsx
MauAraujo: fixed!

dashboard/src/lib/hooks/useStripe.tsx
MauAraujo: Yes, I missed validating these responses, thanks for pointing out.

dashboard/src/lib/hooks/useStripe.tsx
MauAraujo: fixed!

api/types/billing_metronome.go
MauAraujo: Metronome uses `Terms`, but you are right, `Term` is better. Fixed in latest commit

dashboard/src/main/home/project-settings/BillingPage.tsx
MauAraujo: I tried adding the library but the npm run check kept failing, it seems that we would need to update the package.json file or the eslint settings, which I don't want to touch in this PR. For now I will leave this functions and hopefully in the future we can use `date-fs`

PR #4532, opened 2024-04-14
add standardDCSv2Family family

d-g-town:


portersupport (approved):

dashboard/src/components/azureUtils.ts
```diff
@@ -326,6 +326,30 @@ const AzureMachineTypeOptions: MachineTypeOption[] = [
       "westus3",
     ]),
   },
+  {
```
ferozemohideen, line 329: these aren't used anymore - we should delete this file tbh

PR #4544, opened 2024-04-16
dgtown/azure machines

d-g-town:
- **allow porter user override**
- **allow porter user override**
- **refactor useMachineTypeList**

### DIFF
--- a/dashboard/src/lib/hooks/useNodeGroups.ts
+++ b/dashboard/src/lib/hooks/useNodeGroups.ts
@@ -24,7 +24,7 @@ export const useMachineTypeList = ({
 }): TUseMachineTypeList => {
   const { availableMachineTypes } = useClusterFormContext();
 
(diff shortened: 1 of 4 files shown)


dashboard/src/lib/hooks/useNodeGroups.ts
```diff
@@ -0,0 +1,68 @@
...
+      }
+    },
+    {
+      enabled:
+        !!cloudProvider && !!cloudProviderCredentialIdentifier && !!region,
+    }
+  );
+
+  return {
+    machineTypes: machineTypes ?? [],
```
ferozemohideen, line 65: if you change the line above to 
```
const { data: machineTypes = [], isLoading } = useQuery(
```
then this can just be
```
return { machineTypes, isLoading}
```

ferozemohideen (approved):

PR #4574, opened 2024-04-23
Add porter standard trial days

MauAraujo:
## What does this PR do?
This PR adds 15 days as default trial for Porter Standard users.

### DIFF
--- a/internal/billing/metronome.go
+++ b/internal/billing/metronome.go
@@ -57,6 +57,7 @@ func (m MetronomeClient) CreateCustomerWithPlan(ctx context.Context, userEmail s
 	var trialDays int
 	planID := m.PorterStandardPlanID
 	projID := strconv.FormatUint(uint64(projectID), 10)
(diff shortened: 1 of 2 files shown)


internal/billing/metronome.go
```diff
@@ -127,6 +132,12 @@ func (m MetronomeClient) addCustomerPlan(ctx context.Context, customerID uuid.UU
 		StartingOnUTC: startOn,
 	}
 
+	if trialDays > 0 {
```
ianedwards, line 135: maybe make `trialDays` a uint and have this condition be

`if trialDays != 0`

usually just helpful to check whether it's the 0 value so you know if it's a default or not

ianedwards (approved):

internal/billing/metronome.go
MauAraujo: Thanks for pointing out! Fixed

PR #4581, opened 2024-04-25
Add button to show intercom on billing modal

MauAraujo:
## What does this PR do?
Add a small prompt on the blocking billing modal for users who have an expired trial but had already redeemed the startup deal on a call/email. Note that this blocking modal is only shown for users who don't have an active trial, and most of the startup deal users should already have it because of the reconciliation in Metronome. 

[screen.webm](https://github.com/porter-dev/porter/assets/19579265/56d38abb-a2ba-4f0f-a2d0-3458f7c7de6d)

### DIFF
--- a/dashboard/src/main/home/modals/BillingModal.tsx
+++ b/dashboard/src/main/home/modals/BillingModal.tsx
@@ -1,4 +1,4 @@
-import React, { useContext, useState } from "react";
+import React, { useContext } from "react";
 import { Elements } from "@stripe/react-stripe-js";
 import { loadStripe } from "@stripe/stripe-js";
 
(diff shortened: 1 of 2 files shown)


dashboard/src/main/home/modals/BillingModal.tsx
```diff
@@ -24,6 +25,17 @@ const BillingModal = ({
   const { currentProject } = useContext(Context);
   const { publishableKey } = usePublishableKey();
 
+  const [showIntercom, setShowIntercom] = useState(false);
```
ferozemohideen, line 28: don't need this state, onClick can just call showIntercomWithMessage

ferozemohideen (approved):
there's some weird space in the middle of the modal, can we remove that?

dashboard/src/main/home/modals/BillingModal.tsx
MauAraujo: I tried that initially but the intercom popup would appear everytime I opened the dashboard

MauAraujo: nevermind, fixed this and the weird spacing, thanks for pointing out

PR #4582, opened 2024-04-26
add initdeploy job to preview template

ianedwards:
## What does this PR do?

- Gives users the option to add an initial deploy job to their preview templates
- will run following deploy of first revision of an app
- still hidden from existing app screen, since it can't be edited


dashboard/src/lib/porter-apps/index.ts
```diff
@@ -236,41 +242,100 @@ export function serviceOverrides({
...
+          }),
+          expanded: true,
+        })
+    )
+    .otherwise(() =>
+      deserializeService({
+        service: serializedServiceFromProto({
           service: new Service({
-            ...overrides.predeploy,
+            ...(overrides.predeploy ?? {}),
```
ferozemohideen, line 281: can add a match clause for if predeploy/initdeploy is undefined right? not sure if it's worth the verbosity/readability tradeoff tho

dashboard/src/main/home/app-dashboard/validate-apply/services-settings/ServiceList.tsx
```diff
@@ -126,12 +129,35 @@ const ServiceList: React.FC<ServiceListProps> = ({
   const services = useMemo(() => {
     // if predeploy, only show predeploy services
     // if not predeploy, only show non-predeploy services
+    if (lifecycleJobType === "predeploy") {
```
ferozemohideen, line 132: can use match here

ferozemohideen (approved):

dashboard/src/lib/porter-apps/index.ts
ianedwards: yeah i think we can guarantee it's defined here

PR #4583, opened 2024-04-26
Use date-fns for relative time functions

MauAraujo:
## What does this PR do?
The existing function in the code that we were using to display relative time formats had a bug where if the date was between 8 months and a year, it displayed "8 months" regardless of the actual difference. The rest of dates were displayed correctly, but to avoid any future problems I instead change those function to use the `date-fns` library.


dashboard/src/main/home/Home.tsx
```diff
@@ -377,13 +377,7 @@ const Home: React.FC<Props> = (props) => {
...
       return true;
     }
-
-    const diff = timeFrom(timestamp);
-    if (diff.when === "future") {
-      return false;
-    }
-
-    return true;
+    return isPast(new Date(timestamp)) ? true : false;
```
ianedwards, line 380: `return isPast(new Date(timestamp))`

ianedwards (approved):

MauAraujo:
Closing in favor of #4584

PR #4595, opened 2024-05-01
dgtown/auth 2

d-g-town:
- **refactoring main, adding separate authn/z providers**
- **address comments**

### DIFF
--- a/dashboard/src/shared/auth/AuthnContext.tsx
+++ b/dashboard/src/shared/auth/AuthnContext.tsx
@@ -1,0 +1,89 @@
+import React, { useContext, useEffect, useState } from "react";
+
+import api from "shared/api";
+import { Context } from "shared/Context";
+
+type AuthnState = {
+  userId: number;
+  isLoggedIn: boolean;
+  isEmailVerified: boolean;
+  handleLogOut: () => void;
+  authenticate: () => void;
+  isLoading: boolean;
+  hasInfo: boolean;
+};
+
+export const AuthnContext = React.createContext<AuthnState>({} as AuthnState);
+
+const AuthnProvider = ({
+  children,
+}: {
+  children: JSX.Element;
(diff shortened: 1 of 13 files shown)


dashboard/src/shared/auth/AuthnContext.tsx
```diff
@@ -0,0 +1,183 @@
...
+  handleLogOut: () => void;
+};
+
+const NilAuthnState = {
+    userId: -1,
+    handleLogOut: () => {},
+};
+
+export const AuthnContext = React.createContext<AuthnState>(NilAuthnState);
+
```
ianedwards, line 26: I'd maybe just use the pattern that we have in the other context providers, which will just error if the values haven't been set yet.

```suggestion

export const AuthnContext = React.createContext<AuthnState | null>(null);

export const useAuthState = (): AuthnSate => {
  const context = useContext(AuthnContext)
  if (context == null) {
    throw new Error(
      "useAuthState must be used within a AuthnContext"
    );
  }
  return context;
}
```

dashboard/src/shared/auth/AuthnContext.tsx
```diff
@@ -0,0 +1,183 @@
...
+          render={() => {
+            return <VerifyEmail handleLogOut={handleLogOut} />;
+          }}
+        />
+      </Switch>
+    );
+  }
+
+  // Handle case where new user signs up via OAuth and has not set name and company
+  if (!hasInfo && userId > 9312) {
```
ianedwards, line 153: where's this come from out of curiosity

ianedwards (approved):

dashboard/src/shared/auth/AuthnContext.tsx
d-g-town: Comes from the old code (Main.tsx, line 161)

PR #4596, opened 2024-05-01
Consolidate usage and billing tabs

jusrhee:
## What does this PR do?
image
image

image


internal/billing/metronome.go
```diff
@@ -454,6 +490,11 @@ func (m MetronomeClient) do(method string, path string, body interface{}, data i
 		}
 	}
 
+	// Add raw query parameters to the endpoint
+	if queryParams != "" {
+		endpoint += "?" + queryParams
+	}
```
ianedwards, line 496: one option here might be to use the URL type, and consolidate path and params into a single arg:

https://pkg.go.dev/net/url

ianedwards (approved):

PR #4613, opened 2024-05-06
Fix total cost resetting

MauAraujo:
## What does this PR do?
Fix a small bug where the total cost is set to zero initially, and if refreshing. This PR fixes that issue by computing the total when rendering the chart.

### DIFF
--- a/dashboard/src/main/home/project-settings/UsagePage.tsx
+++ b/dashboard/src/main/home/project-settings/UsagePage.tsx
@@ -44,16 +44,7 @@ function UsagePage(): JSX.Element {
   );
 
   const computeTotalCost = (costs: CostList): number => {
-    let total = 0;
-    costs?.forEach(
-      (cost: {
-        start_timestamp: string;
-        end_timestamp: string;
-        cost: number;
-      }) => {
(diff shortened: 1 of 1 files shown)


dashboard/src/main/home/project-settings/UsagePage.tsx
```diff
@@ -41,7 +42,20 @@ function UsagePage(): JSX.Element {
...
+    costs?.forEach(
+      (cost: {
+        start_timestamp: string;
+        end_timestamp: string;
+        cost: number;
+      }) => {
+        total += cost.cost;
+      }
+    );
+    return parseFloat(total.toFixed(2));
```
ianedwards, line 57:
```suggestion
    const total = costs.reduce((acc, curr) => acc + curr, 0);
    return parseFloat(total.toFixed(2));
```

ianedwards (approved):

dashboard/src/main/home/project-settings/UsagePage.tsx
MauAraujo: Thanks!

PR #4616, opened 2024-05-07
move ccp connection to cc link

sunguroku:


dashboard/src/main/home/modals/PaymentSetupForm.tsx
```diff
@@ -57,6 +61,26 @@ const PaymentSetupForm = ({ onCreate }: { onCreate: () => Promise<void> }) => {
     // Confirm the setup and set as default
     if (setupIntent?.payment_method !== null) {
       await setDefaultPaymentMethod(setupIntent?.payment_method as string);
+
+      // create cluster on first payment setup for sandbox
+      if (currentProject?.sandbox_enabled) {
+        await api.connectProjectToCluster(
```
MauAraujo, line 67: If a user adds another payment method, this will create another cluster connection right? Should we limit to one credit card?

dashboard/src/main/home/modals/PaymentSetupForm.tsx
```diff
@@ -57,6 +61,26 @@ const PaymentSetupForm = ({ onCreate }: { onCreate: () => Promise<void> }) => {
...
+          { id: currentProject.id }
+        )
+        .then(() => {
+          api.inviteAdmin(
+            "<token>",
+            {},
+            { project_id: currentProject.id }
+          )
+        })
+        .catch((err: any) => {
```
MauAraujo, line 79: The payment method will be added to Stripe even if this call fails, is there another way to create the cluster if that happens?

PR #4624, opened 2024-05-08
Neon frontend

ferozemohideen:
user can auth with neon, create neon datastore, see credentials and delete it


dashboard/src/main/home/database-dashboard/shared/NeonIntegrationModal.tsx
```diff
@@ -0,0 +1,60 @@
...
+    <Modal closeModal={onClose} width={"800px"}>
+      <Text size={16}>Integrate Neon</Text>
+      <Spacer y={0.5} />
+      <Text color="helper">
+        To continue, you must authenticate with Neon.{" "}
+        <Link target="_blank" onClick={openAuthWindow} hasunderline>
+          Authorize Porter to create Neon datastores on your behalf
+        </Link>
+      </Text>
+    </Modal>
```
ianedwards, line 56: is it worth showing this modal in a blocked / loading state after they go through the flow with neon, so they don't end up proceeding before we save the integration on our end?

ianedwards (approved):
nice

dashboard/src/main/home/database-dashboard/shared/NeonIntegrationModal.tsx
ferozemohideen: they cant proceed; if they click create again, they'll just see the same modal

PR #4629, opened 2024-05-09
add ory creation webhook

d-g-town:


api/server/handlers/user/create_ory.go
```diff
@@ -0,0 +1,133 @@
...
+	ctx, span := telemetry.NewSpan(r.Context(), "serve-create-ory-user")
+	defer span.End()
+
+	// this endpoint is not authenticated through middleware; instead, we check
+	// for the presence of an ory action cookie that matches env
+	oryActionCookie, err := r.Cookie("ory_action")
+	if err != nil {
+		err = telemetry.Error(ctx, span, err, "invalid ory action cookie")
+		reqErr := apierrors.NewErrForbidden(err)
+		apierrors.HandleAPIError(u.Config().Logger, u.Config().Alerter, w, r, reqErr, true)
```
ferozemohideen, line 54: why not use u.HandleAPIError here?

ferozemohideen (approved):

api/server/handlers/user/create_ory.go
d-g-town: no idea, fixed

ianedwards (approved):

PR #4644, opened 2024-05-15
Inference fe

yosefmih:


dashboard/src/components/porter/CenterWrapper.tsx
```diff
@@ -0,0 +1,17 @@
+import React from "react";
```
yosefmih, line 1: pretty much ignore everything in this folder, it is from the previous iteration

yosefmih: the main changes are in main and lib

